PR #6, opened 2019-10-22
add docker build options

gwendolynbrook:


Dockerfile.build
```diff
@@ -0,0 +1,9 @@
+# Build using golang container
+FROM golang:1.12
```
doanac: I'd make this golang:1.12-alpine and the container should build a bit faster.

PR #10, opened 2019-12-13
Create a "targets edit" command

doanac:
This command allows a user to make direct changes to the targets.json
file for a Factory. It tries its best to ensure no garbage is allowed
to be uploaded. However, this command should still be used with
caution.

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -232,6 +232,34 @@ func (a *Api) Post(url string, data []byte) (*[]byte, error) {
 	return &body, nil
 }
 
+func (a *Api) Put(url string, data []byte) (*[]byte, error) {
+	client := http.Client{
+		Timeout: time.Second * 10,
+	}
+
+	req, err := http.NewRequest(http.MethodPut, url, bytes.NewBuffer(data))
+	if err != nil {
+		return nil, err
+	}
+
+	a.setReqHeaders(req, true)
+
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


doanac:
@mike-scott - ping

mike-scott:
@doanac LGTM, but I'm wondering what kind of use-case you envisioned for this?  Seems fairly dangerous.

doanac:
@mike-scott - (ab)use of this command can indeed cause havoc. I've had a few times when something like this would be handy such as:

 * pruning out old dockerapp target entries.
 * changing tags in a way our command couldn't handle
 * or doing 2 changes at once so that we don't queue up 2 CI jobs

But it can be damaging. One thing to note: the CI job keeps a before and after targets.json, so you could "revert" a change by downloading that file.

mike-scott:
I wish there was a way to restrict this to a certain permission or something.  I do like the functionality and agree there are probably good use cases.. but this seems really dangerous for the average user.

doanac:
@mike-scott - i just pushed a new revision. This revision puts a `Hidden: true` flag on the subcommand so that `fioctl targets --help` won't show it. Basically it becomes a hidden command that you'd have to know about to use. You think that's better or should we just not support this command?

mike-scott:
> @mike-scott - i just pushed a new revision. This revision puts a `Hidden: true` flag on the subcommand so that `fioctl targets --help` won't show it. Basically it becomes a hidden command that you'd have to know about to use. You think that's better or should we just not support this command?

That's perfect.

PR #11, opened 2019-12-18
Add login command

MiloCasagrande:
This adds the `login` command: there is no logic yet to use the OAuth token obtained after login to perform the API calls.

We need to update the `client` to use the config as loaded from file and add some logic in there as well: it needs to be able to refresh the token if it's expired before using it for any API calls.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
@@ -52,11 +53,11 @@ func (c *Config) ReadClientCredentials() {
 
 	fmt.Print("Client ID: ")
 	scanner.Scan()
-	clientId := scanner.Text()
+	clientId := strings.Trim(scanner.Text(), " ")
 
 	fmt.Print("Client secret: ")
 	scanner.Scan()
-	clientSecret := scanner.Text()
+	clientSecret := strings.Trim(scanner.Text(), " ")
 
 	if clientId == "" || clientSecret == "" {
 		fmt.Println("Client ID and client credentials are both required.")
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -47,7 +47,9 @@ func (c *Config) ReadClientCredentials() {
 
 	scanner := bufio.NewScanner(os.Stdin)
 
-	fmt.Printf("Please provide the client ID and client secret\n\n")
+	fmt.Println("Please provide the client ID and client secret.")
+	fmt.Printf("To obtain your client ID and client secret, please go to: https://app.foundries.io/settings/tokens/\n\n")
+
 	fmt.Print("Client ID: ")
 	scanner.Scan()
 	clientId := scanner.Text()
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -60,7 +60,7 @@ func (c *Config) ReadClientCredentials() {
 	clientSecret := strings.Trim(scanner.Text(), " ")
 
 	if clientId == "" || clientSecret == "" {
-		fmt.Println("Client ID and client credentials are both required.")
+		fmt.Println("Client ID and client secret are both required.")
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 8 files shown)


internal/config/config.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
+)
+
+type Config struct {
```
doanac: Any specific reason you are marshaling this yourself now? Viper can do this and it will avoid a missed config setting like was done here with "factory" gettting missed and then overwritten without it.

internal/oauth/oauth.go
```diff
@@ -0,0 +1,190 @@
...
+
+const uri string = "https://app.foundries.io/oauth"
+
+// OAuth interface.
+type OAuth interface {
+	Get() int
+	HasRefreshToken() bool
+	IsExpired() (bool, error)
+	Refresh() int
+	post(string, url.Values) (*[]byte, error)
```
doanac: `post` isn't needed here. Since its lower case its not exposed anyway.

internal/oauth/oauth.go
```diff
@@ -0,0 +1,190 @@
...
+	"time"
+
+	"github.com/foundriesio/fioctl/internal/config"
+	"github.com/sirupsen/logrus"
+)
+
+const uri string = "https://app.foundries.io/oauth"
+
+// OAuth interface.
+type OAuth interface {
```
doanac: Any reason you are doing the interface versus just returning ClientCredentials? ie - its exposing the same methods (since they are UpperCase). I guess it also has "Config" and "URL", but you could make those lowercase and get the same effect.

internal/oauth/oauth.go
```diff
@@ -0,0 +1,190 @@
...
+		return 2
+	}
+
+	updateConfig(c.Config, &t)
+
+	return 0
+}
+
+// GetClient creates and returns a new OAuth client.
+func GetClient(c *config.Config) OAuth {
```
doanac: No biggie here, its just more idiomatic to call it `NewClient` or even `NewClientCredentials`.

internal/oauth/oauth.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return false, nil
+}
+
+// HasRefreshToken checks if a refresh token is available.
+func (c *ClientCredentials) HasRefreshToken() bool {
+	if c.Config.ClientCredentials.RefreshToken != "" {
+		return true
+	}
+	return false
```
doanac: linter caught this: you can just do `return c.Config.ClientCredentials.RefreshToken != ""` pretty decent linting logic.

doanac:
Here's a slightly simplified version that helps get client config integrated into the actual api client and user Viper to marshal/unmarshal settings.

Next step is to try and use the token. I'll start working on the backend to handle that and then get back to this.

internal/config/config.go
MiloCasagrande: Viper was marshaling everything, even flags like `-v` or similar, and I didn't really like it. It might have been just me not knowing how to correctly use Viper though.

internal/oauth/oauth.go
MiloCasagrande: My initial idea was to abstract on that in order to potentially handle other OAuth credentials flow, like the device registration, that have slightly different flows but could share a common interface.

internal/oauth/oauth.go
MiloCasagrande: Good to know. Have to check VS Code go linter better, it didn't report that.

internal/oauth/oauth.go
MiloCasagrande: Can't remember why I wanted it there, but yes, lower-case declarations are not exposed.

internal/config/config.go
doanac: I'm struggling with this thing also. I'm going to spend a little more time on it and see if i can clean it up. The "-v" one is the thing that's annoying. If you run a command with "-v" that happens to do a token refresh, then your config will get updated.

doanac:
The other branch that does this has been merged.

PR #13, opened 2020-01-09
Add Oauth Login Support

doanac:


doanac:
@MiloCasagrande - this is close. There's a problem with `fioctl targets ...`, but I think its server side. I'll take a look in the morning, but this should give you something to think on.

cmd/login.go
```diff
@@ -0,0 +1,134 @@
...
+
+func promptForCreds() (string, string) {
+	logrus.Debug("Reading client ID/secret from stdin")
+
+	scanner := bufio.NewScanner(os.Stdin)
+
+	fmt.Printf("Please provide the client ID and client secret\n\n")
+	fmt.Print("Client ID: ")
+	scanner.Scan()
+	clientId := scanner.Text()
```
MiloCasagrande: Shouldn't we trim those input strings? The values from the UI will never contain empty spaces (left/right) and will never be just empty spaces.

doanac: wow - i copy/pasted that. But when I look at your branch it does the trimming. I have no idea how I lost that change. I'll fix.

MiloCasagrande: I sort of saw the same behaviour in VSCode and its automagic linter: if I don't have the import at the top, saving the file will clean up all the code that reference somethings not being imported.

Though in this case it would appear to be some pretty damn smart linter to remove just the trim part. :-)

doanac:
@MiloCasagrande - this latest "Don't persist config and verbose flags " commit  prevents viper from saving the "verbose" and "config" args.

PR #16, opened 2020-01-30
Have targets-prune remove unused docker apps

doanac:
This helps reduce the size of target.json by also removing .dockerapp
files in the reposerver that are no longer referenced as a result of
removing a target(s).


mike-scott:
Oh god yes.  I was wondering about this functionality.

PR #17, opened 2020-02-03
allows for deletion nil/null values

marcstreeter:
allows nil/null to be set to the secret value to trigger deletion of a secret


doanac:
Looks good. I'll give this a spin once we get the JobServ PR merged and deployed

PR #19, opened 2020-02-17
offline key rotation

marcstreeter:
adds offline key rotation


marcstreeter:
> My biggest problem with this is the following:
> 
> ```
> $ fioctl keys rotate /tmp/andy-corp.tgz
> Prerotation: validating environment...
> Pulling aktualizr image √
> Prerotation: creating backup of original offline keys:
> 	"/tmp/andy-corp.tgz.bak"
> Rotating offline credentials x
> 
> Error: failed to parse credentials
> ```
> 
> You've done a lot of work to try and shield the user from the output of the actual key rotation. The problem is:
> 
> 1. You can control the output of the python script you are running. So we've got unnecessary complexity.
> 2. When things go side-ways the user has no idea what the problem is. I suspect somewhere in the swallowed output is something that might help me understand the problem.
> 
> I vote for purely shelling out the docker/python logic and let it stream directly to stdout/stderr. You can try and make that more friendly if you want, but remember: this is only going to be run is something _very_ bad has happened. A user in this situation is going to want too much information instead of too little.

My main objective was to improve the user experience of these long running tasks via easy to read notifications and progress animation (mainly for long running docker, tar, and garage signing bits). I also wasn't sure just how granular the output should be given that we provide the users of this tool with the output.tgz they are inputting.  I agree though that, when the garage signing fails a user will most likely want more detail.  I'll update the output.

cmd/keys_rotate.go
```diff
@@ -0,0 +1,366 @@
...
+	ErrUnpackCredentials, ErrPullCredentials, ErrParseCredentials, ErrRegenerateCredentials, ErrAddNewCredentials,
+	ErrRemoveOldCredentials, ErrSignCredentials, ErrPackCredentials, ErrPushCredentials}
+
+func doKeyRotation(cmd *cobra.Command, args []string) {
+	factory := viper.GetString("factory")
+	logrus.Debugf("Rotating factory keys for: %s", factory)
+	credentialsPath := args[0]
+	credentialsBackupPath := fmt.Sprintf("%s.bak", credentialsPath)
+	fmt.Printf("Prerotation: validating environment...\n")
+	if err := verifyPath(credentialsPath); err != nil {
```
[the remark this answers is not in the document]
marcstreeter: verify path is only meant to provide clear and concise error messaging.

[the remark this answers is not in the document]
doanac: yes - but can't you get that from the error that os.Link gives you? There's a common mistake people do in code like this:
~~~
if not file.exists():
  exit-error
open_file() # BUG - this *can* exit with an error in multi-thread code
~~~

cmd/keys_rotate.go
```diff
@@ -0,0 +1,366 @@
...
+
+func verifyDocker() error {
+	logrus.Debugf("checking if docker is available")
+	if _, err := cli("which docker"); err != nil {
+		return fmt.Errorf("docker not available")
+	}
+	return nil
+}
+
+func copyFile(source string, target string) error {
```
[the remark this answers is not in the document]
doanac: that's true.

cmd/keys_rotate.go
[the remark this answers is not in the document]
doanac: I say just delete verifyPath.

cmd/keys.go
```diff
@@ -0,0 +1,16 @@
...
+
+var keysCmd = &cobra.Command{
+	Use:              "keys",
+	Short:            "Manage keys in use by your factory fleet",
+	PersistentPreRun: assertLogin,
+}
+
+func init() {
+	rootCmd.AddCommand(keysCmd)
+	requireFactory(keysCmd)
```
doanac: This command doesn't require the factory cli opts, so delete this line.

doanac:
This is close, but please just show the entire output of the key rotation. This isn't something we should hide from users. Its good that they see everything done to rotate a key. Also - I don't know if your output is right in the case of a failed command still:
~~~
Pulling aktualizr image √
Offline key backed up: "/tmp/andy-corp.tgz.bak"
Rotating offline credentials x
Error: failed to parse credentials
  | Pulled root.json
current root keyid: 0f352ad129e6614afda4293952db92a0078d7ea58fd23eb382076939abc40e3b
~~~
I bet you are missing stderr. That's why you should just let both stream to the console

marcstreeter:
> This is close, but please just show the entire output of the key rotation. This isn't something we should hide from users. Its good that they see everything done to rotate a key. Also - I don't know if your output is right in the case of a failed command still:
> 
> ```
> Pulling aktualizr image √
> Offline key backed up: "/tmp/andy-corp.tgz.bak"
> Rotating offline credentials x
> Error: failed to parse credentials
>   | Pulled root.json
> current root keyid: 0f352ad129e6614afda4293952db92a0078d7ea58fd23eb382076939abc40e3b
> ```
> 
> I bet you are missing stderr. That's why you should just let both stream to the console

good catch -- I'll update it.   Some how my previous post looks like it is streaming standard err where this one is not?  I'll just combine the two

doanac:
> good catch -- I'll update it. Some how my previous post looks like it is streaming standard err where this one is not? I'll just combine the two
You start to run into trouble/complexity when you want to capture them. If you just let them go straight to the console, things are much easier.

cmd/keys_rotate.go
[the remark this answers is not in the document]
marcstreeter: done. io.Copy has a good error

cmd/keys.go
marcstreeter: removed

cmd/keys.go
```diff
@@ -0,0 +1,15 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var keysCmd = &cobra.Command{
+	Use:              "keys",
+	Short:            "Manage keys in use by your factory fleet",
+	PersistentPreRun: assertLogin,
```
doanac: sorry - just noticed this also. This command doesn't need that. It doesn't interact with api.foundries.io

cmd/keys_rotate.go
```diff
@@ -0,0 +1,360 @@
...
+
+def cmd(*args, cwd=None):
+    print('running: %s' % ' '.join(args))
+    p = subprocess.Popen(
+        args, cwd=cwd,stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
+    for line in p.stdout:
+        sys.stdout.buffer.write(line)
+    p.wait()
+    if p.returncode != 0:
+        raise subprocess.CalledProcessError(p.returncode, args)
```
doanac: Now that you've removed the original formatting I was doing the output, there's no point in all this. calling subprocess.check_call(args, cwd=cwd) would be the same.

cmd/keys_rotate.go
```diff
@@ -0,0 +1,360 @@
...
+      sys.exit(8)
+    try:
+      cmd('garage-sign', 'root', 'push', '--repo', './tufrepo')
+    except:
+      sys.exit(9)
+`
+	progress := newProgress("Pulling aktualizr image %s")
+	if err := pullContainer(aktualizrImageName); err != nil {
+		progress.fail()
+		exitf("failed to pull image, %q", aktualizrImageName)
```
doanac: I just hit this: You aren't printing the error that occurs.

cmd/keys_rotate.go
```diff
@@ -0,0 +1,360 @@
...
+		return fmt.Errorf("backup permissions failed: %w", err)
+	}
+	return nil
+}
+
+func cli(inputs ...string) ([]byte, error) {
+	args := strings.Join(inputs, " && ")
+	cmd := exec.Command("/bin/sh", "-c", args)
+	logrus.Debugf("cli executing: %v", cmd)
+	return cmd.CombinedOutput()
```
doanac: I'm guessing you had bigger plans for this originally. But you aren't making multple commands calls here requiring the "shell -c" and the "&&" stuff. I think you just need: `cmd := exec.Command(inputs)`

marcstreeter: lol that is exactly the case, fixed

cmd/keys_rotate.go
marcstreeter: good catch - fixed

cmd/keys_rotate.go
marcstreeter: the python script fails if I use `subprocess.check_call`

cmd/keys.go
marcstreeter: good catch, removed

marcstreeter:
I have a separate PR -- in the case that streaming all output is preferred. I can definitely see the benefit of always streaming output.  while I could do that with this PR and also add ctrl+c protection rather easily. I can also appreciate that this command should be rarely hit.

Here is the output from the other PR https://github.com/foundriesio/fioctl/pull/21
```
fioctl keys rotate "/tmp/andy-corp.tgz"
Docker version 19.03.5, build 633a0ea
Pulling aktualizr image...
Using default tag: latest
latest: Pulling from aktualizr
Digest: sha256:a89f306e297de7e9b37f30f851d345340f1aa7619e0fcb6566ee84920984de75
Status: Image is up to date for hub.foundries.io/aktualizr:latest
hub.foundries.io/aktualizr:latest
Original credentials --> "/tmp/andy-corp.tgz.bak"
Rotating offline credentials...
running: tar xf /tmp/creds.tgz
running: garage-sign root pull --repo ./tufrepo
Pulled root.json
finding root key name and key id
current root keyid: ecd598fcb44d78c829240b66dd8a3aff7c619f9131832e7fee86cc5f48134222
could not find root key name
Error: failed key rotation Unable to run '[docker run --rm --env PYTHONUNBUFFERED=1 -v /tmp/andy-corp.tgz:/tmp/creds.tgz -v /tmp/500063097tmpScript:/tmp/tmp.py hub.foundries.io/aktualizr /tmp/tmp.py]': err=exit status 1
```

marcstreeter:
closing this PR in favor of https://github.com/foundriesio/fioctl/pull/21

PR #20, opened 2020-02-17
added output for up-to-date attributes

marcstreeter:
added output for up-to-date attributes


doanac:
Looks good so far. This attribute should also be shown in the device-show command as well. And if you reference the bug# your commit message, then github will link the two things together.

cmd/device_show.go
```diff
@@ -57,4 +57,5 @@ func doDeviceShow(cmd *cobra.Command, args []string) {
 		os.Stdout.Write(b)
 		fmt.Println("")
 	}
+	fmt.Printf("Up to date:\t%v\n", device.UpToDate)
```
doanac: Lets move this higher up in the list. Maybe after "Factory".

PR #21, opened 2020-02-21
Feature/key rotation

marcstreeter:
always streaming output for key rotation

### DIFF
--- a/cmd/keys.go
+++ b/cmd/keys.go
@@ -1,0 +1,14 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var keysCmd = &cobra.Command{
+	Use:   "keys",
+	Short: "Manage keys in use by your factory fleet",
+}
+
+func init() {
+	rootCmd.AddCommand(keysCmd)
+}
--- a/cmd/keys_rotate.go
+++ b/cmd/keys_rotate.go
@@ -1,0 +1,230 @@
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+var keysRotateCmd = &cobra.Command{
+	Use:   "rotate <offline key archive>",
+	Short: "Rotate root signing key used by the Factory",
+	Run:   doKeyRotation,
+	Args:  cobra.MinimumNArgs(1),
+}
+
+func init() {
+	keysCmd.AddCommand(keysRotateCmd)
+}
+
+func doKeyRotation(cmd *cobra.Command, args []string) {
+	credentialsPath := args[0]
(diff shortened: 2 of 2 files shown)


marcstreeter:
this could be simplified greater (and not commit the heresy of copy pasta) if the foundries/tuftree/client had a `go.mod` set up

marcstreeter:
this PR is proposed instead of https://github.com/foundriesio/fioctl/pull/19

doanac:
> this could be simplified greater (and not commit the heresy of copy pasta) if the foundries/tuftree/client had a go.mod set up

That's dead code. we don't want to use it like a module.

marcstreeter:
> > this could be simplified greater (and not commit the heresy of copy pasta) if the foundries/tuftree/client had a go.mod set up
> 
> That's dead code. we don't want to use it like a module.

💀oh? I didn't realize.

cmd/keys.go
```diff
@@ -0,0 +1,15 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var keysCmd = &cobra.Command{
+	Use:              "keys",
+	Short:            "Manage keys in use by your factory fleet",
+	PersistentPreRun: assertLogin,
```
doanac: don't need that.

cmd/keys_rotate.go
```diff
@@ -0,0 +1,230 @@
...
+	return scriptPath, nil
+}
+
+func runRotationScript(imageName string, sourcePath string, credentialsPath string) error {
+	targetPath := "/tmp/tmp.py"
+	args := []string{
+		// base args
+		"run", "--rm",
+		// env args
+		"--env", "PYTHONUNBUFFERED=1",
```
doanac: ah - i bet that was driving you mad.

marcstreeter: it was 😄

marcstreeter: quick bit of history, I initially tried using the tuft/client (dead) code  we've been talking about and it didn't change output at all (or so I thought at the time).  After switching back and trying with a Pipeline -- same issue, so I then just ran the command directly and noticed that docker wasn't outputting til the end.  After switching that environment variable - it was fixed.  Not with "CombinedOutput" though ( which I tried when doing this streaming version) -- hence the use again of this code. 

Took quite a tour to get here.  It was fun.

cmd/keys.go
marcstreeter: removed

doanac:
I just toyed around with this and it works pretty well. Lets merge this and then we can circle back and get it using a directory rather than .tgz in the near future so that its more bullet-proof.

marcstreeter:
> I just toyed around with this and it works pretty well. Lets merge this and then we can circle back and get it using a directory rather than .tgz in the near future so that its more bullet-proof.

k - just squashed all the commits before the merge to clean up. I'll merge now.

PR #28, opened 2020-05-05
Detect proper config directory for OSX

doanac:


cmd/root.go
```diff
@@ -55,19 +56,34 @@ func init() {
...
 
+func getConfigDir() string {
+	config, err := homedir.Expand("~/.config")
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	if _, err := os.Stat(config); os.IsNotExist(err) {
+		if runtime.GOOS == "darwin" {
+			config, err = homedir.Expand("~/Library/Preferences")
```
MiloCasagrande, line 67: On my mac I have a `~/.config` directory: there are programs that use the UNIX way of storing configurations on mac as well.

By Apple developer docs, app support files, like configurations, should go into `~/Library/Application Support`, but can also go into `~/Library/Preferences` (I have no idea when/how to pick one though). 

The guidelines are for "mac app (native apps), so `fioctl` might not fall into this category and it can keep storing configuration the UNIX way. I guess is more of a "tell the user where the config is if they need to tweak it".

ricardosalveti: Yeah, I also believe it is fine to use home for traditional unix-based software.

doanac: So you guys think I should just see if `~/.config` exists, and if it doesn't just create it? ie - don't mess with the Libary/... directory stuff?

MiloCasagrande: Yeah, for me it would be fine to use that pattern.

doanac: k - i force pushed the fix i think you will agree with.

PR #29, opened 2020-05-05
login: Set default factory if there is one

doanac:
Most customers only belong to a single factory(org). This change detects
that and provides them the default they'll want in their fioctl config
so that they won't have to use the "-f" flag in commands requiring a
factory name.


MiloCasagrande:
One thing, food for thoughts: should we use the factory `_id` value, along with the `name`, or  is it just enough the `name`? My thinking was that when a device is registered/removed, it is reported back on the polis API, and having the `_id` around is easier to access polis.

doanac:
> should we use the factory _id value, along with the name, or is it just enough the name

This raises something important. The OTA system only understands the factory by its *name*. So the id isn't something it could be aware of.

PR #46, opened 2020-08-14
Return a browser-friendly URL to CI job

mike-sul:

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -703,7 +703,8 @@ func (a *Api) TargetImageCreate(factory string, targetName string) (string, erro
 	if err := json.Unmarshal(*resp, &pr); err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


client/foundries.go
```diff
@@ -703,7 +703,8 @@ func (a *Api) TargetImageCreate(factory string, targetName string) (string, erro
 	if err := json.Unmarshal(*resp, &pr); err != nil {
 		return "", err
 	}
-	return pr.JobServUrl + "runs/assemble-system-image/console.log", nil
+	ciWebUrl := strings.Replace(pr.JobServUrl, "api", "ci", 1)
+	return ciWebUrl, nil
```
doanac, line 707: I don't think changing api and ci produces the same URLs. And if the future the CI urls might be pointing into app.foundries.io for the stuff Milo is doing. I think we have to update the ota-lite API to return both kinds or URLs.

PR #48, opened 2020-08-14
Use web-url for CI URL

mike-sul:

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -249,6 +249,21 @@ func NewApiClient(serverUrl string, config Config, caCertPath string) *Api {
 	return &api
 }
 
+func getResponse(resp *[]byte, err error, runName string) (string, string, error) {
+	if err != nil {
+		return "", "", err
+	}
+	type PatchResp struct {
+		JobServUrl string `json:"jobserv-url"`
+		WebUrl string `json:"web-url"`
+	}
+	pr := PatchResp{}
+	if err := json.Unmarshal(*resp, &pr); err != nil {
+		return "", "", err
+	}
+	return pr.JobServUrl + fmt.Sprintf("runs/%s/console.log", runName), pr.WebUrl, nil
+}
+
 func (a *Api) setReqHeaders(req *http.Request, jsonContent bool) {
 	req.Header.Set("User-Agent", "fioctl-2")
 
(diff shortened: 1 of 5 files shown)


mike-sul:
@doanac This is just a general idea, I see a few options here, not sure which would be the best.

client/foundries.go
```diff
@@ -249,6 +249,20 @@ func NewApiClient(serverUrl string, config Config, caCertPath string) *Api {
 	return &api
 }
 
+func getResponse(resp *[]byte, err error, runName string) (string, error) {
+	if err != nil {
+		return "", err
+	}
+	type PatchResp struct {
+		JobServUrl string `json:"web-url"`
```
doanac, line 257: I think this needs to be:
~~~
		JobServUrl string `json:"url"`
		WebUrl string `json:"web-url"`
~~~

Then we return this tuple to the caller.

mike-sul: Do you mean jobserve-url? I thought about it too. But a caller, in this case, is a `fioctl targets <subcommand>` command, the question is what a user can do with the URL that points to  API. Maybe it's useful for the case when fioctl is used programmatically by some other tool.

doanac: isn't the caller is this function printing the value out and then passing it tail function? ie - one should be a web-url and the other should be an API url

mike-sul: yes, I just thought that the web-url like this https://ci.foundries.io/projects/msul-dev01/lmp/builds/439/assemble-system-image/artifacts/console.log is also good enough for the tailing.
Anyway, I think, it makes sense to return both URLs - it is just more flexible and fulfills more use cases.

mike-sul:
@doanac I updated this PR according to the latest discussion about it.

subcommands/targets/image.go
```diff
@@ -27,13 +27,13 @@ func doImage(cmd *cobra.Command, args []string) {
 	inputTarget := args[0]
 	logrus.Debugf("Generating image of Target %s in Factory %s", inputTarget, factory)
 
-	url, err := api.TargetImageCreate(factory, inputTarget)
+	JobServUrl, WebUrl, err := api.TargetImageCreate(factory, inputTarget)
```
doanac, line 30: this looks good. just one nit. Can you make this `jobServUrl, webUrl`. It seems a bit unusual in golang to name a local variable with a capital letter.

mike-sul: absolutely agree.

mike-sul: done

PR #49, opened 2020-08-14
Add the ability validate changes to device update configuration make sense

doanac:


doanac:
This is a bit of a work in progress we should experiment with. Here's some examples of what it can do:
## Preventing a downgrade:
~~~
$ fioctl devices config updates andy-test --dryrun --tags promoted
ERROR: Given tags [promoted] appear to result in a downgrade from version 111 to 80
~~~

## Preventing a bad tag:
~~~
$ fioctl devices config updates andy-test --dryrun --tags promotedd
ERROR: Given tags [promotedd] are not present in targets.json
~~~

## Preventing a bad apps:
~~~
$ fioctl devices config updates andy-test --dryrun --apps foo
ERROR: Given apps [foo] not present in Target version 113 device would update to: [fiotest homeassistant homelab shellhttpd helloworld]
~~~

I'm not sure about the wording of "downgrade". The device wouldn't actually downgrade it would just sort of be purgatory until a new build was performed. We also need to add a "-f" to ignore this and force the change through.

subcommands/common.go
```diff
@@ -164,3 +164,14 @@ func PrintConfigs(deviceConfigs []client.DeviceConfig, listLimit int) int {
 	}
 	return listLimit
 }
+
+func IntersectionInSlices(list1, list2 []string) bool {
```
mike-sul, line 168: That's really interesting to fidn out that golang (base modules) don't include such basic functionality.

subcommands/devices/config_updates.go
```diff
@@ -80,8 +84,38 @@ func loadSotaConfig(device string) *toml.Tree {
...
+		if !match {
+			return false
+		}
+	}
+	return true
+}
+
 // Validate the inputs: Must be alphanumeric, a dash, underscore, or comma
-func validateUpdateArgs() {
+func validateUpdateArgs(device *client.Device) {
```
mike-sul, line 118: Maybe it makes sense to do this validation in the OTA lite API implementation so direct calls to it are validated too?

subcommands/devices/config_updates.go
```diff
@@ -94,6 +128,82 @@ func validateUpdateArgs() {
 		fmt.Println("       apps must be ", pattern)
 		os.Exit(1)
 	}
+
+	tags := device.Tags
+	if len(updateTags) > 0 {
+		tags = strings.Split(updateTags, ",")
+	}
+	apps := device.DockerApps
```
mike-sul, line 136: Will it work for Compose Apps?

doanac: yes. "docker apps" as a device attribute means what array of apps its reporting up. so its package manager agnostic.

subcommands/devices/config_updates.go
doanac: That's a really interesting idea. I'll look into this today.

mike-sul:
It looks like it's better to have the validation in the OTA Lite API implemenation as we expect that the API will be called not just by fioctl?

doanac:
> It looks like it's better to have the validation in the OTA Lite API implemenation as we expect that the API will be called not just by fioctl?

yes. i think that's way better. This will make it easier to build a GUI that's consistent.

doanac:
I've moved the real checking to happen server side:

 https://git.foundries.io/development/cloudplatforms/infrastructure/ota-lite/merge_requests/64

Once merged, we can clean up and simplify this logic.

doanac:
I've implemented this on the server side.

Issue #50, opened 2020-08-18
Failing to input correct credentials with `fioctl login` fails subsequent attempts


MatthewCroughan:
If you were to provide the incorrect input to `fioctl login`, it will fail the next attempt to use `fioctl login`. with:

`ERROR: Unable to retrieve credentials: HTTP_400`

The user would then have to remove their `.config` before they could attempt to use `fioctl login` again.

doanac:
@MatthewCroughan - i believe this is fixed. so i'm closing this out.

PR #54, opened 2020-08-24
bug #50 - don't save creds for bad login

doanac:
We already have logic to save the credentials if they are valid. There's
no point saving them initially - especially since they can be invalid
and cause the user to need to manually edit fioctl.yaml


doanac:
@MatthewCroughan - can you see if this fixes your bug.

MatthewCroughan:
Works exactly as you have suggested. I encountered no issues when attempting to break it.

MatthewCroughan (approved):
I did not expect that the fix would be the removal of a single line. It works exactly as expected. I encountered no issues when attempting to break it.

PR #56, opened 2020-09-01
Param to list apps to preload into a wic image

mike-sul:

### DIFF
--- a/subcommands/targets/image.go
+++ b/subcommands/targets/image.go
@@ -3,31 +3,51 @@ package targets
 import (
 	"fmt"
 	"os"
+	"regexp"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
-var noTail bool
+var (
+	appsShortlist string
+	noTail bool
+)
 
 func init() {
 	var imageCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


subcommands/targets/image.go
```diff
@@ -19,15 +23,28 @@ func init() {
 		Args:  cobra.ExactArgs(1),
 	}
 	cmd.AddCommand(imageCmd)
+	imageCmd.Flags().StringVarP(&appsShortlist, "apps", "", "",
+		"comma,separate,list of Target apps to preload into a resultant image")
```
doanac, line 27: I suppose we should say the default is "all apps"

mike-sul:
> this is slightly orthogonal, but while we are touching the usage portion of this command. The word "target" (which i think is my fault) can be confusing. Maybe our usage statement should say `<target-name>` that's subtle but might help slightly. We could also include an example so that the user sees you need to use a full target name?

Maybe `<target-full-name>`?

> We could also include an example so that the user sees you need to use a full target name?
makes sense

subcommands/targets/image.go
mike-sul: +

mike-sul:
@doanac Updated, I think adding "Example" to the command is really great idea and we should do it for each command.

PR #64, opened 2020-11-18
add support for compose apps to a target show command

vkhoroz:
Sample output:

```
vkhoroz@osf-ubuntu:~/dev/fioctl$ ./bin/fioctl-linux-amd64 -c ./bin/config.yml targets -f msul-dev01 show compose-app 617 app-07 --manifest
Version:
	 hub.foundries.io/msul-dev01/app-07@sha256:74c5a71eb2f2984ac7eae30074054116f87fd3e176cae4d41e8a6ab76d750588

Docker Manifest:
	annotations:
	  compose-app: v1
	config:
	  digest: sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
	  mediaType: application/vnd.oci.image.config.v1+json
	layers:
	- digest: sha256:0ed690c8ca12e76c23e2bcbd939a9eb158f38528251941b5329dd6f7200d0524
	  mediaType: application/octet-stream
	  size: 520
	mediaType: application/vnd.oci.image.manifest.v1+json
	schemaVersion: 2

Files:
	 Dockerfile
	 docker-compose.yml

Docker Compose Template:
	services:
	  nginx-01:
	    image: hub.foundries.io/msul-dev01/nginx@sha256:813f0553ffec94ed53ca285e807c20735a9b44f708e89730dc20671d3ea37ad2
	    ports:
	    - ${PORT-9999}:80
	    restart: always
	  nginx-02:
	    image: docker.io/library/nginx@sha256:a97eb9ecc708c8aa715ccfb5e9338f5456e4b65575daf304f108301f3b497314
	    ports:
	    - ${PORT-8888}:80
	    restart: always
	  python-www:
	    image: hub.foundries.io/msul-dev01/app-07@sha256:8cc48ee3c2a5cc9c55911853fa6bbd42196b9d4bddd30a7b8142c7e6fc0380e8
	    ports:
	    - 9987:80
	    restart: always
	    volumes:
	    - .:/var/http
	version: "3.2"
```

Without a `--manifest` option a `Docker Manifest` section is omitted.


client/foundries.go
```diff
@@ -703,6 +716,37 @@ func (a *Api) TargetImageCreate(factory string, targetName string, appShortlist
...
+	log := logrus.WithFields(logrus.Fields{"url": url})
+	logrus.Debugf("TargetApp with url: %s", url)
+
+	resp, err := a.RawGet(url, nil)
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	} else if resp.StatusCode != 200 {
+		log.WithFields(logrus.Fields{"status": resp.StatusCode}).Error("HTTP Error")
+		// Don't return, we can get some debug info from response.
```
doanac, line 731: I don't think you can unmarshall *all* http errors as a ComposeAppInspect below.

client/foundries.go
```diff
@@ -174,6 +174,19 @@ type TufCustom struct {
 	OverridesSha   string                `json:"meta-subscriber-overrides-sha,omitempty"`
 }
 
+type ComposeAppContent struct {
+	Files       []string               `json:"files"`
+	ComposeSpec map[string]interface{} `json:"compose_spec"`
+}
+
+type ComposeAppInspect struct {
```
doanac, line 182: Let's call this a ComposeAppBundle. There's some history where that's what we've called it.

go.sum
```diff
@@ -26,6 +26,7 @@ github.com/aristanetworks/goarista v0.0.0-20170210015632-ea17b1a17847/go.mod h1:
 github.com/armon/consul-api v0.0.0-20180202201655-eb2c6b5be1b6/go.mod h1:grANhF5doyWs3UAsr3K4I6qtAmlQcZDesFNEHPZAzj8=
```
doanac, line 26: Was there no corresponding go.mod change? Or is this just a refresh of our current deps?

doanac (commented):
good work. just a couple of small things

go.sum
vkhoroz: This is just what `make build` did to it.

client/foundries.go
vkhoroz: yes, that's true... but if it fails to read content or unmarshall it - we'll return an error at lines 738 and 744 respectively.

client/foundries.go
vkhoroz: no problem, will change this

client/foundries.go
vkhoroz: Actually, the only few cases when we cannot unmarshall are:
- nginx side error (e.g. 502 when ota-lite is down or smth like that)
- an internal application error
- a 400 or 404 error in case if fioctl does something insane

In all other cases ota-lite will return ComposeAppBundle, just some fields may be empty.

doanac: but instead of returning an http error, you'll return a marshalling error and the user wouldn't know the cause, right? i.e maybe this should be captured as "orig-err" and return that if marshalling fails?

vkhoroz: There are several things to it:
- first, I tried to keep the same logic as in other places, the only deviation is to try unmarshal in case of a "sane" HTTP error
- second, a user is able to see all the details with a -v option.

So, I didn't want to complicate the logic with many if statements, though I think that's possible.
Will try to return an original HTTP error in case if unmarshall fails (or combine both into one error).

vkhoroz: This is what I currently get in case if unmarshal fails after an HTTP error:
```
vkhoroz@osf-ubuntu:~/dev/fioctl$ ./bin/fioctl-linux-amd64 -c ./bin/config.yml targets -f msul-dev01 show compose-app 617 app-07
ERRO[0002] HTTP Error                                    status="502 Bad Gateway" url="https://api.foundries.io/ota/factories/msul-dev01/targets/intel-corei7-64-lmp-617/compose-apps/app-07/"
ERRO[0002] invalid character '\x11' looking for beginning of value  url="https://api.foundries.io/ota/factories/msul-dev01/targets/intel-corei7-64-lmp-617/compose-apps/app-07/"
ERROR:invalid character '\x11' looking for beginning of value
```
I think it's good enough, isn't it ?

vkhoroz: Changed this to be more informative:
```
ERRO[0002] HTTP Error: 502 Bad Gateway                   status=502 url="https://api.foundries.io/ota/factories/msul-dev01/targets/raspberrypi4-64-lmp-617/compose-apps/app-07/"
ERRO[0002] Parse Error: invalid character 'g' in literal null (expecting 'u')  url="https://api.foundries.io/ota/factories/msul-dev01/targets/raspberrypi4-64-lmp-617/compose-apps/app-07/"
ERROR: Parse Error: invalid character 'g' in literal null (expecting 'u')
 for 'https://api.foundries.io/ota/factories/msul-dev01/targets/raspberrypi4-64-lmp-617/compose-apps/app-07/' HTTP response: 502 Bad Gateway
=nginx bad gateway html response
```

doanac (approved):

PR #67, opened 2020-12-01
add device filtering by group

vkhoroz:


doanac:
For future reference I think we should drop jira ticket names in public PRs. They used to do this in aktualizr and I found it to be difficult to work with as an "outsider".

subcommands/devices/list.go
```diff
@@ -33,6 +34,7 @@ func init() {
 	listCmd.Flags().BoolVarP(&deviceNoOwner, "skip-owner", "", false, "Do not include owner name when lising. (command will run faster)")
 	listCmd.Flags().StringVarP(&deviceByTag, "by-tag", "", "", "Only list devices configured with the given tag")
 	listCmd.Flags().StringVarP(&deviceByFactory, "by-factory", "f", "", "Only list devices belonging to this factory")
+	listCmd.Flags().StringVarP(&deviceByGroup, "by-group", "g", "", "Only list devices belonging to this group (by ID)")
```
doanac, line 37: I think we should just take the group name and not force the user to look this up every time.

doanac (approved):

PR #69, opened 2020-12-04
add device group management and assignment commands

vkhoroz:
A follow up from https://github.com/foundriesio/fioctl/pull/66
A GitHub doesn't allow to continue a pull request after a force push into a branch.

### DIFF
--- a/subcommands/config/device_group.go
+++ b/subcommands/config/device_group.go
@@ -1,0 +1,116 @@
+package config
+
+import (
+	"fmt"
+
+	"github.com/cheynewallace/tabby"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/foundriesio/fioctl/subcommands"
+)
+
+func init() {
+	groupCmd := &cobra.Command{
+		Use:   "device-group",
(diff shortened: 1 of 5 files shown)


subcommands/config/device_group.go
```diff
@@ -0,0 +1,128 @@
...
+	updateCmd.Flags().StringP("name", "n", "", "Change a device name")
+	updateCmd.Flags().StringP("description", "d", "", "Change a device description")
+}
+
+func doListDeviceGroup(cmd *cobra.Command, args []string) {
+	factory := viper.GetString("factory")
+	logrus.Debugf("Showing a list of device groups for %s", factory)
+
+	lst, err := api.FactoryListDeviceGroup(factory)
+	if err != nil {
```
doanac, line 54: Looks good. There's a new helper I did to deal with these commands exiting on err!=nil:

 https://github.com/foundriesio/fioctl/blob/master/subcommands/common.go#L159

that fixes a bunch of inconsistent ways this has been done. Including here where there's a space in "ERROR: "  - golang will do a space as well so you wind up with 2 spaces instead of one.

vkhoroz: Cool, used it

PR #71, opened 2020-12-10
Generalize factory and device config commands, add ability to manage device group config

vkhoroz:
There is still work to do with a non-trivial updates subcommand, but it's already possible to review config log/set/delete subcommands, all of which now behave consistently for factory/group/device config.

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -667,13 +667,16 @@ func (a *Api) FactoryDeleteConfig(factory, filename string) error {
 	return err
 }
 
-func (a *Api) FactoryPatchConfig(factory string, cfg ConfigCreateRequest) error {
+func (a *Api) FactoryPatchConfig(factory string, cfg ConfigCreateRequest, force bool) error {
 	data, err := json.Marshal(cfg)
 	if err != nil {
 		return err
 	}
 
 	url := a.serverUrl + "/ota/factories/" + factory + "/config/"
+	if force {
+		url += "?force=1"
+	}
 	logrus.Debug("Creating new factory config")
 	_, err = a.Patch(url, data)
 	return err
@@ -709,13 +712,16 @@ func (a *Api) GroupDeleteConfig(factory, group, filename string) error {
 	return err
 }
 
-func (a *Api) GroupPatchConfig(factory, group string, cfg ConfigCreateRequest) error {
+func (a *Api) GroupPatchConfig(factory, group string, cfg ConfigCreateRequest, force bool) error {
 	data, err := json.Marshal(cfg)
 	if err != nil {
 		return err
 	}
 
 	url := a.serverUrl + "/ota/factories/" + factory + "/device-groups/" + group + "/config/"
+	if force {
+		url += "?force=1"
+	}
 	logrus.Debug("Creating new device group config")
 	_, err = a.Patch(url, data)
 	return err
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -59,7 +59,7 @@ type ConfigCreateRequest struct {
 
 type DeviceConfig struct {
 	CreatedAt string       `json:"created-at"`
-	AppliedAt string       `json:"applied-at"`
+	AppliedAt string       `json:"applied-at"` // This is not present in factory config
 	Reason    string       `json:"reason"`
 	Files     []ConfigFile `json:"files"`
 }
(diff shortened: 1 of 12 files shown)


client/foundries.go
```diff
@@ -682,10 +682,53 @@ func (a *Api) FactoryPatchConfig(factory string, cfg ConfigCreateRequest) error
...
+	return a.DeviceListConfigCont(url)
+}
+
+func (a *Api) GroupCreateConfig(factory, group string, cfg ConfigCreateRequest) error {
+	data, err := json.Marshal(cfg)
+	if err != nil {
+		return err
+	}
+
+	url := a.serverUrl + "/ota/factories/" + factory + "/device-groups/" + group + "/config/"
```
mike-sul, line 699: So, a config filename is specified in the POST body, not in URL?

client/foundries.go
```diff
@@ -682,10 +682,53 @@ func (a *Api) FactoryPatchConfig(factory string, cfg ConfigCreateRequest) error
...
+}
+
+func (a *Api) GroupPatchConfig(factory, group string, cfg ConfigCreateRequest) error {
+	data, err := json.Marshal(cfg)
+	if err != nil {
+		return err
+	}
+
+	url := a.serverUrl + "/ota/factories/" + factory + "/device-groups/" + group + "/config/"
+	logrus.Debug("Creating new device group config")
```
mike-sul, line 719: `Patching` instead of `Creating`?

client/foundries.go
```diff
@@ -682,10 +682,53 @@ func (a *Api) FactoryPatchConfig(factory string, cfg ConfigCreateRequest) error
...
+	return err
+}
+
+func (a *Api) GroupPatchConfig(factory, group string, cfg ConfigCreateRequest) error {
+	data, err := json.Marshal(cfg)
+	if err != nil {
+		return err
+	}
+
+	url := a.serverUrl + "/ota/factories/" + factory + "/device-groups/" + group + "/config/"
```
mike-sul, line 718: Don't we need a filename of a config group that is being patched?

mike-sul:
I hardly understand the current API of device and factory configuration management, so it makes me an incompetent reviewer :).

client/foundries.go
vkhoroz: Yes

vkhoroz: and there could be multiple filenames at once

client/foundries.go
vkhoroz: A patch can apply to entire config which has many filenames

client/foundries.go
vkhoroz: I did the same as it is for factory/device configs

doanac (commented):
looks good. just a few minor things.

mike-sul:
How to list devices of a specific device group?

subcommands/config/log.go
```diff
@@ -12,32 +12,36 @@ import (
 func init() {
 	logCmd := &cobra.Command{
 		Use:   "log",
-		Short: "Changelog of configuration",
+		Short: "Show a change log of configuration",
```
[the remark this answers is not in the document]
vkhoroz: ok

subcommands/common_config.go
```diff
@@ -42,6 +80,21 @@ func LogConfigs(opts *LogConfigsOptions) {
 	}
 }
 
+func ReadConfig(configFile string, cfg *client.ConfigCreateRequest) {
```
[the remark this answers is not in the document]
vkhoroz: To me it's a mirror of `PrintConfig` which is already used outside.
A `ReadConfig` is only only used inside this module ATM, but I don't consider it to be private.

subcommands/common_config.go
```diff
@@ -127,3 +137,146 @@ func PrintConfig(cfg *client.DeviceConfig, showAppliedAt, highlightFirstLine boo
 		}
 	}
 }
+
+type SetUpdatesConfigOptions struct {
+	UpdateTags     string
```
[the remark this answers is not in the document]
vkhoroz: ok

vkhoroz:
> 
> 
> How to list devices of a specific device group?

This was already implemented in #67 .
`fioctl devices list -f <factory> -g <group>` or just a group if a factory goes from your config file.

subcommands/common_config.go
```diff
@@ -139,7 +139,7 @@ func PrintConfig(cfg *client.DeviceConfig, showAppliedAt, highlightFirstLine boo
 }
 
 type SetUpdatesConfigOptions struct {
-	UpdateTags     string
+	UpdateTag      string
```
doanac, line 142: sorry - i wasn't clear in my previous comment. this is kind of backwards. I was trying to say - make the CLI only accept "updateTag". The backend is still technically "tags". But lets change fioctl to only let people configure a single tag. 

If this is a pain, we can drop this commit and do it as a separate PR?

vkhoroz: Then I prefer to drop this commit.

A change you propose would be backward incompatible and might break some users expectations.
I'm not a fan to add a new logic and break an old logic within one feature.

It's not so trivial to be able to preserve multi-tag behavior for already present configs, and only disallow it for new configs.

doanac (approved):

PR #72, opened 2020-12-14
Add column customization to device listing

doanac:
A common feature request: More/different columns to be shown in device
listing. This keeps the default the same, but allows a user to show
any columns they'd like including Tags and DeviceGroup.

### DIFF
--- a/subcommands/devices/list.go
+++ b/subcommands/devices/list.go
@@ -2,6 +2,8 @@ package devices
 
 import (
 	"fmt"
+	"os"
+	"sort"
 	"strings"
 
 	"github.com/cheynewallace/tabby"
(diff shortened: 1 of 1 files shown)


doanac:
This was a customer request on Friday and @MatthewCroughan has been asking for this forever.

subcommands/devices/list.go
```diff
@@ -14,29 +17,111 @@ import (
...
 
+type column struct {
+	FieldName string
+	Formatter func(d *client.Device, fieldName string) string
+}
+
+func defFormatter(d *client.Device, fieldName string) string {
+	e := reflect.ValueOf(d).Elem()
+	val := e.FieldByName(fieldName).Interface()
+	if reflect.TypeOf(val).String() == "[]string" {
```
vkhoroz, line 36: It's better to do this via a type assertion: `if str, ok := val.([]string); ok {`; reflect module is "unsafe" in Golang and should be the last resort.

subcommands/devices/list.go
```diff
@@ -14,29 +17,111 @@ import (
...
+
+func groupFormatter(d *client.Device, fieldName string) string {
+	if d.Group != nil {
+		return d.Group.Name
+	}
+	return ""
+}
+
+var Columns = map[string]column{
+	"name":           {FieldName: "Name", Formatter: defFormatter},
```
vkhoroz, line 84: I'm a fan of Python reflection, but for Golang I'd prefer to find a way to solve a problem without it.
There are many pitfalls in Golang reflection e.g. there is no type checking and all sorts of mistakes would compile well, but panic later on.

One idea, which is almost as neat as what you've done is this:
```
type formatter func *client.Device) string
var Columns = map[string]formatter{
  "name": func (d *client.Device) string {return defFormatter(d.Name)},
  ...
}
```

vkhoroz: Note that you might save a some characters (if you care) with a simple type alias `type Device = client.Device`.

vkhoroz: Forgot to mention that this way a default formatter might accept an interface and do simple type assertion instead of reflection i.e. it becomes simpler.
At the same time more specialized formatters would be able to do a proper type checking via a function signature at compile time.

vkhoroz (commented):
LGTM, I'm just worried by using reflection which is unsafe in Golang... there is a way to avoid it (see my comments).

subcommands/devices/list.go
doanac: that's a better idea. i'll try and re-work it with that approach.

doanac: i just did a force-push cleanup. let me know what you think.

vkhoroz: Looks awesome

vkhoroz:
LGTM

PR #76, opened 2021-01-21
Clean up TUF key rotation

doanac:
This does two things:

 * clean up some over-designed/dead code
 * get key rotation working again


vkhoroz (approved):
LGTM

Thanks for doing this cleanup.

Issue #37, opened 2020-07-08
Use correct semantic versioning


MatthewCroughan:
I want to add the following information to the readme, it'd be nice if `go get` was the actual installation method, as this would allow people to update seamlessly as they would with an npm application or any other package manager. `go get` of course works on this repo, the problem is that the versions are labeled incompatibly with go packages which expect `v1.0` instead of simply `1.0` and all the release tags on this repo have no letter `v` behind their version number.

Here's our package. No versions are detected
https://pkg.go.dev/mod/github.com/foundriesio/fioctl?tab=versions

Whereas here's [mux](https://github.com/gorilla/mux) which is properly configured.
https://pkg.go.dev/github.com/gorilla/mux?tab=versions

```
## Installation

### Grab a [release binary](https://github.com/foundriesio/fioctl/releases)

Download a `go` executable for your OS from the [release page](https://github.com/foundriesio/fioctl/releases) 
and place the executable in a directory in your $PATH.

### Build from latest release via `go get`

~~~
`go get -u github.com/foundriesio/fioctl@latest`
~~~
```

doanac:
similar idea, but i think we should model the install more like what kubectl does. its more analogous to what we do than mux.

MatthewCroughan:
@doanac Yeah that's the good stuff. They really did give every possible method. `go get` is really building rather than an installation as you initially said, which is why a lot of Google's go packages do not suggest using `go get`. I suspect this also plugs into a lot of places that we're not yet considering based on the infrastructure of other Go applications, as well as the contribution guidelines for kubectl.

For example, the kubectl README states that your contributions must be:
> Go tools compliant (go get, go test, etc.). It needs to be vendorable somewhere else.

There must be other reasons to be compliant with the go get functionality other than just being able to shorthand reproduce builds via `go get fioctl@1.2` for example.

https://pkg.go.dev/k8s.io/kubernetes/pkg/kubectl?tab=doc

MatthewCroughan:
In addition to prepending v, we also need to adhere to semantic versioning. The latest release `v0.11` needs to be `v0.11.0` for example, otherwise it doesn't adhere to the go module publishing guidelines, and thus does not work with some elements of the toolchain. 

https://blog.golang.org/publishing-go-modules

MatthewCroughan:
This has been done inconsistently. 

v0.14.1 was correct
v0.8.1 was correct
v0.5.1 was correct

[...]

PR #78, opened 2021-01-25
Improve usability of device updates

doanac:

### DIFF
--- a/subcommands/devices/cmd.go
+++ b/subcommands/devices/cmd.go
@@ -27,12 +27,36 @@ var configCmd = &cobra.Command{
 }
 
 var updatesCmd = &cobra.Command{
-	Use:   "updates",
-	Short: "Device update history",
+	Use:   "updates <device> [<update-id>]",
+	Short: "Show updates performed on a device",
+	Args:  cobra.RangeArgs(1, 2),
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 1 {
+			doListUpdates(cmd, args)
+		} else {
+			doShowUpdate(cmd, args)
+		}
+	},
+	Example: `
+# List all updates performed on a device:
+fioctl devices updates <device1>
+
+# List the last 2 updates:
+fioctl devices updates <device> -n2
+
+# Show the details of an update:
+fioctl devices updates <device> <update-id>
+
+# Show the most recent update with bash help:
+fioctl devices updates <device> $(fioctl devices updates <device> -n1 | tail -n1 | cut -f1 -d\ )
+`,
 }
 
 func NewCommand() *cobra.Command {
 	cmd.PersistentFlags().StringP("token", "t", "", "API token from https://app.foundries.io/settings/tokens/")
+
+	updatesCmd.Flags().IntVarP(&listLimit, "limit", "n", 0, "Limit the number of updates displayed.")
+
 	cmd.AddCommand(configCmd)
 	cmd.AddCommand(updatesCmd)
 	return cmd
--- a/subcommands/devices/updates_list.go
+++ b/subcommands/devices/updates_list.go
@@ -11,10 +11,11 @@ import (
 
 func init() {
 	listUpdatesCmd := &cobra.Command{
-		Use:   "list <device>",
-		Short: "List the device's update history",
-		Run:   doListUpdates,
-		Args:  cobra.ExactArgs(1),
+		Use:    "list <device>",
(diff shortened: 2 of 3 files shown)


subcommands/devices/cmd.go
```diff
@@ -27,12 +27,36 @@ var configCmd = &cobra.Command{
...
+fioctl devices updates <device1>
+
+# List the last 2 updates:
+fioctl devices updates <device> -n2
+
+# Show the details of an update:
+fioctl devices updates <device> <update-id>
+
+# Show the most recent update with bash help:
+fioctl devices updates <device> $(fioctl devices updates <device> -n1 | tail -n1 | cut -f1 -d\ )
```
vkhoroz, line 51: I'd move this whole thing about updatesCmd into it's own module e.g. `updates.go`, so that this place doesn't become overcrowded.
You might even want to merge `updates_list.go` and `updates_show.go` into that module too to easily navigate things.

subcommands/devices/cmd.go
```diff
@@ -27,12 +27,36 @@ var configCmd = &cobra.Command{
...
+	},
+	Example: `
+# List all updates performed on a device:
+fioctl devices updates <device1>
+
+# List the last 2 updates:
+fioctl devices updates <device> -n2
+
+# Show the details of an update:
+fioctl devices updates <device> <update-id>
```
vkhoroz, line 48: From a compatibility perspective it'd be better to have `fioctl devices updates show <device> <update-id>`, as there is a small chance some angry user named their device as "list" or "show" and things will start breaking for them.
Though honestly that's more a funny comment than a practical one.
In general if we don't plan to add any sub-commands - this seems good.

vkhoroz (commented):
What about a deprecation notice in show/list sub-commands that they will be removed in one of the upcoming releases ?

MatthewCroughan:
I've wanted deprecation warnings for a while. Notifying users that they are on an outdated version of Fioctl by checking the latest release on GH would be good enough though. If we adhered to semantic versioning as noted in https://github.com/foundriesio/fioctl/issues/37 this would be trivial. Additionally fioctl would begin to work with `go get`.

[`go-latest`](https://github.com/tcnksm/go-latest) would let you write a typical "You are out of date" msg based on our releases if #37 is solved:

```
githubTag := &latest.GithubTag{
    Owner: "username",
    Repository: "reponame",
}

res, _ := latest.Check(githubTag, "0.1.0")
if res.Outdated {
    fmt.Printf("0.1.0 is not latest, you should upgrade to %s", res.Current)
}
```

subcommands/devices/cmd.go
doanac: there's nothing you can do to an update-event except view so subcommands for this just aren't useful.

subcommands/devices/cmd.go
doanac: that's easy enough as it gets crowded. no need yet. as we eventually deprecate, i agree that a single updates.go would be nice. for now i'm just wanting a simple improvment to this though.

doanac:
> I've wanted deprecation warnings for a while.

We've never deprecated anything before.

> What about a deprecation notice in show/list sub-commands that they will be removed in one of the upcoming releases ?

I was hoping to make this a simple change and not get drug into that conversation. I guess I should one more commit that:

 * mark show/list as "hidden" commands in cobra. then nobody sees them
 * add "preexec" or something to them that prints a deprecation warning.

MatthewCroughan:
@doanac 

> We've never deprecated anything before.

Is it not the case that we've avoided deprecation because of a lack of system to deprecate things? Have we never renamed things and had people be confused about where the cmd went/what was changed? 

What qualifies for a deprecation? In Python we see deprecation warnings (in our sphinx docs for example) when functions are renamed. When functions are "moved to a subcommand" I do not see how this is any different than renaming.

https://github.com/foundriesio/fioctl/commit/efe71853fa5ab54ff78bb9aa31701ce352e81463

doanac:
> Is it not the case that we've avoided deprecation because of a lack of system to deprecate things?

Since getting users on this tool, i don't believe we've changed things. 

> efe7185

That was before we had users. If we remove something, yes, we need to deprecate it.

MatthewCroughan:
> That was before we had users. 

Understood. I was not aware.

doanac:
k. this should be set now. I've hidden the sub-commands and marked them deprecated. I doubt anyone knows of them or has used them, but I'll wait a few months before removing.

vkhoroz:
LGTM

vkhoroz (approved):

Issue #77, opened 2021-01-25
fioctl config device-group update help changes


mike-scott:
When using `fioctl config device-group update` the help text references `device` and not `device-group`:

```
Flags:
  -d, --description string   Change a device description
  -h, --help                 help for update
  -n, --name string          Change a device name
```

vkhoroz:
Fixed by #80

PR #80, opened 2021-02-01
fix a device-group help message referring to device instead of device group

vkhoroz:
This fixed issue #77


mike-scott (approved):
LGTM!  Sorry, I was supposed to submit a PR to fix this and never got around to it.  Thank you for putting this up.

PR #82, opened 2021-02-09
Add commands to create, rollout, complete, cancel, list and view waves

vkhoroz:
The first draft so that I don't lose it

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -275,14 +275,21 @@ type TargetTestList struct {
 	Next  *string      `json:"next"`
 }
 
+type WaveRolloutGroupRef struct {
+	GroupId   int    `json:"group-id"`
+	GroupName string `json:"group-name"`
+	CreatedAt string `json:"created-at"`
+}
+
 type Wave struct {
-	Name       string           `json:"name"`
-	Version    string           `json:"version"`
-	Tag        string           `json:"tag"`
-	Targets    *json.RawMessage `json:"targets"`
-	CreatedAt  string           `json:"created-at"`
-	FinishedAt string           `json:"finished-at"`
-	Status     string           `json:"status"`
+	Name          string                         `json:"name"`
+	Version       string                         `json:"version"`
+	Tag           string                         `json:"tag"`
+	Targets       *json.RawMessage               `json:"targets"`
+	CreatedAt     string                         `json:"created-at"`
+	FinishedAt    string                         `json:"finished-at"`
+	Status        string                         `json:"status"`
+	RolloutGroups map[string]WaveRolloutGroupRef `json:"rollout-groups"`
 }
 
 type WaveCreate struct {
@@ -290,6 +297,10 @@ type WaveCreate struct {
 	Version string     `json:"version"`
 	Tag     string     `json:"tag"`
 	Targets tuf.Signed `json:"targets"`
+}
+
+type WaveRolloutOptions struct {
+	Group string `json:"group"`
 }
 
 // This is an error returned in case if we've successfully received an HTTP response which contains
@@ -1320,6 +1331,19 @@ func (a *Api) FactoryGetWave(factory string, wave string, showTargets bool) (*Wa
 	var resp Wave
 	err = json.Unmarshal(*body, &resp)
 	return &resp, err
+}
+
+func (a *Api) FactoryRolloutWave(factory string, wave string, options WaveRolloutOptions) error {
+	url := a.serverUrl + "/ota/factories/" + factory + "/waves/" + wave + "/rollout/"
+	logrus.Debugf("Rolling out factory wave %s", url)
+
+	data, err := json.Marshal(options)
+	if err != nil {
+		return err
+	}
+
+	_, err = a.Post(url, data)
+	return err
 }
 
 func (a *Api) FactoryCancelWave(factory string, wave string) error {
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -276,20 +276,20 @@ type TargetTestList struct {
 }
 
 type Wave struct {
-	Name       string            `json:"name"`
-	Version    string            `json:"version"`
-	Tag        string            `json:"tag"`
-	Targets    tuf.SignedTargets `json:"targets"`
-	CreatedAt  string            `json:"created-at"`
-	FinishedAt string            `json:"finished-at"`
-	Status     string            `json:"status"`
+	Name       string           `json:"name"`
+	Version    string           `json:"version"`
+	Tag        string           `json:"tag"`
+	Targets    *json.RawMessage `json:"targets"`
+	CreatedAt  string           `json:"created-at"`
+	FinishedAt string           `json:"finished-at"`
+	Status     string           `json:"status"`
 }
 
 type WaveCreate struct {
-	Name    string      `json:"name"`
-	Version string      `json:"version"`
-	Tag     string      `json:"tag"`
-	Targets tuf.Targets `json:"targets"`
+	Name    string     `json:"name"`
+	Version string     `json:"version"`
+	Tag     string     `json:"tag"`
+	Targets tuf.Signed `json:"targets"`
 }
 
 // This is an error returned in case if we've successfully received an HTTP response which contains
@@ -1278,7 +1278,7 @@ func (a *Api) FactoryPatchCA(factory string, certs CaCerts) error {
 	return err
 }
 
-func (a *Api) FactoryCreateWave(factory string, wave WaveCreate) error {
+func (a *Api) FactoryCreateWave(factory string, wave *WaveCreate) error {
 	url := a.serverUrl + "/ota/factories/" + factory + "/waves/"
 	logrus.Debugf("Creating factory wave %s", url)
 
@@ -1310,7 +1310,7 @@ func (a *Api) FactoryGetWave(factory string, wave string, showTargets bool) (*Wa
 	if showTargets {
 		url += "?show-targets=1"
 	}
-	logrus.Debugf("Viewing factory wave %s", url)
+	logrus.Debugf("Fetching factory wave %s", url)
 
 	body, err := a.Get(url)
 	if err != nil {
@@ -1320,4 +1320,46 @@ func (a *Api) FactoryGetWave(factory string, wave string, showTargets bool) (*Wa
 	var resp Wave
 	err = json.Unmarshal(*body, &resp)
 	return &resp, err
+}
+
+func (a *Api) FactoryCancelWave(factory string, wave string) error {
+	url := a.serverUrl + "/ota/factories/" + factory + "/waves/" + wave + "/cancel/"
+	logrus.Debugf("Canceling factory wave %s", url)
+	_, err := a.Post(url, nil)
+	return err
+}
+
+func (a *Api) FactoryCompleteWave(factory string, wave string) error {
+	url := a.serverUrl + "/ota/factories/" + factory + "/waves/" + wave + "/complete/"
+	logrus.Debugf("Completing factory wave %s", url)
+	_, err := a.Post(url, nil)
+	return err
+}
+
+func (a *Api) ProdTargetsList(factory string, tags ...string) (map[string]tuf.SignedTargets, error) {
+	url := a.serverUrl + "/ota/factories/" + factory + "/prod-targets/?tag=" + strings.Join(tags, ",")
+	logrus.Debugf("Fetching factory production targets %s", url)
+
+	body, err := a.Get(url)
+	if err != nil {
+		return nil, err
+	}
+
+	resp := make(map[string]tuf.SignedTargets)
+	err = json.Unmarshal(*body, &resp)
+	return resp, err
+}
+
+func (a *Api) ProdTargetsGet(factory string, tag string, failNotExist bool) (*tuf.SignedTargets, error) {
+	targets_map, err := a.ProdTargetsList(factory, tag)
+	if err != nil {
+		if !failNotExist {
+			if herr := AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
+				return nil, nil
+			}
+		}
+		return nil, err
+	}
+	targets := targets_map[tag]
+	return &targets, nil
 }
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -908,18 +908,22 @@ func (a *Api) TargetsListRaw(factory string) (*[]byte, error) {
 	return a.Get(url)
 }
 
-func (a *Api) TargetsList(factory string) (*tuf.SignedTargets, error) {
-	body, err := a.TargetsListRaw(factory)
+func (a *Api) TargetsList(factory string, version ...string) (tuf.Files, error) {
+	url := a.serverUrl + "/ota/factories/" + factory + "/targets/"
+	if len(version) == 1 {
+		url += "?version=" + version[0]
+	}
+	body, err := a.Get(url)
 	if err != nil {
 		return nil, err
 	}
-	targets := tuf.SignedTargets{}
+	targets := make(tuf.Files)
 	err = json.Unmarshal(*body, &targets)
 	if err != nil {
 		return nil, err
 	}
 
-	return &targets, nil
+	return targets, nil
 }
 
 func (a *Api) TargetCustom(target tuf.FileMeta) (*TufCustom, error) {
@@ -1255,4 +1259,48 @@ func (a *Api) FactoryPatchCA(factory string, certs CaCerts) error {
 
 	_, err = a.Patch(url, data)
 	return err
+}
+
+func (a *Api) FactoryCreateWave(factory string, wave WaveCreate) error {
+	url := a.serverUrl + "/ota/factories/" + factory + "/waves/"
+	logrus.Debugf("Creating factory wave %s", url)
+
+	data, err := json.Marshal(wave)
+	if err != nil {
+		return err
+	}
+
+	_, err = a.Post(url, data)
+	return err
+}
+
+func (a *Api) FactoryListWaves(factory string) ([]Wave, error) {
+	url := a.serverUrl + "/ota/factories/" + factory + "/waves/"
+	logrus.Debugf("Listing factory waves %s", url)
+
+	body, err := a.Get(url)
+	if err != nil {
+		return nil, err
+	}
+
+	var resp []Wave
+	err = json.Unmarshal(*body, &resp)
+	return resp, err
+}
+
+func (a *Api) FactoryGetWave(factory string, wave string, showTargets bool) (*Wave, error) {
+	url := a.serverUrl + "/ota/factories/" + factory + "/waves/" + wave + "/"
+	if showTargets {
+		url += "?show-targets=1"
+	}
+	logrus.Debugf("Viewing factory wave %s", url)
+
+	body, err := a.Get(url)
+	if err != nil {
+		return nil, err
+	}
+
+	var resp Wave
+	err = json.Unmarshal(*body, &resp)
+	return &resp, err
 }
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -275,6 +275,23 @@ type TargetTestList struct {
 	Next  *string      `json:"next"`
 }
 
+type Wave struct {
+	Name       string            `json:"name"`
+	Version    string            `json:"version"`
+	Tag        string            `json:"tag"`
+	Targets    tuf.SignedTargets `json:"targets"`
+	CreatedAt  string            `json:"created-at"`
+	FinishedAt string            `json:"finished-at"`
+	Status     string            `json:"status"`
+}
+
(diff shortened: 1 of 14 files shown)


doanac:
seems sane so far

subcommands/targets/edit.go
```diff
@@ -36,7 +36,7 @@ func doEdit(cmd *cobra.Command, args []string) {
 	// Get raw json
 	targets, err := api.TargetsList(factory)
 	subcommands.DieNotNil(err)
-	orig, err := json.MarshalIndent(targets.Signed.Targets, "", "  ")
+	orig, err := json.MarshalIndent(targets, "", "  ")
```
doanac, line 39: that's not going to work. I think `TargetList` was returning the exact targets content. Now its returning your representation of it? The edit command needs the *exact* content so that it can edit things correctly.

doanac:
I've read up to cc778dd and things seem on the right track so far.

subcommands/targets/edit.go
vkhoroz: I'm not sure I understand what does _exact content_ mean in this context.
Before my change it was returning _[]tuf.Files_ wrapped into a _SignedTargets_, not it returns unwrapped _[]tuf.Files_, which is the same as before, there is no such thing as _my representation_ in this command.

There is _my representation_ for waves, but it's an internal struct inside a function, so that nobody can use it anywhere else.

vkhoroz: I use a slightly different representation of signed targets for waves (because of an issue with delegations), but even there it's still the same _[]tuf.Files_;  moreover, that's an inner struct inside a _doCreateWave_ function i.e. cannot be used elsewhere.

doanac: the old TargetsList was calling:

 https://api.foundries.io/ota/repo/lmp/api/v1/user_repo/targets.json

Your change has it calling:

  https://api.foundries.io/ota/factories/lmp/targets/

Are these really the *exact* same data? I thought your new view had small differences?

vkhoroz: The only difference is in a wrapping envelope:
- old data was like this:
  `signatures: <signatures>
   signed:
      _type: Targets
      expires: <expires>
      targets: <targets-list>
      version: <version>
  `
- new data is like this:
  `<targets-list>`
But it is the same `<targets-list>` in both cases, and `fioctl targets edit` doesn't use the other data, it only uses a list of targets.  To what I know at least a canonical form of them must be _exact_ same data.

doanac: ack. thanks

client/foundries.go
```diff
@@ -845,18 +862,22 @@ func (a *Api) TargetsListRaw(factory string) (*[]byte, error) {
 	return a.Get(url)
 }
 
-func (a *Api) TargetsList(factory string) (*tuf.SignedTargets, error) {
-	body, err := a.TargetsListRaw(factory)
+func (a *Api) TargetsList(factory string, version ...string) (tuf.Files, error) {
```
mike-sul, line 865: Interesting way to handle an optional parameter in Go :)

subcommands/waves/init.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// With tuf.Targets there is a problem: it sneaks a delegations field into a resulting JSON.
+	// As we don't want that field, a workaround is to use our own simplified type.
+	type TufTargets struct {
+		tuf.SignedCommon
+		Targets tuf.Files `json:"targets"`
+	}
+	targets := TufTargets{}
+	targets.Type = tuf.TUFTypes["targets"]
+	targets.Version += 1
```
mike-sul, line 70: I am not sure if I understand this `targets.Version += 1`,  it looks like targets.Version will always be equal `1`, how this field `targets.Version` is initialized?

client/foundries.go
```diff
@@ -242,6 +242,23 @@ type TargetTestList struct {
...
+	CreatedAt  string           `json:"created-at"`
+	FinishedAt string           `json:"finished-at"`
+	Status     string           `json:"status"`
+}
+
+type WaveCreate struct {
+	Name    string     `json:"name"`
+	Version string     `json:"version"`
+	Tag     string     `json:"tag"`
+	Targets tuf.Signed `json:"targets"`
```
mike-sul, line 259: So, we actually passing an overall target.json from a client to the backend, I remember talking about sending just signature?

subcommands/waves/init.go
vkhoroz: You're right, in the initial implementation it'll always be 1 (because this logic only creates the first wave).  My follow-up changes which implement promoted targets should increase a version of promoted targets instead (if present), or fall-back to a current logic if there are no promoted targets yet.

client/foundries.go
vkhoroz: I'd say, it's standard :)

client/foundries.go
vkhoroz: I've reconsidered this during a discussion with @doanac .
Yes, we send entire JSON as it allows to keep this logic of constructing a `targets.json` in a single place (client side) that's good from both consistency and performance points of view, and also allows to keep an implementation simple.

subcommands/waves/init.go
```diff
@@ -0,0 +1,132 @@
...
+	name, version, tag := args[0], args[1], "master"
+	if len(args) > 2 {
+		tag = args[2]
+	}
+	expires := readExpiration(cmd)
+	dryRun, _ := cmd.Flags().GetBool("dry-run")
+	logrus.Debugf("Creating a wave %s for factory %s targets version %s and new tag %s expires %s",
+		name, factory, version, tag, expires.Format(time.RFC3339))
+
+	new_targets, err := api.TargetsList(factory, version)
```
doanac, line 53: this is WIP, but as you progress the list of targets will need to be something like `GetCurrentProductionTargets + TargetsList(factory, version)`.

vkhoroz: Yes, sure, see my comment below.

vkhoroz:
This version was tested and works end-to-end.
So, it's time to make a thorough code review now.

vkhoroz:
@doanac please, pay attention to 0259c58, an aktualizr doesn't work well with Golang's default time.Time formatting.
You might see the same issue in your root role updates work too.

doanac:
Some really high level comments after playing with the usage options from my xterm:

* discard - That word would probably scare me. It feels like its going to delete the wave. I think maybe `fioctl wave stop` and then we'll eventually need to provide a good description of what this means.
* promote - That word probably won't mean much to a user. It looks like you'll have a `fioctl wave rollout` command add soon. Maybe that command can just take a single `--all` type switch. Then all the rollout logic is in one command

And maybe "discard/stop" should just be in the `rollout` command as well. eg:
~~~
 fioctl wave init hope-this-works 42 master
 fioctl wave rollout hope-this-works us-east-1
 fioctl wave rollout hope-this-works --complete
 fioctl wave rollout hope-this-works --cancel
~~~

client/foundries.go
```diff
@@ -1193,3 +1214,89 @@ func (a *Api) FactoryPatchCA(factory string, certs CaCerts) error {
...
+}
+
+func (a *Api) FactoryListWaves(factory string) ([]Wave, error) {
+	url := a.serverUrl + "/ota/factories/" + factory + "/waves/"
+	logrus.Debugf("Listing factory waves %s", url)
+
+	body, err := a.Get(url)
+	if err != nil {
+		return nil, err
+	}
```
doanac, line 1238: I'm wondering if we should make this a paginated API?

client/foundries.go
```diff
@@ -1193,3 +1214,89 @@ func (a *Api) FactoryPatchCA(factory string, certs CaCerts) error {
...
+
+func (a *Api) FactoryPromoteWave(factory string, wave string) error {
+	url := a.serverUrl + "/ota/factories/" + factory + "/waves/" + wave + "/promote/"
+	logrus.Debugf("Promoting factory wave %s", url)
+	_, err := a.Post(url, nil)
+	return err
+}
+
+func (a *Api) PromotedTargetsList(factory string, tags ...string) (map[string]tuf.SignedTargets, error) {
+	url := a.serverUrl + "/ota/factories/" + factory + "/promoted-targets/?tag=" + strings.Join(tags, ",")
```
doanac, line 1277: We could probably use the current targets API and do `<factory>/targets/?tag=<tag>&production=1`.

subcommands/waves/init.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/foundriesio/fioctl/subcommands"
+	tuf "github.com/theupdateframework/notary/tuf/data"
+)
+
+func init() {
+	initCmd := &cobra.Command{
+		Use:   "init <wave> <version> [<tag>]",
+		Short: "Create a new wave for a given targets version.",
+		Long: `Create a new wave for a given targets version.",
+Optionally, provide a new tag to use for these targets ('master' by default).
```
doanac, line 23: s/a new tag/the tag/

subcommands/waves/init.go
```diff
@@ -0,0 +1,160 @@
...
+
+This command only initializes a wave, but does not provision its updates to devices.
+Use a "fioctl wave rollout <wave> <group>" to trigger updates of this wave to one group.
+Use a "fioctl wave promote <wave>" to update all devices (make it globally available).`,
+		Run:  doInitWave,
+		Args: cobra.RangeArgs(2, 3),
+	}
+	cmd.AddCommand(initCmd)
+	initCmd.Flags().IntP("expires-days", "e", 0, `Role expiration in days; default 90.
+The same expiration will be used when a wave is promoted.`)
```
doanac, line 34: @ricardosalveti - ATS tooling does 1 year for the target expiration. The notary tooling seems a little more conservative. What do you think?

subcommands/waves/init.go
```diff
@@ -0,0 +1,160 @@
...
+				"Cannot create a wave for a version lower than current promoted targets"))
+		}
+	}
+
+	// Add new targets, replacing tags in each of them
+	for name, file := range new_targets {
+		// There is a dilemma what to do if a specific target already exists in current promoted
+		// targets for a given tag.  A solution below would fail in this scenario as there is no point
+		// to start a new wave for something that is already generally available in production.
+		// Under normal circumstances a target, once created, is immutable making this approach sane.
```
doanac, line 99: I think this is so obvious that the comment isn't that useful. The error message shown to the user is enough to guide the person reading the code.

doanac (commented):
looks good still. just some minor stuff.

client/foundries.go
vkhoroz: We don't have a pagination for targets API, and we're going to have less waves than targets.
Thus I thought it's not worth to introduce a pagination into this API.

Though we might make a `fioctl` command paginated... I'm inclined to a `git log -n N` mode which shows only N recent waves, we could set it to 20 by default and use `--all` to show all waves.  An API can stay without pagination until we have a customer creating 1000 waves.

client/foundries.go
vkhoroz: A current Targets API returns a list of individual CI targets, whereas this API returns a map of production "targets.json" files i.e. they are not compatible.

subcommands/waves/init.go
vkhoroz: ok

client/foundries.go
doanac: targets is hard to paginate since its format is dictated by TUF. Its likely that over time you'll have more waves than targets. Old targets will get pruned so that the production targets is only ever a few versions. 

We do the "-n" type thing for `config log`.  

I don't like the complexity pagination adds, but it does limit the damage that can be caused in a single API call. And its hard to "backport" pagination.

client/foundries.go
doanac: Do we really need to return this map? We are going to be based off a single tag. Feels like we are trying to optimize for a scenario we don't want to encourage/support?

subcommands/waves/init.go
vkhoroz: I was waiting for this comment, cause I'm not sure on this default expiration interval.

Currently, I've found the following expiration policies:
1. `theupdateframework/notary/const.go` used by a notary server defaults are: root - 10 years, targets - 3 years, snapshot - 3 years, timestamp - 14 days.
2. `theupdateframework/notary/tuf/data/types.go` recommended TUF defaults are: root - 1 year, targets - 90 days, snapshot - 7 days, timestamp - 1 day.
3. `advancedtelematic/ota-tuf/.../libtuf_server/repo/server/SignedRoleGeneration.scala` uses a 31 days default for targets, snapshot, and timestamp roles (keeps them in sync).
4. Current ota-lite implementation for condensed targets uses 30 days expiration for targets, and then via tuf_reposerver the same expiration is set for snapshot/timestamp roles.

My thinking on this is the following:
Clearly, notary server many-years expiration policy is way too conservative.  Still, a one month expiration in use for CI targets is too aggressive for production, where release cycles tend to take a longer time.  Thus, I voluntarily used a 90 days expiration recommended by TUF for targets role, though I'm thinking of a longer default, like 1 year.
A problem I see with "too short" expiration is that a user would need to re-sign production targets in order to keep them valid in case if their release cycle is longer than an expiration interval.  I think we can add a `fioctl targets sign` command for this, but that simply requires too many moves from a customer side.

Still, for snapshot/timestamp roles I'd use a rather aggressive expiration policy in production e.g. 2-4 weeks for a snapshot roles and 3-7 days for a timetamp role.  Still, I find a 1 day expiration for timestamp too aggressive, because in case of a re-signer daemon failure we might end up with invalidated timestamp roles for all customers within a day, that would be a maintenance nightmare.

That said, my current plan is this: 90 days expiration for targets, 14 days for snapshot, 3 days for timestamp;   though I may easily change my mind to be more conservative like this: 1 year for targets, 30 days for snapshot, 7 days for timestamp.

@doanac @ricardosalveti @mike-sul what do you think ?

subcommands/waves/init.go
vkhoroz: Yes, sure... at a time I wrote this comment a code wasn't obvious, but now it is.

client/foundries.go
vkhoroz: You would need this API when doing an offline targets key rotation.  In that case we need to re-sign all production targets and it's best to fetch all of them in a single shot.
I don't want to create two separate APIs for a wave versus key rotation use cases where a single API is easier to implement.
On client side a trivial `PromotedTargetsGet` function adapts that API to a single tag.

client/foundries.go
vkhoroz: Will add an ability to limit a number of returned rows; but probably not a pagination (i.e. just limit, but not offset).
I don't see a use case when we'd need an offset for waves, whereas it adds up to complexity.

doanac: don't bother. either paginate or do nothing. something in between isn't worth the extra code. the case for pagination is when you've done something like 100 waves. Its a lot to return in a single API call.

client/foundries.go
doanac: I'm just getting concerned about API confusion. We already have two APIs for getting targets. Now we are adding a 3rd API because it might be a little quicker on the client side of a command that might get run once a month?

This also affects a command like `fioctl targets list` which would need a totally separate code path for the `--prod` option.

I do realize on the backend its roughly the same code so I might be overthinking.

client/foundries.go
vkhoroz: Done in a `device log` fashion; imho that's enough

subcommands/waves/init.go
vkhoroz: Resolved offline:
Will use 1 year default for targets expiration (configurable), snapshot expiration equal to targets, timestamp expiration 7 days.

vkhoroz:
Looks good to me, @doanac @mike-sul please approve or give comments

subcommands/waves/init.go
```diff
@@ -0,0 +1,156 @@
...
+	subcommands.DieNotNil(err, "Version must be an integer")
+	expires := readExpiration(cmd)
+	dryRun, _ := cmd.Flags().GetBool("dry-run")
+	logrus.Debugf("Creating a wave %s for factory %s targets version %s and new tag %s expires %s",
+		name, factory, version, tag, expires.Format(time.RFC3339))
+
+	new_targets, err := api.TargetsList(factory, version)
+	subcommands.DieNotNil(err)
+	if len(new_targets) == 0 {
+		subcommands.DieNotNil(fmt.Errorf("No new targets of a given version"))
```
doanac, line 59: a little nitpick, but i just hit this by accident and found this message confusing. I think something like "Target version %d does not exist".

doanac:
This is growing out of control. I think we more or less had create/cancel/list/view done. Lets get that merged here, and then do the "rollout group" in a new PR.

At a minimum, could you squash the fixup commits to clean up the history on this? It would be really nice if you could re-shuffle this around so that the final product is a few clean commits.

vkhoroz:
@doanac Will try to reshuffle to get rid of sime small commits, where it's possible.
Though I really don't want to do rollout as a separate patch set.

A current PR is a complete feature which seems quite organic.

vkhoroz:
I rebased this with master and made a major cleanup of commits, so now a history should look sane.

@doanac @mike-sul feel free to make a final review and I think it's ready to be merged.

subcommands/waves/cmd.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/foundriesio/fioctl/client"
+	"github.com/foundriesio/fioctl/subcommands"
+)
+
+var api *client.Api
+
+var cmd = &cobra.Command{
+	Use:   "wave",
+	Short: "Manage factory's waves",
```
doanac, line 14: Capitalize the F in factories.

subcommands/waves/show.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/foundriesio/fioctl/subcommands"
+)
+
+func init() {
+	showCmd := &cobra.Command{
+		Use:   "show",
+		Short: "Show Ra given wave",
```
doanac: "Show a given way"?

subcommands/waves/init.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/foundriesio/fioctl/client"
+	"github.com/foundriesio/fioctl/subcommands"
+	tuf "github.com/theupdateframework/notary/tuf/data"
+)
+
+func init() {
+	initCmd := &cobra.Command{
+		Use:   "init <wave> <version> [<tag>]",
+		Short: "Create a new wave for a given targets version.",
```
doanac, line 21: "for a given Target's version"

doanac (commented):
i just found a couple of grammar things. Just fix and squash, and this is good to merge.

subcommands/waves/show.go
vkhoroz: OMG. this looks like a bad squash... it went really hard

subcommands/waves/init.go
vkhoroz: ack

subcommands/waves/cmd.go
vkhoroz: ack

subcommands/waves/show.go
vkhoroz: well no, it seems like you're looking at some old commit
the latest version has this correct

subcommands/waves/cmd.go
vkhoroz: hmm... we have it lowercase in other places:
```
Available Commands:
  config      Manage configuration common to all devices in a factory
  devices     Manage devices registered to a factory
  help        Help about any command
  keys        Manage keys in use by your factory fleet
  login       Access Foundries.io services with your client credentials
  secrets     Manage secret crendentials configured in a factory
  status      Get dashboard view of a factory and its devices
  targets     Manage factory's TUF targets
  users       List users with access to a factory
  version     Show version information of this tool.
  wave        Manage factory's waves
```

subcommands/waves/show.go
doanac: np.

subcommands/waves/cmd.go
doanac: Ugh. keep it like this for now. the first usage statement i looked at was different. so there's inconsistency that has nothing to do with you.

subcommands/waves/init.go
vkhoroz: After some thought either way seems confusing.
I rephrased this to be more clear: `Create a new wave from targets of a given version.`

vkhoroz: https://github.com/foundriesio/fioctl/pull/82/commits/e03c23a3501c7d4b5462266d00b37f28460de38a

doanac: that's much better.

PR #83, opened 2021-02-17
Move to golang implementation of root key management

doanac:


client/foundries.go
```diff
@@ -204,6 +204,28 @@ type TufCustom struct {
...
+type AtsKeyVal struct {
+	Public  string `json:"public,omitempty"`
+	Private string `json:"private,omitempty"`
+}
+type AtsKey struct {
+	KeyType  string    `json:"keytype"`
+	KeyValue AtsKeyVal `json:"keyval"`
+}
+type AtsRootMeta struct {
+	tuf.SignedCommon
```
vkhoroz, line 219: Unfortunately, there is an `expires time.Time` field which Golang marshals in a format not understood by aktualizr.
Please, see how I overcome this in my waves PR and apply here too.

subcommands/keys/rotate_root.go
```diff
@@ -0,0 +1,290 @@
...
+	subcommands.DieNotNil(err)
+
+	pubBytes, err := x509.MarshalPKIXPublicKey(&pk.PublicKey)
+	subcommands.DieNotNil(err)
+
+	block = &pem.Block{
+		Type:  "PUBLIC KEY",
+		Bytes: pubBytes,
+	}
+	id := fmt.Sprintf("%x", sha256.Sum256(privBytes))
```
vkhoroz, line 172: Can you put a comment why we generate an ID from privBytes instead of a canonical form of a AtsKey (as prescribed by a TUF spec) ?
Also, did you verify that this is really what aktualizr requires ?

vkhoroz: An again, generating from privBytes is a security hole, it should be pubBytes in any case.

client/foundries.go
doanac: This works because this line sets the time properly: https://github.com/foundriesio/fioctl/pull/83/files#diff-9b655247cb8326727517d0aeca1b07126f2602091675507157316aa1b4d1ac95R177

subcommands/keys/rotate_root.go
doanac: it was just missed when reshuffling. https://github.com/foundriesio/fioctl/pull/83/commits/c8e1ffb3e85fcee03282bad2f65058c5d76b1075

client/foundries.go
vkhoroz: Do you mean that if we round a time to a second (strip microseconds) - Golang will format this properly ?
I that's true - I will to use it as well in my PR.

subcommands/keys/rotate_root.go
vkhoroz: ok, but why do we create an ID off a key, and not a key JSON (as TUF spec advices)?

What I mean is this:
```
	pubKey = client.AtsKey{
		KeyType:  "RSA",
		KeyValue: client.AtsKeyVal{Public: string(pem.EncodeToMemory(block))},
	}
        // In your code below two lines are swapped.
        pubBytes, err = json.Marshal(pubKey)
	id := fmt.Sprintf("%x", sha256.Sum256(pubBytes))
        root.Signed.Keys[id] = pubKey
```

Also, I guess we need to use canonical, not a simple json here.

vkhoroz (approved):

subcommands/keys/rotate_root.go
doanac: i've just done it because aklite and ota-reposerver don't. I've pushed https://github.com/foundriesio/fioctl/pull/83/commits/9bfd1e0b65610bea959256436a9c89eeed1f134e .

PR #84, opened 2021-02-21
keys: Add a rotate-targets command

doanac:
This allows a user to have their own private key capable of signing
targets.json

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -873,6 +873,17 @@ func (a *Api) FactoryListDeviceGroup(factory string) (*[]DeviceGroup, error) {
 	return &resp.Groups, nil
 }
 
+func (a *Api) GetFoundriesTargetsKey(factory string) (*AtsKey, error) {
+	url := a.serverUrl + "/ota/factories/" + factory + "/ci-targets.pub"
+	body, err := a.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	key := AtsKey{}
+	err = json.Unmarshal(*body, &key)
+	return &key, err
+}
+
 func (a *Api) TufRootGet(factory string) (*AtsTufRoot, error) {
 	url := a.serverUrl + "/ota/repo/" + factory + "/api/v1/user_repo/root.json"
 	body, err := a.Get(url)
--- a/subcommands/keys/rotate_targets.go
+++ b/subcommands/keys/rotate_targets.go
@@ -1,12 +1,7 @@
 package keys
 
 import (
-	"crypto/rand"
-	"crypto/rsa"
-	"crypto/sha256"
-	"crypto/x509"
 	"encoding/json"
-	"encoding/pem"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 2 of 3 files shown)


doanac:
This is really a single commit based on:

 https://github.com/foundriesio/fioctl/pull/83

It requires a new backend resource - https://git.foundries.io/development/cloudplatforms/infrastructure/ota-lite/-/merge_requests/104

client/foundries.go
```diff
@@ -862,6 +862,17 @@ func (a *Api) FactoryListDeviceGroup(factory string) (*[]DeviceGroup, error) {
 	return &resp.Groups, nil
 }
 
+func (a *Api) CiTargets(factory string) (*AtsKey, error) {
```
vkhoroz, line 865: Can you rename this into smth like `GetOnlineTargetsKey` ?
As is I constantly have a feeling that you fetch targets when calling this method.

subcommands/keys/rotate_targets.go
```diff
@@ -0,0 +1,129 @@
...
+	block = &pem.Block{
+		Type:  "PUBLIC KEY",
+		Bytes: pubBytes,
+	}
+	pubBytes = pem.EncodeToMemory(block)
+	id := fmt.Sprintf("%x", sha256.Sum256(pubBytes))
+	root.Signed.Keys[id] = client.AtsKey{
+		KeyType:  "RSA",
+		KeyValue: client.AtsKeyVal{Public: string(pubBytes)},
+	}
```
vkhoroz, line 117: Around 20 lines from above are common with root key rotation;  and I already see some tiny differences (well, discrepancies).  I think they deserve to be extracted into a separate function.

client/foundries.go
doanac: https://github.com/foundriesio/fioctl/pull/84/commits/9b5c948da0ef274d59c51c22bc3173f21bf18046

subcommands/keys/rotate_targets.go
doanac: not that pretty, but consolidated - https://github.com/foundriesio/fioctl/pull/84/commits/14e292e1b25fbf6a44b4db98b4ffac4e012e5a3b

once reviewed, i'll squash that into the prior commit.

subcommands/keys/rotate_targets.go
```diff
@@ -0,0 +1,95 @@
...
+	for _, keyid := range root.Signed.Roles["targets"].KeyIDs {
+		pub := root.Signed.Keys[keyid].KeyValue.Public
+		if pub == onlinePub.KeyValue.Public {
+			return keyid, nil
+		}
+	}
+	return "", errors.New("Unable to find online target key for factory")
+}
+
+func replaceProdTargetKey(root *client.AtsTufRoot, onlineTargetId string, creds OfflineCreds) (string, OfflineCreds) {
```
vkhoroz, line 84: Can you replace Prod with Offline everywhere ?
This patch set doesn't even create prod root role - it's CI root role, which has online and offline targets keys.

I'm scary we might overload a word "prod" rather quickly if we put it in every place which has something to do with production environment.

vkhoroz (approved):

subcommands/keys/rotate_targets.go
vkhoroz: I like how it looks, even better than I thought ;-)

vkhoroz:
LGTM

subcommands/keys/rotate_targets.go
doanac: good point: https://github.com/foundriesio/fioctl/pull/84/commits/26c454a725366fb61d10a83d015f7d95d6966a86

if that looks good, i'll squash fix ups and merge

vkhoroz: looks good

doanac:
squash fixed everything.

PR #87, opened 2021-02-26
target artifacts: Give better error for invalid target

doanac:


mwasilew:
@doanac works well
```
$ ./bin/fioctl-linux-amd64 targets artifacts list
ERROR: Invalid Target number: list
```

PR #89, opened 2021-03-11
keys: Add logic for syncing production root

doanac:
A production root is the *exact* same as the root.json used in CI except
is has a targets signing threshold of 2.

### DIFF
--- a/subcommands/keys/rotate_root.go
+++ b/subcommands/keys/rotate_root.go
@@ -31,6 +31,8 @@ type TufSigner struct {
 	Key *rsa.PrivateKey
 }
 
+var doRootSync bool
+
 func init() {
 	rotate := &cobra.Command{
 		Use:     "rotate-root <offline key archive>",
@@ -40,6 +42,7 @@ func init() {
 		Args:    cobra.ExactArgs(1),
 	}
 	subcommands.RequireFactory(rotate)
+	rotate.Flags().BoolVarP(&doRootSync, "sync-prod", "", false, "Make sure production root.json is up-to-date and exit")
 	cmd.AddCommand(rotate)
 }
 
@@ -51,6 +54,11 @@ func doRotateRoot(cmd *cobra.Command, args []string) {
 
 	root, err := api.TufRootGet(factory)
 	subcommands.DieNotNil(err)
+
+	if doRootSync {
+		subcommands.DieNotNil(syncProdRoot(factory, *root, creds))
+		return
+	}
 
 	curid, curPk, err := findRoot(*root, creds)
 	fmt.Println("= Current root:", curid)
@@ -68,6 +76,7 @@ func doRotateRoot(cmd *cobra.Command, args []string) {
 		{Id: curid, Key: curPk},
 		{Id: newid, Key: newPk},
 	}
+	removeUnusedKeys(root)
 	subcommands.DieNotNil(signRoot(root, signers...))
 
 	tufRootPost(factory, credsFile, root, newCreds)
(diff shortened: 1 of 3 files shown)


subcommands/keys/rotate_root.go
```diff
@@ -52,6 +55,11 @@ func doRotateRoot(cmd *cobra.Command, args []string) {
 	root, err := api.TufRootGet(factory)
 	subcommands.DieNotNil(err)
 
+	if doRootSync {
+		subcommands.DieNotNil(syncProdRoot(factory, *root, creds))
+		return
+	}
```
vkhoroz, line 61: As I understand an API doesn't allow to rotate CI root if a prod root is out of sync (that's good).

So, when we handle that 409 status code - I'd like to show a message to a user which shows instructions how to get around that problem.
Smth like `Your production root role is out of sync.  To fix this, please, run "fioctl rotate root --sync-prod"`

subcommands/keys/rotate_root.go
```diff
@@ -314,3 +324,45 @@ func getOfflineCreds(credsFile string) OfflineCreds {
...
+	if err := signRoot(&root, signers...); err != nil {
+		return err
+	}
+
+	bytes, err := json.Marshal(root)
+	if err != nil {
+		return err
+	}
+	_, err = api.TufProdRootPost(factory, bytes)
```
vkhoroz, line 363: When checking a validation logic on API side I've found that we expect prod_root_version <= ci_root_version... but shouldn't we expect them to be equal ?
https://git.foundries.io/development/cloudplatforms/infrastructure/ota-lite/-/blob/master/ota_lite/api/reposerver.py#L262

I'm sorry that I didn't find it during an ota-lite code review.

Also, in case if a prod root role with the same version already exists - I'd expect a 409 status code (currently it returns 400 as I understand).

vkhoroz (approved):
LGTM

The only important comment goes to an ota-lite side which I somehow missed.

subcommands/keys/rotate_root.go
doanac: > we expect prod_root_version <= ci_root_version... but shouldn't we expect them to be equal ?

good catch. that's left over logic from the original backfill approach.

> we expect prod_root_version <= ci_root_version... but shouldn't we expect them to be equal ?

sounds reasonable.

subcommands/keys/rotate_root.go
```diff
@@ -31,6 +31,8 @@ type TufSigner struct {
 	Key *rsa.PrivateKey
 }
 
+var doRootSync bool
```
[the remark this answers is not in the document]
doanac: the original thinking was that by doing this way you catch a certain class of problems at compile time such as doing
~~~
syncProd, _ := cmd.Flags().GetString("sync-prod")
~~~

There's so much of it at this point that i've just stuck to making all the code look the same. The real thing is that cobra isn't the easiest CLI tool. Complex CLIs really start to hit a bunch of friction I'm finding.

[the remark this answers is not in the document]
vkhoroz: I'm fine either way

subcommands/keys/rotate_root.go
doanac: https://git.foundries.io/development/cloudplatforms/infrastructure/ota-lite/-/merge_requests/109

subcommands/keys/rotate_root.go
doanac: i'm working on this now. I actually need to consolidate a bit of replicated logic. should have it in the next day as i catch up

doanac: I've just done a force push to change this some. 
1) consolidate logic for updating root - https://github.com/foundriesio/fioctl/pull/89/commits/fb92c726b046137d8db2dadcbdbe1a82884b63b3
2) the original PR - https://github.com/foundriesio/fioctl/pull/89/commits/c0f32263a693e6b3baed8b66cf855c470eaefd02
3) the fix you asked for here - https://github.com/foundriesio/fioctl/pull/89/commits/67fcd9e70660e722f958dcb8fe4fc02716498561

I can squash 3 in with 2 if you think this looks good.

vkhoroz (approved):
Looks nice

PR #93, opened 2021-03-24
Change pagination logic of device-list

doanac:
As we grow in the number of devices inside factories, our logic to
show everything in one pass breaks down. The backend now gives us 200
devices at a time. In the case there are more, the "-p <page>" argument
can be added to work through the device. Or, better yet, use filtering
logic.

### DIFF
--- a/subcommands/devices/list.go
+++ b/subcommands/devices/list.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/cheynewallace/tabby"
@@ -22,6 +23,9 @@ var (
 	deviceInactiveHours int
 	deviceUuid          string
 	showColumns         []string
+	showPage            int
+	paginationLimit     int
+	paginationLimits    []int
 )
 
 type column struct {
@@ -84,6 +88,14 @@ func init() {
 	var defCols = []string{
 		"name", "factory", "target", "status", "apps", "up-to-date", "is-prod",
 	}
+	paginationLimits = []int{10, 20, 30, 40, 50, 100, 200, 500, 1000}
+	limitsStr := ""
+	for i, limit := range paginationLimits {
+		if i > 0 {
+			limitsStr += ","
+		}
+		limitsStr += strconv.Itoa(limit)
+	}
 
 	allCols := make([]string, 0, len(Columns))
 	for k := range Columns {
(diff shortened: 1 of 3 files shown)


subcommands/devices/list.go
```diff
@@ -137,40 +139,48 @@ func doList(cmd *cobra.Command, args []string) {
...
+				fmt.Printf("-p%d ", showPage+1)
+				found = true
+			} else if len(arg) == 6 && arg[:6] == "--page" {
+				fmt.Printf("-p%d ", showPage+1)
+				found = true
+				i++
 			} else {
-				break
+				fmt.Print(os.Args[i], " ")
 			}
```
vkhoroz, line 179: This is quite elaborate logic, but I'm not sure it gives any added value to the user - they should be able to increment a page number themselves.

I think it should be enough to just say that "There are more devices matching your query.  Please use `--page=<page>` option to see next page(s)."

subcommands/devices/list.go
```diff
@@ -105,6 +106,7 @@ func init() {
 	listCmd.Flags().IntVarP(&deviceInactiveHours, "offline-threshold", "", 4, "List the device as 'OFFLINE' if not seen in the last X hours")
 	listCmd.Flags().StringVarP(&deviceUuid, "uuid", "", "", "Find device with the given UUID")
 	listCmd.Flags().StringSliceVarP(&showColumns, "columns", "", defCols, "Specify which columns to display")
+	listCmd.Flags().IntVarP(&showPage, "page", "p", 1, "Page of devices to display when pagination is needed")
```
vkhoroz, line 109: Maybe, let's allow a user to specify a limit too (e.g. with tail I'm rarely interested in more than top 10-20 results, many users would think the same about this command) ?
It's not about performance, it's about convenience.

vkhoroz (approved):

subcommands/devices/list.go
doanac: good idea - https://github.com/foundriesio/fioctl/pull/93/commits/41c5dd8f190b21cb60fd2ac7866d96a0c34d1cfa

subcommands/devices/list.go
```diff
@@ -125,8 +137,20 @@ func sqlLikeIfy(filePathLike string) string {
...
 
+func assertPagination() {
+	// hack until: https://github.com/spf13/pflag/issues/236
+	for _, x := range paginationLimits {
+		if x == paginationLimit {
+			return
+		}
+	}
+	fmt.Println("ERROR: Invalid limit:", paginationLimit)
+	os.Exit(1)
```
vkhoroz, line 148: subsommands.DieNotNil(fmt.Errorf("Invalid limit: %d", paginationLimit))

vkhoroz:
LGTM

doanac:
rebased on master

subcommands/devices/list.go
doanac: fixed with https://github.com/foundriesio/fioctl/pull/93/commits/7bf5cc9970dfd96606829f80a02893da151e5cff

doanac:
sqush push fix

PR #94, opened 2021-03-24
Add ability to sign wave targets offline 

vkhoroz:


subcommands/waves/init.go
```diff
@@ -50,6 +52,12 @@ func doInitWave(cmd *cobra.Command, args []string) {
 	subcommands.DieNotNil(err, "Version must be an integer")
 	expires := readExpiration(cmd)
 	dryRun, _ := cmd.Flags().GetBool("dry-run")
+	offlineKeysFile, _ := cmd.Flags().GetString("keys")
+	var offlineKeys keys.OfflineCreds
+	if offlineKeysFile != "" {
```
doanac, line 57: how could this be an empty string? If possible, it seems like an error to die on and not a conditional to check?

subcommands/waves/init.go
```diff
@@ -94,13 +102,19 @@ func doInitWave(cmd *cobra.Command, args []string) {
 
 	meta, err := json.MarshalCanonical(targets)
 	subcommands.DieNotNil(err, "Failed to serialize new targets")
+	var signatures []tuf.Signature
+	if offlineKeys != nil {
```
doanac, line 106: Are you trying to make the offline signing optional? I'm not sure why we'd allow this - the backend is presumably going to start rejecting production targets not signed with this key?

vkhoroz: A logic is simple (binary):
- if offline targets were configured for this factory - a backend will reject non-signed wave.
- if offline targets were not configured - a backend will reject signed wave.
I'm not sure all customers would want to take targets keys offline, and I see no business reason why deny using waves in this case.

vkhoroz: Think of this, I want to allow new customers to try using waves without going fully offline (and some wouldn't have proper infrastructure to go offline anyway).
Yes, we will encourage going offline by all means, and at one day a customer will prepare infra for their own PKI and go offline, at which point all existing prod targets will be re-signed, and for new prod/wave targets offline signing becomes mandatory.

doanac: they already have to have their own PKI set up for this to create the proper certs. If we make this a possibility someone will eventually go to production in the wrong way. We will indeed hit friction enabling this for customers. We need to work out  helping with the friction. It needs to be *impossible* for waves to go out to devices without a customer's cryptographic consent.

vkhoroz: I think this way, in general I buy your arguments but they seem too strong to me.
For some small customers we are at risk of pushing them into using an unprepared infrastructure for PKI only because it's mandatory.
But I really want to have this finished in ANY form, so I will make this mandatory for now.

doanac: a primary requirement for this whole body of work was to make it technical impossible for our backend to push an update to a production device.

subcommands/waves/init.go
```diff
@@ -135,6 +168,17 @@ func readExpiration(cmd *cobra.Command) (expires time.Time) {
 	return expires.Round(time.Second)
 }
 
+func readOfflineKeys(cmd *cobra.Command) keys.OfflineCreds {
+	offlineKeysFile, _ := cmd.Flags().GetString("keys")
+	if offlineKeysFile == "" {
```
doanac, line 173: things like this are why i hate cobra. i'm sure there's a better way but the API leads you in bad directions. no worries, just ranting.

doanac (approved):

subcommands/waves/init.go
vkhoroz: your are the biggest foe of my laziness: 31255be

PR #99, opened 2021-04-28
Add new static-delta command

doanac:

### DIFF
--- a/subcommands/targets/deltas.go
+++ b/subcommands/targets/deltas.go
@@ -1,0 +1,112 @@
+package targets
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/foundriesio/fioctl/client"
+	"github.com/foundriesio/fioctl/subcommands"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var (
+	byTag  string
+	dryRun bool
+)
+
+func init() {
+	var deltas = &cobra.Command{
+		Use:   "static-deltas <target-version> [<from-version>...]",
+		Short: "Generate static deltas to the given target version to make OTAs faster",
+		Run:   doDeltas,
+		Args:  cobra.MinimumNArgs(1),
+		Long: `In many cases OTA updates will have many OSTree changes. These updates
+can be downloaded significantly faster by generating OSTree static
+deltas. Static deltas are generated with a "from(sha) -> to(sha)" type
+logic. This command takes the given Target version and will produce a
+number of static deltas to ensure devices will be updated efficiently.
+
+There are two ways to run this command:
+
+ # Generate static deltas for 30->42 and 31->42
+ fioctl targets static-deltas 42 30 31
+
+ # Find the target versions of all devices configured to the "prod" tag.
+ # Generate a static delta from those versions to version 42.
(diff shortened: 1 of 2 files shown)


doanac:
@mike-sul - you probably want to look at this also

subcommands/targets/deltas.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+func init() {
+	var deltas = &cobra.Command{
+		Use:   "static-deltas <target-version> [<from-version>...]",
+		Short: "Generate static deltas to the given target version to make OTAs faster",
+		Run:   doDeltas,
+		Args:  cobra.MinimumNArgs(1),
+		Long: `In many cases OTAs will have many ostree changes. These OTAs can be 
+sped up significantly by generating OSTree static deltas. Static deltas
```
mike-sul, line 27: `sped` --> `speed`

subcommands/targets/deltas.go
```diff
@@ -0,0 +1,109 @@
...
+	dryRun bool
+)
+
+func init() {
+	var deltas = &cobra.Command{
+		Use:   "static-deltas <target-version> [<from-version>...]",
+		Short: "Generate static deltas to the given target version to make OTAs faster",
+		Run:   doDeltas,
+		Args:  cobra.MinimumNArgs(1),
+		Long: `In many cases OTAs will have many ostree changes. These OTAs can be
```
mike-sul, line 26: I am not sure about it, just think that `OTAs` should be `OTA updates` since OTA stands just for Over-The-Air?

mike-sul (approved):
Just minor comments. I actually tried it by following the guide https://github.com/foundriesio/docs/pull/213/files and it worked for me, device/aklite pulled static deltas.

mike-sul:
Maybe it makes sense to check whether `from-version` is actually lower than `target-version` to forbid doing something like this `fioctl targets static-deltas 803 806` which doesn't make much sense.

subcommands/targets/deltas.go
doanac: i fear we probably say "OTAs" or just OTA in many places now. After a big of googling i'll say Android is probably the best to follow:

 https://source.android.com/devices/tech/ota

they say "OTA updates". I'll fix this to not make things worse, but there's probably a lot of stuff docs/blogs/code that repeats this.

subcommands/targets/deltas.go
doanac: no this is "sped up" as in made faster. maybe i should say "These updates can be downloaded faster by generating .."

doanac: redid the wording - https://github.com/foundriesio/fioctl/pull/99/commits/abb98ca62cfc1ce83520149e69977c0ff7656ee8

subcommands/targets/deltas.go
doanac: https://github.com/foundriesio/fioctl/pull/99/commits/abb98ca62cfc1ce83520149e69977c0ff7656ee8

doanac:
> Maybe it makes sense to check whether from-version is actually lower than target-version

Good idea: https://github.com/foundriesio/fioctl/pull/99/commits/4bf5fb03579e7f649216e23dcb13b13acc01b57e

doanac:
squash push

doanac:
rebase on master

Issue #100, opened 2021-04-29
config updates mentions <device> param erroneously


mike-scott:
The help texts for `fioctl config updates` mentions `<device>`, but this parameter is silently ignored.
https://github.com/foundriesio/fioctl/blob/master/subcommands/config/updates.go#L14

Further, the short description could be clarified to:
`Configure aktualizr-lite settings for how updates are applied factory wide to devices`

and the long description:
`View or change factory wide configuration parameters used by aktualizr-lite for updating devices.`
where all of the following examples would have the `<device>` param removed.

mike-scott:
Fixed with: https://github.com/foundriesio/fioctl/pull/101

PR #101, opened 2021-04-29
subcommands: config updates: fix help texts

mike-scott:
- Drop the erroneous <device> parameter which is silently ignored
- Clarify the short text by making device plural
- Clarify the long text by making device plural and dropping the
  <device> param in the examples.
- Clarify status message after change with "factory wide" or
  "group wide"

Signed-off-by: Michael Scott <mike@foundries.io>

### DIFF
--- a/subcommands/config/updates.go
+++ b/subcommands/config/updates.go
@@ -11,28 +11,28 @@ import (
 
 func init() {
 	configUpdatesCmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


mike-scott:
Fixes: https://github.com/foundriesio/fioctl/issues/100

doanac (approved):
that was definite copy/paste on my part. thanks

PR #105, opened 2021-05-24
Use new wireguard-ips API for finding IP

doanac:
The original method we used can't work on large numbers of devices. This
change uses a new API from the server to more efficiently find a free IP
address.


doanac:
rebased on master

PR #108, opened 2021-06-01
Change references of device tags to tag

doanac:
A device can only follow a single tag. This fixes all the code referencing that.


vkhoroz:
IMHO dropping compose apps logic should go as a separate PR.

Anyway, good job!

vkhoroz (approved):

doanac:
>  IMHO dropping compose apps logic should go as a separate PR.

Yep. I got greedy/lazy near the end.

PR #114, opened 2021-06-24
Fix bug in targets prune

doanac:
Target names used to look like intel-corei7-64-lmp-38  but now they
can look like intel-corei7-64-lmp-38-master and
intel-corei7-64-lmp-38-production if a factory does advanced tagging.

This uses a more fool-proof approach for finding the version


Tim-Anderson (approved):
I will believe that api.TargetCustom() will properly parse the name LGTM

doanac:
> I will believe that api.TargetCustom() will properly parse the name LGTM

It actually gets the TUF Custom Metadata which includes the "version" field. So this is more precise. The old approach was a little lazy on my part.

PR #124, opened 2021-11-15
Require a factory argument for device list

vkhoroz:
This also fixes a previously malfunctioning just-mine argument in case if the user specified the factory


doanac (commented):
this works as expected. I think we can make one other small improvement. Now that we list for a single factory, the factory column isn't very useful. I think we can remove it "factory" from:

 https://github.com/foundriesio/fioctl/blob/d837bc4bc263502796988885196bf515ceb5b86a/subcommands/devices/list.go#L91

and then things are even better

PR #126, opened 2021-11-19
targets: Show origin target where applicable

doanac:
The webui includes an "Origin Target" which can be quite handy when trying to debug factory targets. This adds the same type functionality. It also lets users customize what columns to show when listing targets.

### DIFF
--- a/subcommands/targets/list.go
+++ b/subcommands/targets/list.go
@@ -19,9 +19,10 @@ import (
 )
 
 var (
-	listProd  bool
-	listRaw   bool
-	listByTag string
+	listProd    bool
+	listRaw     bool
+	listByTag   string
+	showColumns []string
 )
 
 // Represents the details we use for displaying a single OTA "build"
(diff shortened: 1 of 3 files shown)


doanac:
@Tim-Anderson - you may want to give this a spin and see if it helps

@vkhoroz - FYI.

subcommands/targets/list.go
```diff
@@ -52,16 +53,36 @@ func (t byTargetKey) Less(i, j int) bool {
...
+var Columns = map[string]column{
+	"version":      {func(tl *targetListing) string { return strconv.Itoa(tl.version) }},
+	"tags":         {func(tl *targetListing) string { return strings.Join(tl.tags, ",") }},
+	"apps":         {func(tl *targetListing) string { return strings.Join(tl.apps, ",") }},
+	"hardware-ids": {func(tl *targetListing) string { return strings.Join(tl.hardwareIds, ",") }},
+	"origin":       {func(tl *targetListing) string { return tl.origin }},
+}
+
 func init() {
+	var defCols = []string{"version", "tags", "apps", "origin"}
```
vkhoroz, line 69: We just have 5 columns atm, I think it's fine to show them all by default.
This column selector is neat in any case.

subcommands/targets/list.go
```diff
@@ -30,6 +30,7 @@ type targetListing struct {
 	hardwareIds []string
 	tags        []string
 	apps        []string
+	origin      string
 }
```
vkhoroz, line 35: What do you think about other TufCustom fields (target format, 3 hashes)?
Does it make sense to add these columns as optional values ?

vkhoroz (approved):
LGTM

subcommands/targets/list.go
doanac: it started scrolling off my screen, so i got mad. hardware-id isn't that useful for most customers. they build for a single platform. However, a few factories and things like the public LMP do. So you sometimes need to see, but not usually.

subcommands/targets/list.go
doanac: probably not. target-format != ostree don't get skipped. The "sha" fields are really long git urls and wouldn't fit well in column style list

doanac: but - why not have the "sha" fields optional. saves a support ticket some day. i'll add it.

doanac: https://github.com/foundriesio/fioctl/pull/126/commits/b4a98a43a71ae62f5fc27f477c0e261ec4ec81f5

vkhoroz (approved):

mike-scott (approved):
Tested this just now.  Looks great.

PR #128, opened 2021-11-19
Improvements to target tagging

doanac:
This changeset adds some usability improvements to how targets are tagged.

### DIFF
--- a/subcommands/targets/tag.go
+++ b/subcommands/targets/tag.go
@@ -34,6 +34,7 @@ func init() {
 	tagCmd.Flags().StringVarP(&tagTags, "tags", "T", "", "comma,separate,list")
 	tagCmd.Flags().BoolVarP(&tagNoTail, "no-tail", "", false, "Don't tail output of CI Job")
 	tagCmd.Flags().BoolVarP(&tagByVersion, "by-version", "", false, "Apply tags to all targets matching the given version(s)")
+	tagCmd.Flags().BoolVarP(&dryRun, "dryrun", "", false, "Just show the changes that would be applied")
 }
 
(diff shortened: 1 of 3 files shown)


doanac:
@mike-scott @Tim-Anderson - I added a `--dryrun` option to this command as well. I think this option is actually more powerful/useful than the `--append` I added.

subcommands/targets/tag.go
```diff
@@ -32,33 +35,55 @@ func init() {
 	}
 	cmd.AddCommand(tagCmd)
 	tagCmd.Flags().StringVarP(&tagTags, "tags", "T", "", "comma,separate,list")
+	tagCmd.Flags().BoolVarP(&tagAppend, "append", "", false, "Append the given tags rather than set them")
```
Tim-Anderson, line 38: should this be called appendtag or append_tag to make it clearer?

Tim-Anderson (commented):
still reviewing

subcommands/targets/tag.go
doanac: I didn't have a strong opinion, but it felt redundant. Its the *tag* command. Feels a bit like saying "ATM machine", but some people prefer that.

Tim-Anderson: we can let Mike chime in on that

mike-scott: @Tim-Anderson the full command would look like this:
```
fioctl targets tag --append <tag> --by-version 42
```

When I read that back to myself it makes perfect sense.

mike-scott (approved):
Approach looks good to me.  Haven't had time to play test it.

PR #130, opened 2021-11-29
Allow secrets to come from a file

doanac:
This is the same approach that we added for fioctl-config-set


mike-scott (approved):
LGTM

PR #132, opened 2021-12-01
Pass the required factory argument to an API for all device commands

vkhoroz:


vkhoroz:
@doanac I will test this tomorrow; it should work well with the new API

vkhoroz:
@doanac tested with the recent API changes, it works.

doanac (approved):
that was a pain. thanks.

PR #133, opened 2021-12-20
Introduce new initial key rotation logic

doanac:
We now have customer's first root key available to download (to factory
admins). This allows us to make the first root rotation a little easier.

### DIFF
--- a/subcommands/keys/rotate_root.go
+++ b/subcommands/keys/rotate_root.go
@@ -19,7 +19,10 @@ import (
 	"github.com/spf13/viper"
 )
 
-var doRootSync bool
+var (
+	doRootSync      bool
+	initialRotation bool
+)
 
 func init() {
 	rotate := &cobra.Command{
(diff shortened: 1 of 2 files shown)


doanac:
this will compliment - https://git.foundries.io/development/cloudplatforms/infrastructure/ota-lite/-/tree/initial-tuf-root once we get that merged. this is sort of a "preview PR".

vkhoroz:
> this will compliment - https://git.foundries.io/development/cloudplatforms/infrastructure/ota-lite/-/tree/initial-tuf-root once we get that merged. this is sort of a "preview PR".

I'm not sure it's OK to reference a private repo from the public repo.
The opposite direction usually works fine.

subcommands/keys/rotate_root.go
```diff
@@ -31,15 +34,44 @@ func init() {
...
-	assertWritable(credsFile)
-	creds, err := GetOfflineCreds(credsFile)
-	subcommands.DieNotNil(err)
+
+	var creds OfflineCreds
+
+	if initialRotation {
+		// This lacks in efficiency (we wind up reading root.json from
+		// from server twice and create the creds.tgz file twice), but it makes
+		// up for that in clarity and ease of testing.
```
vkhoroz, line 50: We can have as readable code without fetching root.json twice if we move lines 76-77 inside the else clause (and declare the root variable above).

doanac: this is why code comments are overrated :) thanks for the tip. force push fixed

vkhoroz (approved):
LGTM

PR #134, opened 2021-12-24
Show devices as offline when last-seen is not set

vkhoroz:
I have a bunch of synthetic devices which have never contacted our APIs.
And still all of them show as OK in the device list despite their last-seen is not set.

### DIFF
--- a/subcommands/devices/list.go
+++ b/subcommands/devices/list.go
@@ -37,7 +37,7 @@ func statusFormatter(d *client.Device) string {
 	if len(d.Status) > 0 {
 		status = d.Status
 	}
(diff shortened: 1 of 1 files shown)


vkhoroz:
@doanac a side-effect of this change is that newly added devices will initially start as "OFFLINE" before transitioning into the OK status after they first contact the device gateway.

I think that's good for now, as that's how we treat these devices throughout our APIs.

An alternative would be to treat them as OK if `Now - CreatedAt < Threshold`, but in this case we need to change `factory/group/wave status` APIs.  I don't think that is worth an effort, and I'm not even sure that's a reasonable alternative.

doanac (approved):

PR #137, opened 2022-01-10
Console tailing improvements

doanac:
* make more visible highlights to status changes
* fix tailing crash when client hits HTTP error
* add ability to tail ci jobs


vkhoroz (approved):
LGTM

PR #138, opened 2022-01-13
Config updates

doanac:
small fixes to configuration handling.

### DIFF
--- a/subcommands/common_config.go
+++ b/subcommands/common_config.go
@@ -198,6 +198,8 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
 		fmt.Printf("Changing apps from: [%s] -> [%s]\n", configuredApps, opts.UpdateApps)
 		sota.Set("pacman.compose_apps", opts.UpdateApps)
 		changed = true
+	} else if len(configuredApps) > 0 {
+		sota.Set("pacman.compose_apps", configuredApps)
 	}
 	if opts.UpdateTag != "" && configuredTag != opts.UpdateTag {
 		if strings.TrimSpace(opts.UpdateTag) == "," {
@@ -206,6 +208,8 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
 		fmt.Printf("Changing tag from: %s -> %s\n", configuredTag, opts.UpdateTag)
 		sota.Set("pacman.tags", opts.UpdateTag)
 		changed = true
+	} else if len(configuredTag) > 0 {
+		sota.Set("pacman.tags", configuredTag)
 	}
 
 	if !changed {
--- a/subcommands/common_config.go
+++ b/subcommands/common_config.go
@@ -196,7 +196,6 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
 			opts.UpdateApps = ""
 		}
 		fmt.Printf("Changing apps from: [%s] -> [%s]\n", configuredApps, opts.UpdateApps)
-		sota.Set("pacman.docker_apps", opts.UpdateApps)
 		sota.Set("pacman.compose_apps", opts.UpdateApps)
 		changed = true
 	}


doanac:
@vkhoroz - this is really about the "common_config" fix, but I noticed another small cleanup while i was digging into this.

subcommands/common_config.go
```diff
@@ -196,18 +196,19 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
...
 		if strings.TrimSpace(opts.UpdateTag) == "," {
 			opts.UpdateTag = ""
 		}
 		fmt.Printf("Changing tag from: %s -> %s\n", configuredTag, opts.UpdateTag)
-		sota.Set("pacman.tags", opts.UpdateTag)
 		changed = true
 	}
 
+	sota.Set("pacman.compose_apps", opts.UpdateApps)
+	sota.Set("pacman.tags", opts.UpdateTag)
```
vkhoroz, line 210: I'm not sure this is the correct fix.

`fioctl config updates --tags=devel` should only set tags and not apps.

So I think we should instead change the `if` logic e.g. smth like this:
```
if opts.UpdateTag != "" {
    // trim space logic
    sota.Set("pacman.tags", opts.UpdateTag)
    if configuredTag != opts.UpdateTag {
        // print "changing" message
    }
}
```

vkhoroz: In short - we should always set **those** fields in the toml **a user asked to set**.

doanac: correct. that's what this fix does. if the user doesn't provide a field we set that field to what the device is reporting. without this fix we were *unsetting* that field.

vkhoroz: hmm... I'm not sure I understand the logic corrently then.

Lets imagine these 2 use cases:

1. ```
/var/sota/sota.toml:
apps=httpd
tags=devel

fioctl config updates --apps=httpd --tags=master
```
should produce
```
/var/sota/conf.d/z-50-fioctl.toml:
apps=httpd
tags=master
```

2. ```
/var/sota/sota.toml:
apps=httpd

fioctl config updates --tags=master
```
should produce
```
/var/sota/conf.d/z-50-fioctl.toml:
tags=master
```

This is my understanding... is it correct ?

This new code for the second use case will produce this:
```
/var/sota/conf.d/z-50-fioctl.toml:
apps=
tags=master
```

Am I reading this new code correctly ?
If yes, I assume that it will override the `apps=httpd` from `/var/sota/sota.toml`, but the user didn't ask for that.
Is that assumption correct ?

doanac: you may want to re-read my commit message. it covers the problem, but i'll restate here for context in the thread. you have a device with:
```
# /var/sota/sota.toml
apps=foo
tags=devel
```
Factory wide config is:
```
apps=blah
tags=master
```

if you run `fioctl devices config updates --apps=bar`, our code is currently just producing 
```
apps=bar
```
which means the device will be subscribed to the `devel` tag (from /var/sota/sota.toml) rather than the `master` tag that it should stay on.

doanac: also as per your 
```
> This new code for the second use case will produce this:

/var/sota/conf.d/z-50-fioctl.toml:
apps=httpd`
tags=master
```
That's not correct it would have `apps=

vkhoroz: From another perspective, what is an eexpected result if somebody runs these two commands in a row:

```
fioctl config updates --apps=httpd
fioctl config updates --tags=master
```

I think the resulting `z-50-fioctl.toml` should be this:
```
apps=httpd
tags=master
```
because in the first round the user requested to change apps, while in the second command - tags

With this new code we will produce this file:
```
apps=
tags=master
```
i.e. we effectively cleared the apps setting. Is that intended ?

Note that with the previous implementation user still can achieve that result:
```
fioctl config updates --apps=httpd
fioctl config updates --tags=master --apps=,
```

vkhoroz: I actually don't see how the previous implementation has a bug as per my above comment.

vkhoroz: Anyway, my initial suggestion would still work if I understand the problem correctly.

vkhoroz:
@doanac this second patch LGTM in achieving the goal to "prefill" the device config.
Just as we discussed in Slack, maybe it's worth to investigate the "inheritance" idea for the updates config which is more complete but requires a more complex yet feasible change to the device gateway.

doanac:
we've come up with a way that this can be handled from our backend.

PR #140, opened 2022-01-18
shell: Added fish completions using Cobra

StealthyCoder:
Just using the standard function exposed by Cobra to generate Fish
completions. Had to update the dependency of cobra to latest version
though.


doanac:
should be a simple change, but this is a huge update to cobra. we should poke around and make sure nothing has broken

doanac:
beyond some smoke testing of the commands, we should look compare before/after changes to `fioctl gen-rst` to make sure its still going to produce valid content for docs.foundries.io.

StealthyCoder:
I just generated the docs with `fioctl gen-rst` in a directory `new_docs` with the build from this branch. I generated docs with `fioctl gen-rst` into `old-docs` with the release v0.23. 

I then did a diff between all files and the only difference is the following:

```
fioctl_completion.rst
46,48d45
<   # Fish:
<   $ fioctl completion fish > ~/.config/fish/completions/fioctl.fish
< 
``` 

Which makes sense as that is the one thing I added in this commit. Is that enough of a check to make it all good?

doanac:
can you added [signed-off-by](https://gerrit-review.googlesource.com/Documentation/user-signedoffby.html) i.e. `git commit --sigoff ...`

StealthyCoder:
Added the Signed-off-by and Reviewed-by :+1:

doanac (approved):

PR #144, opened 2022-01-27
keys: Check if file exists during initial rotation

StealthyCoder:
During initial key rotation make sure the file does not exist already as
it could happen that when you do the initial rotation for two different
factories you might accidentially overwrite a file.

### DIFF
--- a/subcommands/keys/rotate_root.go
+++ b/subcommands/keys/rotate_root.go
@@ -8,6 +8,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"os"
 	"time"
@@ -48,6 +49,10 @@ func doRotateRoot(cmd *cobra.Command, args []string) {
 	subcommands.DieNotNil(err)
 
 	if initialRotation {
+		if _, err := os.Stat(credsFile); err == nil {
+			subcommands.DieNotNil(errors.New("Destination file exists. Please make sure you aren't accidentally overwriting another factory's keys"))
+		}
+
 		key, err := api.TufRootFirstKey(factory)
 		subcommands.DieNotNil(err)
 


StealthyCoder:
I could use some suggestions for the error messages though. What do you think? @doanac @vkhoroz

subcommands/keys/rotate_root.go
```diff
@@ -48,6 +48,13 @@ func doRotateRoot(cmd *cobra.Command, args []string) {
 	subcommands.DieNotNil(err)
 
 	if initialRotation {
+		if _, err := os.Stat(credsFile); err == nil {
+			fmt.Printf("ERROR: %v already exists\n", credsFile)
+			fmt.Printf("WARNING: do not just remove %v, could contain other keys\n", credsFile)
+			fmt.Println("Make sure to run fioctl keys rotate-root --initial /absolute/path/to/non.existent.file.tgz")
+			os.Exit(1)
```
doanac, line 55: i don't want to turn this into bikeshedding but we could probably keep it more concise and consistent with:
```
subcommands.DieNotNil(errors.New("Destination file exists. Please make sure you aren't accidentally overwriting another factory's keys"))
```

StealthyCoder: Done.

StealthyCoder:
Just as an added extra information we don't check if the file is a directory because `os.Stat` just returns a success on both. The only thing this might fail now is if you do not have permissions, but it will fail when writing the keys as well.

subcommands/keys/rotate_root.go
```diff
@@ -48,6 +49,10 @@ func doRotateRoot(cmd *cobra.Command, args []string) {
 	subcommands.DieNotNil(err)
 
 	if initialRotation {
+		if _, err := os.Stat(credsFile); err == nil {
+			subcommands.DieNotNil(errors.New("Destination file exists. Please make sure you aren't accidentally overwriting another factory's keys"))
+		}
```
vkhoroz, line 54: There is one more use case when things can go wrong.
At one point I've broken my factory by making the initial rotation into the inaccessible folder.

So, I think it makes sense to do some magic with `touch credsFile` after this to make sure the folder exists and is writable.
Currently, we rotate first, and try to write next... and if that write fails we're doomed.

I know we cannot prevent all use cases when it can break, but at least protecting against silly mistakes is worth it.

vkhoroz: Sorry @StealthyCoder please, disregard this.  I always forget that we write first/post next

doanac: this branch doesn't rotate anything - its just downloading and saving *before* we get to the fun stuff. We also try and make sure you don't lose things if the rotate works in the cloud but fails to persist locally:

https://github.com/foundriesio/fioctl/blob/main/subcommands/keys/rotate_root.go#L105-L124

vkhoroz (approved):

doanac (approved):

subcommands/keys/rotate_root.go
StealthyCoder: Added a check to make sure directory exists, and is writable. :+1:

StealthyCoder: @vkhoroz you are still right, because if os.Create fails because the path does not exist or is not writable we get into potential weird states still.

subcommands/keys/rotate_root.go
```diff
@@ -48,6 +50,27 @@ func doRotateRoot(cmd *cobra.Command, args []string) {
...
+				return
+			} else if os.IsPermission(err) {
+				subcommands.DieNotNil(fmt.Errorf("Destination path is not writable: %v", path))
+				return
+			}
+
+			checkPath(filepath.Dir(path))
+		}
+
+		checkPath(filepath.Dir(credsFile))
```
doanac, line 72: this is cluttering up the main function and should be broken out, but I don't think it really adds much value. If we skip the check you already get: "ERROR: /tmp/blah/foo: no such file or directory" that seems sufficient, no?

StealthyCoder: Reverted :+1:

doanac (approved):

subcommands/keys/rotate_root.go
StealthyCoder: I just ran the command on a path that contains non-existent directories, but it will fail after already getting the files for key rotation. Is that an issue?

doanac: nope. it can re-run and get them again. https://git.foundries.io/development/cloudplatforms/infrastructure/ota-lite/-/blob/master/ota_lite/api/factory.py#L446-463

doanac: as a general note about stuff kind of like this. you'll see code all the time like:
```
if file.exists():
   content = open(file).read()
```
it isn't foolproof. Its mostly fine on single-threaded single-user systems, but its easier to just rely on the OS with something like:
```
try:
   content = open(file).read()
except FileNotFoundError
```

StealthyCoder: Gotcha. Then I will merge it as is. :smile:

vkhoroz: Sorry for confusion from my side
I really only cared about loosing keys, not the error message.
A fix as is seems to close all the remaining gaps, so it's fine... I gave my LGTM yesterday.

PR #149, opened 2022-03-21
common: Fix check for factory

doanac:
When running a command that requires a factory such as `fioctl targets
list`, you'd see:

 ERROR: Required flag "factory" not set

This change checks for the login first (which might set the factory) to
give the user a more accurate reponse.


MrCry0 (approved):
LGTM

vkhoroz (approved):
LGTM

PR #150, opened 2022-03-21
Add support for managing a factory's denied devices

doanac:


subcommands/devices/list_denied.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/cheynewallace/tabby"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/foundriesio/fioctl/subcommands"
+)
+
+func init() {
+	paginationLimits = []int{10, 20, 30, 40, 50, 100, 200, 500, 1000}
```
vkhoroz, line 17: I think it's worth extracting this as a constant... it's the same for both device list and denied device list.
Maybe, even extract into common (if can be done easily, not sure about this).

subcommands/devices/list_denied.go
```diff
@@ -0,0 +1,76 @@
...
+		device.Factory = factory // to make ownerFormatter work
+		t.AddLine(device.Uuid, device.Name, ownerFormatter(&device))
+	}
+	t.Print()
+	if dl.Next != nil {
+		fmt.Print("\nNext page of devices can be viewed with: ")
+		found := false
+		for i := 0; i < len(os.Args); i++ {
+			arg := os.Args[i]
+			if len(arg) > 2 && arg[:2] == "-p" {
```
vkhoroz, line 60: IIRC `-p <page>` is as valid as `-p<page>`.

So, we need to also check for `if len(arg) == 2 && arg == "-p"` and use the same logic as below for a `--page`.

subcommands/devices/list_denied.go
```diff
@@ -0,0 +1,76 @@
...
+				fmt.Printf("-p%d ", showPage+1)
+				found = true
+			} else if len(arg) == 6 && arg[:6] == "--page" {
```
vkhoroz, line 63: This can come in two forms: `--page <page>` and `--page=<page>`.
Need to handle both.

subcommands/devices/list_denied.go
```diff
@@ -0,0 +1,76 @@
...
+				i++
+			} else {
+				fmt.Print(os.Args[i], " ")
+			}
+		}
+		if !found {
+			fmt.Print("-p", showPage+1)
+		}
+		fmt.Println()
+	}
```
vkhoroz, line 75: I don't see an easier way to write this logic.
But I have a question if we need it at all.
A contemporary user should be able to understand and add the `-p` argument.

subcommands/devices/delete_denied.go
```diff
@@ -0,0 +1,37 @@
...
+	logrus.Debug("Deleting %r", args)
+
+	for _, uuid := range args {
+		fmt.Printf("Deleting %s .. ", uuid)
+		if err := api.DeviceDeleteDenied(factory, uuid); err != nil {
+			fmt.Printf("failed\n%s", err)
+			os.Exit(1)
+		} else {
+			fmt.Printf("ok\n")
+		}
```
vkhoroz, line 35: Why not `DieNotNil`?

Also, it is the first command which prints 'ok' in case of success. All other commands like this print nothing.

subcommands/devices/list_denied.go
doanac: I just copied what you did in device-list for this. I guess I should first try and make this common code first.

subcommands/devices/delete_denied.go
doanac: i copied the code as-is from delete.go. i think it was done like this because it can remove multiple, so just dying isn't going to show you which device it failed on.

subcommands/devices/list_denied.go
vkhoroz: it wasn't my logic though :-)

subcommands/devices/delete_denied.go
vkhoroz: Maybe, it is an instance of code I missed when refactoring to use DieNotNil everywhere; maybe it was added later; I don't know.
What I know is that `commands.DieNotNil(api.DeviceDelete(factory, name), "failed"); fmt.Println("ok")` is equivalent; but I'd prefer `commands.DieNotNil(api.DeviceDelete(factory, name))` as I see no value in these `failed, ok` things.

Just dying will show clearly on which device it failed due to these `Deleting xyz...` messages.

doanac:
@vkhoroz - i've rebased this on a change that first uses common device list display logic. Then I've fixed the pagination display logic with [3aa5e3f](https://github.com/foundriesio/fioctl/pull/150/commits/3aa5e3f5e484d24135ff7de51b31c7bbfc2a4fce). I know you aren't a big fan of that, but I think we should try and help there cause its not very obvious to me what to do without it.

subcommands/devices/list.go
```diff
@@ -183,10 +183,17 @@ func showDeviceList(dl *client.DeviceList, showColumns []string) {
 			if len(arg) > 2 && arg[:2] == "-p" {
 				fmt.Printf("-p%d ", showPage+1)
 				found = true
-			} else if len(arg) == 6 && arg[:6] == "--page" {
+			} else if arg == "-p" {
```
MrCry0, line 186: The following 2 blocks may be rewritten as
```
} else if arg == "-p" || arg == "--page" {
	fmt.Printf("-p%d ", showPage+1)
	found = true
	i++
}
```

subcommands/devices/list.go
```diff
@@ -183,10 +183,13 @@ func showDeviceList(dl *client.DeviceList, showColumns []string) {
...
 				fmt.Printf("-p%d ", showPage+1)
 				found = true
-			} else if len(arg) == 6 && arg[:6] == "--page" {
+			} else if arg == "-p" || arg == "--page" {
 				fmt.Printf("-p%d ", showPage+1)
 				found = true
 				i++
+			} else if len(arg) > 7 && arg[:7] == "--page=" {
+				fmt.Printf("-p%d ", showPage+1)
+				found = true
```
vkhoroz, line 192: I'm sorry for nitpicking, but this can also be combined with the above as:
`if (len(arg) > 2 && arg[:2] == "-p") || (len(arg) > 7 && arg[:7] == "--page=") {...}`

vkhoroz (approved):
Nice, LGTM.

subcommands/devices/list.go
doanac: not a nit. it cleans things up. force push fixed

vkhoroz:
ship it!

MrCry0 (approved):
LGTM

PR #151, opened 2022-03-24
Add configure-docker helper

doanac:
This allows you to do:

 sudo fioctl configure-docker

It will create a symlink from fioctl->/usr/bin/docker-credential-fio
and update your docker config with the helper. After that, docker
commands "just work".

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -13,6 +13,7 @@ import (
 	"github.com/foundriesio/fioctl/subcommands"
 	cfgcmd "github.com/foundriesio/fioctl/subcommands/config"
 	"github.com/foundriesio/fioctl/subcommands/devices"
+	"github.com/foundriesio/fioctl/subcommands/docker"
 	"github.com/foundriesio/fioctl/subcommands/keys"
 	"github.com/foundriesio/fioctl/subcommands/login"
 	"github.com/foundriesio/fioctl/subcommands/logout"
@@ -37,6 +38,15 @@ var rootCmd = &cobra.Command{
 }
 
 func Execute() {
+	if os.Args[0] == docker.DOCKER_CREDS_HELPER {
+		if len(os.Args) != 2 || os.Args[1] != "get" {
+			fmt.Printf("Usage: %s get\n", os.Args[0])
+			os.Exit(1)
+		}
+		initConfig()
+		os.Exit(docker.RunCredsHelper())
+	}
+
 	if err := rootCmd.Execute(); err != nil {
 		fmt.Println(err)
 		os.Exit(1)
@@ -53,6 +63,7 @@ func init() {
 
 	rootCmd.AddCommand(cfgcmd.NewCommand())
 	rootCmd.AddCommand(devices.NewCommand())
+	rootCmd.AddCommand(docker.NewCommand())
 	rootCmd.AddCommand(keys.NewCommand())
 	rootCmd.AddCommand(login.NewCommand())
 	rootCmd.AddCommand(logout.NewCommand())
--- a/subcommands/docker/cmd.go
+++ b/subcommands/docker/cmd.go
@@ -1,0 +1,129 @@
+package docker
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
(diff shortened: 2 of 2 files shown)


cmd/root.go
```diff
@@ -44,6 +45,11 @@ func Execute() {
 }
 
 func init() {
+	if os.Args[0] == docker.DOCKER_CREDS_HELPER {
+		initConfig()
+		os.Exit(docker.RunCredsHelper())
+	}
```
vkhoroz, line 51: It seems dangerous to me to exit from inside the `init` function (with success). In general, this behavior is undefined as there might be other `init` functions which will or will not be called in random order.
I'd recommend to move this logic into `func Execute`.

subcommands/docker/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+func NewCommand() *cobra.Command {
+	path, err := exec.LookPath("docker")
+	subcommands.DieNotNil(err, "Docker not found on system")
+	helperPath = filepath.Dir(path)
+
+	dockerConfigFile = dockerConfigPath()
+
+	cmd := &cobra.Command{
+		Use:   "configure-docker",
+		Short: "Configure a hub.foundries.io Docker credential helper",
```
vkhoroz, line 34: Some bigger help message about how to use the `creds-path` or run this in sudo mode would be helpful.

subcommands/docker/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+	dst := filepath.Join(helperPath, DOCKER_CREDS_HELPER)
+	fmt.Println("Symlinking", self, "to", dst)
+	subcommands.DieNotNil(os.Symlink(self, dst))
+
+	fmt.Println("Adding hub.foundries.io helper to", dockerConfigFile)
+	subcommands.DieNotNil(ioutil.WriteFile(dockerConfigFile, configBytes, 0o600))
+}
+
+func RunCredsHelper() int {
+	subcommands.Login(NewCommand())
```
vkhoroz, line 104: Please, add a comment how this call helps. IIUC it is to make sure that the access token is set properly, right?

But here is the problem. A user might also set the Token config parameter (I usually do that) in which case the AccessToken is empty. Need to workaround this somehow e.g. unset the Token attribute in Config.

subcommands/docker/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+	creds := struct {
+		Username string
+		Secret   string
+	}{
+		Username: "_fioctl_dch",
```
vkhoroz, line 109: This seem to work well; also please, consider this flow: https://raw.githubusercontent.com/google/go-containerregistry/main/images/credhelper-oauth.svg

vkhoroz:
Looks good, just a few areas where I've spotted a place for improvement.

subcommands/docker/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+		Secret:   subcommands.Config.ClientCredentials.AccessToken,
+	}
+
+	bytes, err := json.Marshal(creds)
+	subcommands.DieNotNil(err)
+	os.Stdout.Write(bytes)
+	return 0
```
vkhoroz, line 116: The https://docs.docker.com/engine/reference/commandline/login/#credential-helpers documentation is very sparse, but what I understood from it as well as the google diagrams (see my other comments) - docker login will send a `docker-credential-fio get` command when it needs to login. Maybe, it makes sense to check for the `get` subcommand here explicitly; and bail out if docker sends some other (unexpected) subcommand.

cmd/root.go
doanac: fixed in force-push

subcommands/docker/cmd.go
doanac: added more in the force-push - see what you think

subcommands/docker/cmd.go
doanac: force-push includes a comment. it also has a check - it will exit with an error if oauth2 creds aren't being used

vkhoroz:
LGTM

PR #154, opened 2022-04-19
Add support for event queues

doanac:


doanac:
@mike-sul @vkhoroz @StealthyCoder - this depends on the ota-lite API that's under review. Assuming nothing major changes, there, I think this will be okay.

The big gap is documentation. I set off to do more the `Long` attributes of the subcommands, but it gets too detailed too quick. I think I need to focus on dedicated documentation that we can instead link to.

subcommands/events/create.go
```diff
@@ -0,0 +1,56 @@
...
+func init() {
+	cmd.AddCommand(&cobra.Command{
+		Use:   "mk-push <label> <url>",
+		Short: "Create an event queue that will ingest events at the URL",
+		Args:  cobra.ExactArgs(2),
+		Run:   doCreatePush,
+	})
+
+	cmd.AddCommand(&cobra.Command{
+		Use:   "mk-pull <label> <creds file>",
```
mike-sul, line 22: Do we need to comment/explain to users what `<creds file>` mean?

subcommands/events/create.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/foundriesio/fioctl/client"
+	"github.com/foundriesio/fioctl/subcommands"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func init() {
+	cmd.AddCommand(&cobra.Command{
+		Use:   "mk-push <label> <url>",
```
mike-sul, line 15: Customer may want to provide some auth token along with a URL to push to?

subcommands/events/create.go
```diff
@@ -0,0 +1,56 @@
...
+		Run:   doCreatePull,
+	})
+}
+
+func doCreatePush(cmd *cobra.Command, args []string) {
+	factory := viper.GetString("factory")
+	logrus.Debugf("Create a push queue for: %s", factory)
+
+	queue := client.EventQueue{
+		Label:   args[0],
```
mike-sul, line 34: Do we know that `args` is not empty?

subcommands/events/create.go
doanac: no - this is done in standard webhook fashion where the sender will include a JWT that can be verified by the receiver. This is described in the "Webhook Security" section of the design doc and will be covered in the user facing docs as well.

subcommands/events/create.go
doanac: yes. just above is `cobra.ExactArgs(2)`

subcommands/events/list.go
```diff
@@ -0,0 +1,36 @@
...
+	})
+}
+
+func doList(cmd *cobra.Command, args []string) {
+	factory := viper.GetString("factory")
+	logrus.Debugf("Listing event queues for: %s", factory)
+
+	queues, err := api.EventQueuesList(factory)
+	subcommands.DieNotNil(err)
+	fmt.Println(queues)
```
mike-sul, line 28: Do we need to print `queues` twice? One time here and the second time in the loop below.

doanac: oops - debug slipped in there. i usually do Printl("ANDY", data) so its more obvious :)

mike-sul (approved):
LGTM, just minor comments and clarifications.

subcommands/events/listen.go
```diff
@@ -0,0 +1,55 @@
...
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		for range c {
+			fmt.Println("Exiting...")
+			cancel()
+		}
+	}()
+
+	fmt.Println("Listening for events...")
+	sub := client.Subscription(fmt.Sprintf("f-%s-%s", rid, args[0]))
```
vkhoroz, line 49: I think we leverage the knowledge of the internal details of the server-side implementation here too much, that makes this vulnerable to the future backward-incompatible changes.

I'd prefer if there is an API which returns a `pull-url` for a given event queue i.e. this calculation is done on the server.

subcommands/events/listen.go
```diff
@@ -0,0 +1,55 @@
...
+	err = sub.Receive(ctx, func(ctx context.Context, m *pubsub.Message) {
+		fmt.Println(m.Attributes["event-type"], string(m.Data))
```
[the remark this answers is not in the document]
doanac: The event data includes the timestamp. e.g.
```
DEVICE_OTA_COMPLETED {"Uuid":"fd3c16f3-805a-4b39-8e8a-d8342646694d","Time":1650468508,"Target":"intel-corei7-64-lmp-183-master","Id":"183-9a0cf4f3-1e3b-4d62-80ea-2a56fd20181f","Success":true}
```

client/foundries_events.go
```diff
@@ -0,0 +1,42 @@
+package client
```
vkhoroz, line 1: Is this an effort to split the `foundries.go` into submodules?

subcommands/events/create.go
vkhoroz: This can be confused with the TUF `<creds file>`; I'd be more verbose here e.g. `<oauth creds file to save to>` (or what is it?).

subcommands/events/listen.go
```diff
@@ -0,0 +1,55 @@
...
+		m.Ack()
+	})
```
vkhoroz, line 53: Just wondering, how does the user exit from here? Does a regular `^C` end this loop?

client/foundries_events.go
doanac: not submodules, just files. this is big enough we could start doing a types.go, etc. the one file is just getting painful to edit

subcommands/events/listen.go
doanac: yes. there's a call to signal above this.

subcommands/events/create.go
doanac: commit 2a3de1e7598590f507f05c0d729f6d4f2df7826a

subcommands/events/list.go
doanac: commit 2a3de1e7598590f507f05c0d729f6d4f2df7826a

subcommands/events/listen.go
doanac: > that makes this vulnerable to the future backward-incompatible changes.

I agree with your point - but this will be backwards incompatible - we are creating subscription names and won't be able to modify them.

I've added a change to the backend to include the name in the response and fixed it with commit 7425dbcc2612ef661ce01008cb6f395659b9ee51

subcommands/events/listen.go
vkhoroz: sorry, I was blind

vkhoroz (approved):

vkhoroz:
LGTM

doanac:
docs are about to be merged and device-gateway reviews are coming along. i'll merge this once that's all been completed

PR #157, opened 2022-04-21
Serviceability fixes

doanac:


client/foundries.go
```diff
@@ -531,7 +533,7 @@ func parseJobServResponse(resp *[]byte, err error, runName string) (string, stri
 }
 
 func (a *Api) setReqHeaders(req *http.Request, jsonContent bool) {
-	req.Header.Set("User-Agent", "fioctl-2")
+	req.Header.Set("User-Agent", "fioctl-"+a.clientVer)
```
vkhoroz, line 536: IIRC we've had some code in ota-lite which relies onto the value of this header.
I'm not sure if it is still there. Anyway, can you check if we don't break anything?

PR #159, opened 2022-05-04
Add support for Ed25519 EC signatures for TUF operations

vkhoroz:


vkhoroz:
@doanac I think I will split the last commit, so that you see the "polimorph refactoring" vs "new code".
So, if it looks to dirty, feel free to skip for now.

subcommands/devices/list.go
```diff
@@ -180,7 +180,7 @@ func showDeviceList(dl *client.DeviceList, showColumns []string) {
 		found := false
 		for i := 0; i < len(os.Args); i++ {
 			arg := os.Args[i]
-			if (len(arg) > 2 && arg[:2] == "-p") ||  (len(arg) > 7 && arg[:7] == "--page=") {
+			if (len(arg) > 2 && arg[:2] == "-p") || (len(arg) > 7 && arg[:7] == "--page=") {
```
doanac, line 183: just push this comit to main. no need for a review on it.

subcommands/keys/common.go
```diff
@@ -28,6 +29,59 @@ type TufSigner struct {
...
+	atsPriv      client.AtsKey
+	atsPrivBytes []byte
+
+	atsPub      client.AtsKey
+	atsPubBytes []byte
+
+	keyid string
+}
+
+func GenKeyPair() keypair {
```
doanac: i'm not sure it makes sense to make the function Public but the return type private. doesn't really matter in the big picture, it just grabbed my attention

go.mod
```diff
@@ -1,23 +1,62 @@
...
+go 1.17
 
 require (
-	cloud.google.com/go/kms v1.4.0 // indirect
-	cloud.google.com/go/pubsub v1.3.1
+	cloud.google.com/go/pubsub v1.21.0
 	github.com/cheynewallace/tabby v1.1.1
 	github.com/docker/go v1.5.1-1.0.20160303222718-d30aec9fd63c
-	github.com/ethereum/go-ethereum v1.10.15
+	github.com/ethereum/go-ethereum v1.10.17
```
doanac, line 9: i'm sure its safe (it's always been), but you might want to run a "fioctl devices config set ...." and make sure your device decrypts its correctly.

subcommands/keys/common.go
```diff
@@ -25,45 +26,142 @@ import (
...
+	pemPrivateKeySuffix = " PRIVATE KEY"
+	// These are case insensitive
+	tufKeyTypeNameEd25519    = "ED25519"
+	tufKeyTypeNameRSA        = "RSA"
+	tufKeyTypeSigNameEd25519 = "ed25519"
+	tufKeyTypeSigNameRSA     = "rsassa-pss-sha256"
+)
+
+var (
+	gtufKeyTypeEd25519 = &tufKeyTypeEd25519{}
```
doanac, line 62: annoying how golang makes you do this. i just went to suggest 3 different alternatives but they are probably more annoying.

subcommands/keys/common.go
```diff
@@ -138,7 +239,7 @@ func GetOfflineCreds(credsFile string) (OfflineCreds, error) {
 	return files, nil
 }
 
-func FindPrivKey(pubkey string, creds OfflineCreds) (*rsa.PrivateKey, error) {
+func FindSigner(keyid, pubkey string, creds OfflineCreds) (*TufSigner, error) {
```
doanac, line 242: I can see you've had HSM's on your mind :)

doanac (commented):
looks like a good start.

subcommands/devices/list.go
vkhoroz: done

go.mod
vkhoroz: Probably I need to reset and run tidy again, maybe with some options...
For some weird reason `go mod tidy` upgraded all packages when I told it to upgrade just Golang.

subcommands/keys/common.go
vkhoroz: yeah... I am quite annoyed that Golang doesn't allow to cast inline structs to an interface... they need a pointer in memory.

subcommands/keys/common.go
vkhoroz: Actually, just trying to find a common demoninator between Golang's crypto/xyz libraries.
In general they are almost incompatible; those `crypto.Signer` interface and `x509.XxxPKCS8yyy` routines are the only things that made my life a little bit easier.

subcommands/keys/common.go
vkhoroz: Done

go.mod
vkhoroz: Fixed this

vkhoroz:
@doanac both commits and the code look much cleaner.
Also, I fixed a subtle bug related to hashing.

Feel free to review while I'm testing this.

doanac:
looks good

vkhoroz:
@doanac this is still not fully tested, I just wanted to save my intermediate work on storing Ed25519 keys as hexdump instead of PKCS8 format.

subcommands/keys/rotate_root.go
```diff
@@ -34,12 +34,14 @@ func init() {
 	rotate.Flags().BoolVarP(&doRootSync, "sync-prod", "", false, "Make sure production root.json is up-to-date and exit")
 	rotate.Flags().BoolVarP(&initialRotation, "initial", "", false, "Used for the first customer rotation. The command will download the initial root key")
 	rotate.Flags().StringVarP(&changeReason, "changelog", "m", "", "Reason for doing rotation. Saved in root metadata for tracking change history")
+	rotate.Flags().StringP("key-type", "k", tufKeyTypeNameRSA, "Key type, supported: Ed25519, RSA (default).")
```
doanac, line 37: i'd put this change and the one to rotate_targets.go into their own commit and i *think* we should probably default to ed25519.

good work.

doanac: works for me.

doanac (approved):

PR #161, opened 2022-05-06
keys: Include a changelog for root key rotation

doanac:
This change includes details for who/when/why root keys were rotated.
The backend API will verify this before committing the change.
Specifically:

 * polis-id *must* be ID of the user making the change
 * timestamp *must* be within 5 minutes of our clocks

### DIFF
--- a/subcommands/keys/rotate_root.go
+++ b/subcommands/keys/rotate_root.go
@@ -23,6 +23,7 @@ import (
 var (
 	doRootSync      bool
 	initialRotation bool
+	changeReason    string
 )
 
 func init() {
@@ -36,6 +37,7 @@ func init() {
 	subcommands.RequireFactory(rotate)
 	rotate.Flags().BoolVarP(&doRootSync, "sync-prod", "", false, "Make sure production root.json is up-to-date and exit")
 	rotate.Flags().BoolVarP(&initialRotation, "initial", "", false, "Used for the first customer rotation. The command will download the initial root key")
+	rotate.Flags().StringVarP(&changeReason, "changelog", "m", "", "Reason for doing rotation. Saved in root metadata for tracking change history")
 	cmd.AddCommand(rotate)
 }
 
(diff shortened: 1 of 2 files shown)


subcommands/keys/rotate_root.go
```diff
@@ -91,6 +96,9 @@ func doRotateRoot(cmd *cobra.Command, args []string) {
 
 	newid, newPk, newCreds := swapRootKey(root, curid, creds)
 	fmt.Println("= New root:", newid)
+	root.Signed.Reason.PolisId = user.PolisId
+	root.Signed.Reason.Message = changeReason
```
mike-sul, line 100: Do we want to limit a size of `changeReason`?

It looks a bit risky to put a user generated content into `root.json` since a user can put anything into that string.
Maybe, this is just my phobias :).

doanac: The backend enforces a 120byte max-len

doanac:
this was just tested again on my staging after being rebased on main. it's good to go.

subcommands/keys/rotate_root.go
```diff
@@ -91,6 +96,11 @@ func doRotateRoot(cmd *cobra.Command, args []string) {
 
 	newid, newPk, newCreds := swapRootKey(root, curid, creds)
 	fmt.Println("= New root:", newid)
+	root.Signed.Reason = &client.RootChangeReason{
+		PolisId:   user.PolisId,
+		Message:   changeReason,
```
mike-sul, line 101: Maybe this is just my phobia, but passing the user input `changeReason` to the signer and then transferring it as a payload of http request to the OTA server looks a bit scary to me :).

doanac: It's a REST API that anyone can call. We have to protect at the API level or its just security theater. Your concern is valid but that becomes a "I want to review our backend" topic.

mike-sul: My concern is that a malicious user may input huge value that has to be transferred and accepted by our backend (and maybe copied a few times before it arrives to the handler) before it's checked by the handler.

But, you're right, such user may just do the same directly via the rest api.

mike-sul (approved):

subcommands/keys/rotate_root.go
doanac: still - doing monthly-ish reviews of nginx proxy settings is always a good idea

PR #166, opened 2022-06-21
targets: Add command to fetch content for offline update

mike-sul:
Add a new command to download content required for an offline update.
1. TUF metadata.
2. Target's Apps.
3. Target's ostree repo.

Refreshing TUF metadata is out of scope atm.
Require extending of ota-lite to download production TUF metadata.

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -254,6 +254,7 @@ type TufCustom struct {
 	ContainersSha  string                `json:"containers-sha,omitempty"`
 	LmpManifestSha string                `json:"lmp-manifest-sha,omitempty"`
 	OverridesSha   string                `json:"meta-subscriber-overrides-sha,omitempty"`
+	Uri            string                `json:"uri,omitempty"`
 	OrigUri        string                `json:"origUri,omitempty"`
 	CreatedAt      string                `json:"createdAt,omitempty"`
 }
@@ -1109,6 +1110,15 @@ func (a *Api) tufRootPost(factory string, prod bool, root []byte) (string, error
 	}
 	return "", err
 }
+
+func (a *Api) TufMetadataGet(factory string, metadata string, tag string, prod bool) (*[]byte, error) {
+	url := a.serverUrl + "/ota/repo/" + factory + "/api/v1/user_repo/" + metadata + "?tag=" + tag
+	if prod {
+		url += "&production=1"
+	}
+	return a.Get(url)
+}
+
 func (a *Api) TufRootPost(factory string, root []byte) (string, error) {
 	return a.tufRootPost(factory, false, root)
 }
--- a/subcommands/targets/offline-update.go
+++ b/subcommands/targets/offline-update.go
@@ -1,0 +1,274 @@
+package targets
+
+import (
+	"archive/tar"
+	"compress/bzip2"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/foundriesio/fioctl/client"
+	"github.com/foundriesio/fioctl/subcommands"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"io"
+	"os"
+	"path"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type (
+	ouTargetInfo struct {
+		version       int
+		ostreeVersion int
+		hardwareID    string
+		buildTag      string
+	}
+)
+
+var (
+	ouTag  string
+	ouProd bool
+)
+
+func init() {
+	offlineUpdateCmd := &cobra.Command{
+		Use:   "offline-update <target-name> <dst> --tag <tag> [--prod]",
+		Short: "Download Target content for an offline update",
+		Run:   doOfflineUpdate,
+		Args:  cobra.ExactArgs(2),
+		Example: `
+	# Download update content of the production target #1451 tagged by "release-01" for "intel-corei7-64" hardware type
+	fioctl targets offline-update intel-corei7-64-lmp-1451 /mnt/flash-drive/offline-update-content --tag release-01 --prod
+
+	# Download update content of the CI target #1451 tagged by "devel" for "raspberrypi4-64" hardware type
+	fioctl targets raspberrypi4-64-lmp-1448 /mnt/flash-drive/offline-update-content --tag devel
+
+	`,
+	}
+	cmd.AddCommand(offlineUpdateCmd)
+	offlineUpdateCmd.Flags().StringVarP(&ouTag, "tag", "", "",
+		"Target tag")
+	offlineUpdateCmd.Flags().BoolVarP(&ouProd, "prod", "", false,
+		"Instruct to fetch content of production Target")
+}
+
+func doOfflineUpdate(cmd *cobra.Command, args []string) {
+	factory := viper.GetString("factory")
+	targetName := args[0]
+	dstDir := args[1]
+
+	if len(ouTag) == 0 {
+		subcommands.DieNotNil(errors.New("missing mandatory flag `--tag`"))
+	}
+
+	fmt.Printf("Checking whether Target exists; target: %s, tag: %s, production: %v\n", targetName, ouTag, ouProd)
+	subcommands.DieNotNil(checkIfTargetExists(factory, targetName, ouTag, ouProd))
+
+	ti, err := getTargetInfo(factory, targetName)
+	subcommands.DieNotNil(err, "Failed to obtain Target's details:")
+
+	fmt.Printf("Downloading offline update content of Target %s to %s\n", targetName, dstDir)
+
+	fmt.Println("Downloading TUF metadata...")
+	subcommands.DieNotNil(downloadTufRepo(factory, ouTag, ouProd, path.Join(dstDir, "tuf")), "Failed to download TUF metadata:")
+
+	fmt.Printf("Downloading an ostree repo from the Target's OE build %d...\n", ti.ostreeVersion)
+	subcommands.DieNotNil(downloadOstree(factory, ti.ostreeVersion, ti.hardwareID, dstDir), "Failed to download Target's ostree repo:")
+
+	fmt.Printf("Downloading Apps fetched by the `assemble-system-image` run; build number:  %d, tag: %s...\n", ti.version, ti.buildTag)
+	subcommands.DieNotNil(downloadApps(factory, targetName, ti.version, ti.buildTag, path.Join(dstDir, "apps")), "Failed to download Target's Apps:")
+
+	fmt.Println("Successfully downloaded offline update content")
+}
+
+func checkIfTargetExists(factory string, targetName string, tag string, prod bool) error {
+	data, err := api.TufMetadataGet(factory, "targets.json", tag, prod)
+	if err != nil {
+		if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
+			return fmt.Errorf("the specified Target has not been found; target: %s, tag: %s, production: %v", targetName, ouTag, ouProd)
+		}
+		return fmt.Errorf("failed to check whether Target exists: %s", err.Error())
+	}
+	targets := client.AtsTufTargets{}
+	err = json.Unmarshal(*data, &targets)
+	if err != nil {
+		return fmt.Errorf("failed to check whether Target exists: %s", err.Error())
+	}
+	for tn := range targets.Signed.Targets {
+		if tn == targetName {
+			return nil
+		}
+	}
+	return fmt.Errorf("the specified Target has not been found; target: %s, tag: %s, production: %v", targetName, ouTag, ouProd)
+}
+
+func getTargetInfo(factory string, targetName string) (*ouTargetInfo, error) {
+	// Getting Target's custom info from the `/ota/factories/<factory>/targets/<target-name>` because:
+	// 1. a target name is unique and represents the same Target across all "tagged" targets set including prod;
+	// 2. only this target version/representation contains an original tag(s)/branch that
+	// the `image-assemble` and apps fetching was performed for (needed for determining where to download Apps from).
+	custom, err := getTargetCustomInfo(factory, targetName)
+	if err != nil {
+		return nil, err
+	}
+
+	info := ouTargetInfo{}
+	info.version, err = strconv.Atoi(custom.Version)
+	if err != nil {
+		return nil, err
+	}
+	info.hardwareID = custom.HardwareIds[0]
+	info.buildTag = custom.Tags[0] // See the assemble.py script in ci-scripts https://github.com/foundriesio/ci-scripts/blob/18b4fb154c37b6ad1bc6e7b7903a540b7a758f5d/assemble.py#L300
+	info.ostreeVersion = info.version
+	if len(custom.OrigUri) > 0 {
+		indx := strings.LastIndexByte(custom.OrigUri, '/')
+		if indx == -1 {
+			return nil, fmt.Errorf("failed to determine Target's OE build version: %s", custom.OrigUri)
+		}
+		targetOstreeVerStr := custom.OrigUri[indx+1:]
+		info.ostreeVersion, err = strconv.Atoi(targetOstreeVerStr)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return &info, nil
+}
+
+func downloadTufRepo(factory string, tag string, prod bool, dstDir string) error {
+	ver := 1
+
+	downloadMetadataFile := func(metadataFileName string) error {
+		data, err := api.TufMetadataGet(factory, metadataFileName, tag, prod)
+		if err != nil {
+			return err
+		}
+		f, err := os.Create(path.Join(dstDir, metadataFileName))
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		_, err = f.Write(*data)
+		return err
+	}
+
+	err := os.MkdirAll(dstDir, 0755)
+	if err != nil {
+		return err
+	}
+
+	for {
+		metadataFileName := fmt.Sprintf("%d.root.json", ver)
+		err := downloadMetadataFile(metadataFileName)
+		if err != nil {
+			httpErr := client.AsHttpError(err)
+			if httpErr != nil {
+				// TODO: Fix an issue in ota-lite that returns 500 for non-existing non-prod N.root.json,
+				// and then remove 500 check here
+				if httpErr.Response.StatusCode != 404 && httpErr.Response.StatusCode != 500 {
+					return err
+				}
+				break
+			}
+			return err
+		}
+		ver += 1
+	}
+
+	metadataFileNames := []string{
+		"timestamp.json", "snapshot.json", "targets.json",
+	}
+	for _, file := range metadataFileNames {
+		err = downloadMetadataFile(file)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func downloadOstree(factory string, targetVer int, hardwareID string, dstDir string) error {
+	runName := hardwareID
+	artifactName := hardwareID + "-ostree_repo.tar.bz2"
+	artifactPath := path.Join("other", artifactName)
+
+	return downloadItem(factory, targetVer, runName, artifactPath, func(r io.Reader) error {
+		bzr := bzip2.NewReader(r)
+		if bzr == nil {
+			return fmt.Errorf("failed to create bzip2 reader")
+		}
+		return untar(bzr, dstDir)
+	})
+}
+
+func downloadApps(factory string, targetName string, targetVer int, tag string, dstDir string) error {
+	runName := "assemble-system-image"
+	artifactPath := path.Join(tag, targetName+"-apps.tar")
+
+	return downloadItem(factory, targetVer, runName, artifactPath, func(r io.Reader) error {
+		return untar(r, dstDir)
+	})
+}
+
+func getTargetCustomInfo(factory string, targetName string) (*client.TufCustom, error) {
+	targetFile, err := api.TargetGet(factory, targetName)
+	if err != nil {
+		return nil, err
+	}
+	custom, err := api.TargetCustom(*targetFile)
+	return custom, err
+}
+
+func downloadItem(factory string, targetVer int, runName string, artifactPath string, storeHandler func(r io.Reader) error) error {
+	resp, err := api.JobservRunArtifact(factory, targetVer, runName, artifactPath)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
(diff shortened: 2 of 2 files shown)


subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,264 @@
...
+
+	fmt.Printf("Getting info about Target %s \n", targetName)
+
+	// We get the Target's custom data from the list of all targets regardless whether it's a production target or not
+	// because we need to obtain the Target's tag that was used to build OE and Apps/containers, which in turn is needed
+	// to deduce a path to the Apps archive artifact.
+	// E.g. https://ci.foundries.io/projects/msul-dev01/lmp/builds/1420/assemble-system-image/ is built as `master`
+	// while in the production targets the corresponding target is marked as `prod`.
+	// TODO: switch to usage of /api/v1/user_repo/targets.json API once it supports `production` param
+	// instead of doing the stuff below
```
vkhoroz, line 54: Lets extend the API right now in scope of this feature.

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,264 @@
...
+	custom, err := getTargetCustomInfo(targetName)
```
vkhoroz, line 55: I'm not sure how is this supposed to work, especially for the prod targets (which require both online and an offline signatures).

I believe that for offline update we should mimic what the device does during an online update as close as possible in order to be fully interoperable and respect the TUF guidelines. Respective this line I think we should fetch targets.json by tag (just like the device does) and search for all necessary information inside that specific tag-based targets.json file, not the CI targets.json. The device never downloads the CI targets.json, and in fact data in it may be different from the tag-based targets.json, especially when we take production targets into consideration.

We have an API to get tag-based TUF roles already (your new `TufMetadataGet` method does it right), just need to extend it with the production targets support. An interesting use case would be to extend it with the wave targets support (by wave name or device group name???).

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,264 @@
...
+		Use:   "offline-update <target-name> <dst>",
+		Short: "Download Target content for an offline update",
+		Run:   doOfflineUpdate,
+		Args:  cobra.ExactArgs(2),
+	}
+	cmd.AddCommand(offlineUpdateCmd)
+	offlineUpdateCmd.Flags().BoolVarP(&prod, "prod", "", false,
+		"Instruct to fetch content of production Target")
+	offlineUpdateCmd.Flags().StringVarP(&tag, "tag", "", "master",
+		"Target tag")
```
vkhoroz, line 36: The tag parameter is the key here so I'd recommend to make it mandatory. The doesn't just fetch "any" targets, it fetches targets for the specific tag. In particular, for prod targets different tags may contain slightly different target items of the same name; and they are almost 100% of the time different from the CI targets of the same name.

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	fmt.Printf("Downloading an ostree repo from the Target's OE build %d...\n", targetOstreeVer)
+	err = downloadOstree(targetOstreeVer, hardwareID, dstDir)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to download Target ostree repo: %s", err.Error())
+		os.Exit(1)
+	}
+
+	fmt.Printf("Downloading Apps fetched by the `assemble-system-image` run of the build:  %d...\n", targetVer)
+	err = downloadApps(targetName, targetVer, buildTag, path.Join(dstDir, "apps"))
```
vkhoroz, line 105: IIRC most of our updates are either ostree-only or apps-only i.e. an ostree update versus an apps update are normally delivered as two separate targets to the devices (via online updates). I think that offline updates should support that behavior as well (i.e. allow ostree-only or apps-only updates). AFAIK in case of waves and production targets this is the only way how it works i.e. a user delivers one or more ostree-only updates followed by one or more apps-only updates for each tag.

I don't know how it is planned to implement the offline update on the device side, but IMHO supporting ostree-only and apps-only updates is at least a necessity, and at most may be even the only supported way to do this.

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,264 @@
...
+	targetFile, err := api.TargetGet(factory, targetName)
+	if err != nil {
+		return nil, err
+	}
+	custom, err := api.TargetCustom(*targetFile)
+	return custom, err
+}
+
+func downloadTufRepo(factory string, tag string, prod bool, dstDir string) error {
+	ver := 1 // TODO: start from the second version once the first one is provisioned on a device during manufacturing process.
```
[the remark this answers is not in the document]
mike-sul: > and let the device decide which ones it wants to take from the offline source.

Yep, although downloading and carrying 1.root.json on a flash disk looks like a security flaw to me.

Also, during manufacturing process, the initial root version must be provisioned starting from the first version, otherwise aklite will fail.

subcommands/targets/offline-update.go
mike-sul: > Respective this line I think we should fetch targets.json by tag (just like the device does) 

This line does NOT fetch targets.json for an offline update, the downloaded targets as well as the other role metadata are fetched via the API you suggest (api/v1/user_repo/).

> getTargetCustomInfo is needed solely to get a tag (effectively branch) that was used at a CI build. The corresponding prod target has different tag than the tag that was used to build the given target. Therefore, we cannot use prod targets for it, only CI one.

I don't like it too, so I think we should change the way we store the apps artifact in a CI run so it is tag agnostic...

subcommands/targets/offline-update.go
mike-sul: Yep, it makes sense to make the `tag` param mandatory.

subcommands/targets/offline-update.go
mike-sul: It implies knowing exactly which version is currently running on each target device, and those versions may differ. How to guarantee that the Apps of a given offline update can be supported by all ostree versions devices currently running and vice versa.

It's not big deal to support it, I'd postpone with this feature customization once we get some feedback about initial implementation. Maybe we'll get zero customers for it :).

subcommands/targets/offline-update.go
vkhoroz: > > Respective this line I think we should fetch targets.json by tag (just like the device does)
> 
> This line does NOT fetch targets.json for an offline update, the downloaded targets as well as the other role metadata are fetched via the API you suggest (api/v1/user_repo/).
> 
> > getTargetCustomInfo is needed solely to get a tag (effectively branch) that was used at a CI build. The corresponding prod target has different tag than the tag that was used to build the given target. Therefore, we cannot use prod targets for it, only CI one.
> 
> I don't like it too, so I think we should change the way we store the apps artifact in a CI run so it is tag agnostic...

I think I understand the use case, and it troubles me even more this way, as this code ties the user to our CI artifacts storage. I agree it may be a convenient way to simplify the implementation. However, this looks like a completely different way to fetch targets artifacts compared to the way device does that.

Is it possible to fetch those artifacts from the same storage the device does, e.g. implement the ostree-proxy like signed-url handler in ota-lite or support the user auth in ostree-proxy (like we do for e.g. hub.fio)?

I suspect that could turn into a "feature" on itself, and we might have zero customers for this. That said, you can ignore this comment, I'm just trying to think of 100500 use cases which could be completely irrelevant to you work.

mike-sul: Yep, there are many aspects, it's better to discuss it offline.

mike-sul:
@vkhoroz @doanac It's ready for review.

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func doOfflineUpdate(cmd *cobra.Command, args []string) {
+	factory = viper.GetString("factory")
+	targetName := args[0]
+	dstDir := args[1]
+
+	if len(tag) == 0 {
+		fmt.Fprintln(os.Stderr, "Mandatory flag `tag` is not specified")
+		os.Exit(1)
```
doanac, line 56: i'm not a huge fan of our approach, but we generally error out on bad CLI usage with something like:
```
subcommands.DieNotNil(errors.New("--tag is required"))
```

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,279 @@
...
+		buildTag      string
+	}
+)
+
+func init() {
+	offlineUpdateCmd := &cobra.Command{
+		Use:   "offline-update <target-name> <dst> --tag <tag> [--prod]",
+		Short: "Download Target content for an offline update",
+		Run:   doOfflineUpdate,
+		Args:  cobra.ExactArgs(2),
```
doanac, line 40: you might include an `Example` attribute (see how we do one in subcommands/config/set.got) to help emphasize target-name is something like `intel-corei7-64-lmp-133`.

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,279 @@
...
+		fmt.Fprintln(os.Stderr, "Mandatory flag `tag` is not specified")
+		os.Exit(1)
+	}
+
+	fmt.Printf("Checking whether Target exists; target: %s, tag: %s, production: %v\n", targetName, tag, prod)
+	ok, err := checkIfTargetExists(targetName)
+	subcommands.DieNotNil(err, "Failed to check whether Target exists:")
+	if !ok {
+		fmt.Fprintf(os.Stderr, "The specified Target has not been found; target: %s, tag: %s, production: %v\n", targetName, tag, prod)
+		os.Exit(1)
```
doanac, line 64: again - volodya converted all our code to exit prematurely with `subcommands.DieNotNil`. Let's stay with that for now. You could probably just have `checkIfTargetExists` return just an error and skip the OK. Or just call it `assertTargetExists` and die there.

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,279 @@
...
+	fmt.Printf("Downloading an ostree repo from the Target's OE build %d...\n", params.ostreeVersion)
+	err = downloadOstree(params.ostreeVersion, params.hardwareID, dstDir)
+	subcommands.DieNotNil(err, "Failed to download Target's ostree repo:")
+
+	fmt.Printf("Downloading Apps fetched by the `assemble-system-image` run of the build:  %d...\n", params.version)
+	err = downloadApps(targetName, params.version, params.buildTag, path.Join(dstDir, "apps"))
+	subcommands.DieNotNil(err, "Failed to download Target's Apps:")
+
+	fmt.Println("Successfully downloaded offline update content")
+	os.Exit(0)
```
doanac, line 85: os.Exit isn't needed

client/foundries.go
```diff
@@ -1109,6 +1110,19 @@ func (a *Api) tufRootPost(factory string, prod bool, root []byte) (string, error
...
+func (a *Api) TufMetadataGet(factory string, metadata string, tag string, prod bool) (*[]byte, error) {
+	url := a.serverUrl + "/ota/repo/" + factory + "/api/v1/user_repo/" + metadata + "?tag=" + tag
+	if prod {
+		url += "&production=1"
+	}
+	body, err := a.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	return body, nil
```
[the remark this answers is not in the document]
mike-sul: Body is not `nil` if err is not nil. Although, with a new version it may be not important, let me check it.

subcommands/targets/offline-update.go
mike-sul: `checkIfTargetExists` fetches the tagged condensed Target, what a device would fetch, `getTargetInfo` fetches the original CI target in order to get the correct branch/tag.

mike-sul: > again - volodya converted all our code to exit prematurely with subcommands.DieNotNil

Makes sense to some extend.

client/foundries.go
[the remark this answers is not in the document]
mike-sul: Indeed this updated version can just `return`, updated the PR accordingly.

subcommands/targets/offline-update.go
doanac: i'm still not seeing why you can't do something like get the whole list of targets and return the `client.CustomTarget` for it rather than "ok" and pass that to `getTargetInfo`. I don't see what you are gaining by two round-trips to the server that can't be done in one trip?

mike-sul: The targets downloaded for offline update do not contain the target custom required here (with original CI tag), so there is no two requests to the same handler. The TUF targets are downloaded from the repo API, while the target custom details are obtained from the `/ota/factories/<factory>/targets/target` call.

mike-sul: exactly, e.g.

```
curl -s -H "OSF-TOKEN: `cat $FIO_TOKEN`" "${FIO_API}/repo/$FACTORY/api/v1/user_repo/targets.json?tag=beta&production=1" | jq '.signed.targets."raspberrypi4-64-lmp-1452"."custom"."tags"'
[
  "beta"
]
[mike@t14s ota-lite]$ curl -s -H "OSF-TOKEN: `cat $FIO_TOKEN`" "${FIO_API}/repo/$FACTORY/api/v1/user_repo/targets.json?tag=beta" | jq '.signed.targets."raspberrypi4-64-lmp-1452"."custom"."tags"'
[
  "master",
  "beta"
]

```

And making `if !prod` to optimize leads to more source code and less readability, and we think that the prod use-case will be prevailing anyway.

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
vkhoroz: IMHO if we're talking about security here - then both 1.root.json and 2.root.json (I'd say 3.root.json too) should be pre-provisioned, as version 2 is the first one we trust, and version 3 is the first one customer can trust:
- 1.root.json - auto-generated by tuf-reposerver;
- 2.root.json - we took root and targets role keys away from tuf-reposerver i.e. generated new online keys;
- 3.root.json - the customer took the root keys offline (the first secure version from the customer perspective).

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,286 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+var (
+	factory string
+	prod    bool
+	tag     string
+)
```
vkhoroz, line 25: These are globals for the entire `subcommands.targets` subpackage.
So I'd recommend to either prefix them somehow (to make less generic) or use the `XxxVar` rather than `XxxVarP` below and pass into functions explicitly.
I generally prefer the second way as it is safer (although more verbose and less convenient to someone).

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,286 @@
...
+	if len(tag) == 0 {
+		subcommands.DieNotNil(errors.New("missing mandatory flag `--tag`"))
+	}
+
+	fmt.Printf("Checking whether Target exists; target: %s, tag: %s, production: %v\n", targetName, tag, prod)
+	ok, err := checkIfTargetExists(targetName)
+	subcommands.DieNotNil(err, "Failed to check whether Target exists:")
+	if !ok {
+		subcommands.DieNotNil(fmt.Errorf("The specified Target has not been found; target: %s, tag: %s, production: %v\n", targetName, tag, prod))
+	}
```
vkhoroz, line 72: I still don't get why do you need the `ok, err := checkIfTargetExists(targetName)` above.
You might just `return fmt.Errorf("The specified Target has not been found ...")` where you currently `return false, nil` from inside `checkIfTargetExists` to make the code flow simpler.

subcommands/targets/offline-update.go
mike-sul: I'll do the explicit way.

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,286 @@
...
+		return nil, err
+	}
+
+	info := targetInfo{}
+	info.version, err = strconv.Atoi(custom.Version)
+	if err != nil {
+		return nil, err
+	}
+	info.hardwareID = custom.HardwareIds[0]
+	info.buildTag = custom.Tags[0] // See the assemble.py script in ci-scripts https://github.com/foundriesio/ci-scripts/blob/18b4fb154c37b6ad1bc6e7b7903a540b7a758f5d/assemble.py#L300
```
vkhoroz, line 133: I think we need to set the `custom.apps_artifact` target attribute instead inside the ci-scripts; actually I think we need to set the `custom.ostree_artifact` target attribute too.  Otherwise, we're not enough safe.

Yes, I know that it means changes to the ci-scripts and would require the customer to switch to a new version of ci-scripts before using offline updates. But I don't think it creates any onerous problems for us.

The other option would be to not rely onto the jobserv and instead download these artifacts from the GCP buckets instead.  Not sure how feasible this approach is, however, as we allow user access to the hub.fio but not to the ostree-proxy.

One more thing to consider is how long we keep these artifacts on the jobserv shares? Do we prune them periodically? If yes - the current solution might not work as expected in long run.

subcommands/targets/offline-update.go
mike-sul: looks like it should work.

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,286 @@
...
+		metadataFileName := fmt.Sprintf("%d.root.json", ver)
+		err := downloadMetadataFile(metadataFileName)
+		if err != nil {
+			httpErr := client.AsHttpError(err)
+			if httpErr != nil {
+				// TODO: Fix an issue in ota-lite that returns 500 for non-existing non-prod N.root.json,
+				// and then remove 500 check here
+				if httpErr.Response.StatusCode != 404 && httpErr.Response.StatusCode != 500 {
+					return err
+				}
```
vkhoroz, line 182: Please, do us a favor - fix that ;-)
Thanks!

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,286 @@
...
+		return nil, err
+	}
+	info.hardwareID = custom.HardwareIds[0]
+	info.buildTag = custom.Tags[0] // See the assemble.py script in ci-scripts https://github.com/foundriesio/ci-scripts/blob/18b4fb154c37b6ad1bc6e7b7903a540b7a758f5d/assemble.py#L300
+	info.ostreeVersion = info.version
+	if len(custom.OrigUri) > 0 {
+		indx := strings.LastIndexByte(custom.OrigUri, '/')
+		if indx == -1 {
+			fmt.Fprintf(os.Stderr, "Failed to determine Target's OE build version: %s", custom.OrigUri)
+			os.Exit(1)
```
vkhoroz, line 139: Why not `return nil, fmt.Errorf(...)`?

The reason we don't do `fmt.Fprintf` for errors is consistency. The `return err` and `DieNotNil` things were called to give our error messages at least some degree of the consistent look-and-feel (the primary intent not to save code lines as one might think).

subcommands/targets/offline-update.go
doanac: jobserv artifacts are in a gcp bucket, the api just gives you a signed url to them. and we don't delete them until a factory is deleted.

mike-sul: > I think we need to set the `custom.apps_artifact` target attribute instead inside the ci-scripts; actually I think we need to set the `custom.ostree_artifact` target attribute too. Otherwise, we're not enough safe.

It makes sense to do something like that once we agree that we "really" need this feature.

mike-sul: > The other option would be to not rely onto the jobserv and instead download these artifacts from the GCP buckets instead. Not sure how feasible this approach is, however, as we allow user access to the hub.fio but not to the ostree-proxy.

An ostree repo downloading via the ostree service/proxy looks quite feasible, not sure about Apps downloading since it depends on `skopeo` and secrets/auth to get access to 3rd party registries.

subcommands/targets/offline-update.go
mike-sul: Sure, but not in this PR :).

subcommands/targets/offline-update.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+func checkIfTargetExists(targetName string) (bool, error) {
+	data, err := api.TufMetadataGet(factory, "targets.json", tag, prod)
+	if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
```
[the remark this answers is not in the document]
vkhoroz: No worries, both ways truly work; just one looks more intuitive to me.
It is truly subjective, sorry for nitpicking.

subcommands/targets/offline-update.go
mike-sul: Good catch, it seems like leftover from the previous version.

subcommands/targets/offline-update.go
vkhoroz: I mean: please, fix that and update this PR.

Sorry for that bug, we just cannot ship with it as is :( .

subcommands/targets/offline-update.go
mike-sul: Actually it does not work exactly as I would like it, no way to distinguish between error case and non error case when targets are not found.
IMO, it is not correct to return an error if the check was successful but a specified target was not found.

vkhoroz: Hmm... I don't quite understand what the problem is?

IIUC from the perspective of _this script_ no found target is an error because it cannot continue, isn't it?
If yes, then there is no reason to distinguish between _this error_ versus _other error_, other than a bit different error message: in any case the program should terminate with the same exit code.

This is scripting application, so I don't buy purely philosophical categories like exception versus error; lets stay practical.

subcommands/targets/offline-update.go
mike-sul: done.

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
mike-sul: I actually did in the way you proposed.

subcommands/targets/offline-update.go
mike-sul: done.

subcommands/targets/offline-update.go
mike-sul: This bug is already there and is not a blocker for offline update at all.
It's not a big deal to fix it, but it takes time, so maybe we deliver this to the support/platform team at first so they start integration conversation with the customer. Meanwhile, I'll fix this issue (the fix in ota-lite is trivial, just testing doesn't work as I expected...)

vkhoroz:
Seems good.

My only comments go fo the places outside this PR; and it makes sense to treate them as _out of scope_ as this is PoC.

mike-sul:
Can I merge it?

PR #167, opened 2022-06-28
config-updates: Require a device-group

doanac:
We are going to start requiring device-groups when setting aklite update
settings. Doing fleet wide is too risky.


vkhoroz (commented):
LGTM

PR #169, opened 2022-07-15
targets: Stop root meta fetching if 404 received

mike-sul:
Remove stopping of root meta fetching if 500 is received.
The issue in ota-lite was fixed so we should stop the fetch only if 404
is received. Any other response status codes are considered as errors.


vkhoroz:
LGTM

PR #170, opened 2022-07-21
keys: Update "reason" if target key is rotated

mike-sul:
Update the root role's reason field if a new version is generated
because a target role key was rotated.

### DIFF
--- a/subcommands/keys/rotate_targets.go
+++ b/subcommands/keys/rotate_targets.go
@@ -3,6 +3,7 @@ package keys
 import (
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -30,12 +31,14 @@ This command is not allowed if there is an active wave in your factory.`,
 	}
 	subcommands.RequireFactory(rotateTargets)
 	rotateTargets.Flags().StringP("key-type", "k", tufKeyTypeNameRSA, "Key type, supported: Ed25519, RSA (default).")
+	rotateTargets.Flags().StringP("changelog", "m", "", "Reason for doing rotation. Saved in root metadata for tracking change history")
 	cmd.AddCommand(rotateTargets)
 }
 
 func doRotateTargets(cmd *cobra.Command, args []string) {
 	factory := viper.GetString("factory")
 	keyTypeStr, _ := cmd.Flags().GetString("key-type")
+	changeLog, _ := cmd.Flags().GetString("changelog")
 	keyType := ParseTufKeyType(keyTypeStr)
 	credsFile := args[0]
 	assertWritable(credsFile)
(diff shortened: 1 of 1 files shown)


subcommands/keys/rotate_targets.go
```diff
@@ -60,6 +61,13 @@ func doRotateTargets(cmd *cobra.Command, args []string) {
 	targetid, newCreds := replaceOfflineTargetKey(root, onlineTargetId, creds, keyType)
 	fmt.Println("= New target:", targetid)
 	removeUnusedKeys(root)
+	user, err := api.UserAccessDetails(factory, "self")
+	subcommands.DieNotNil(err)
+	root.Signed.Reason = &client.RootChangeReason{
+		PolisId:   user.PolisId,
+		Message:   "Target role key rotation",
```
vkhoroz, line 68: nitpick: a TUF role name is `targets`, not `target`.

vkhoroz (approved):
LGTM

subcommands/keys/rotate_targets.go
mike-sul: Sure, always make this mistake.

subcommands/keys/rotate_targets.go
```diff
@@ -60,6 +61,13 @@ func doRotateTargets(cmd *cobra.Command, args []string) {
 	targetid, newCreds := replaceOfflineTargetKey(root, onlineTargetId, creds, keyType)
 	fmt.Println("= New target:", targetid)
 	removeUnusedKeys(root)
+	user, err := api.UserAccessDetails(factory, "self")
+	subcommands.DieNotNil(err)
+	root.Signed.Reason = &client.RootChangeReason{
+		PolisId:   user.PolisId,
+		Message:   "Targets role key rotation",
+		Timestamp: time.Now(),
+	}
```
doanac, line 70: sorry - i missed this. We should probably try and make the CLI consistent with:

 https://github.com/foundriesio/fioctl/commit/b6ca47189c8764f0b335b32eb0530774ba3c1559

doanac: i.e. - the user can optionally provide a change-reason

mike-sul: What would be the help string in this case? Or just us the same as for the root rotation cmd: `"Reason for resigning root.json. Saved in root metadata for tracking change history"`?

doanac: i think the default reason you have is fine. I'm more saying we need the same flag as that command so that they can do a `--reason "Cause Johnny said he lost the key and we had to fire him"`

mike-sul: I meant the help message :)

doanac: ah - i bet you can just copy the one as-is

doanac (approved):

PR #171, opened 2022-07-21
subcommands: Fix nil pointer issue in ca-show

doanac:
The `keys ca show --pretty` has a null pointer issue when users have
run `keys ca update` with a file that has a bunch of trailing
whitespace. This detects that situation and prevents the crash.


vkhoroz (approved):
OMG! Good catch!

PR #176, opened 2022-08-31
Bring in el2g support

doanac:


subcommands/el2g/gateway.go
```diff
@@ -0,0 +1,72 @@
...
+	subcommands.DieNotNil(err)
+
+	ca, err := api.FactoryGetCA(factory)
+	subcommands.DieNotNil(err)
+
+	tmpfile, err := ioutil.TempFile("", "el2g-*.csr")
+	subcommands.DieNotNil(err)
+	defer os.Remove(tmpfile.Name())
+
+	fmt.Println("Requesting CSR from Edgelock2Go")
```
vkhoroz, line 50: type: Edgelock2Go -> EdgeLock2Go.

subcommands/el2g/gateway.go
```diff
@@ -0,0 +1,72 @@
...
+	sign.Dir = pkiDir
+	var out bytes.Buffer
+	sign.Stdout = &out
+	sign.Stderr = os.Stderr
+	subcommands.DieNotNil(sign.Run())
+
+	fmt.Println("Uploading signed certificate")
+	subcommands.DieNotNil(api.El2gUploadDgCert(factory, csr.Id, ca.RootCrt, out.String()))
+
+	fmt.Println("Upadating factory allowed CA's with")
```
vkhoroz, line 67: typo: Upadating -> Updating

client/foundries_el2g.go
```diff
@@ -0,0 +1,47 @@
...
+		SignedCertificate: cert,
+	})
+	if err != nil {
+		return err
+	}
+	_, err = a.Put(url, body)
+	return err
+}
+
+func (a *Api) El2gDeleteDg(factory string) error {
```
vkhoroz, line 59: Is it use anywhere?

subcommands/el2g/gateway.go
```diff
@@ -0,0 +1,72 @@
...
+	fmt.Println(out.String())
+	newCa := ca.CaCrt + "\n" + out.String()
+	certs := client.CaCerts{CaCrt: newCa}
+	subcommands.DieNotNil(api.FactoryPatchCA(factory, certs))
```
vkhoroz, line 71: IIUC this appends the El2Go CA cert to the regular factory CA cert.
Does this work well on the `device-gateway` side?

I believe you've already tested this; just double-checking.

subcommands/el2g/aws.go
```diff
@@ -0,0 +1,74 @@
...
+		"--new-status", "ACTIVE",
+		"--new-auto-registration-status", "ENABLE",
+	)
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+	subcommands.DieNotNil(c.Run())
+	fmt.Println(" |-> ACTIVE")
+}
+
+func run(args ...string) map[string]string {
```
vkhoroz, line 65: This looks like a good generic addition to `common.go`.

subcommands/el2g/aws.go
```diff
@@ -0,0 +1,74 @@
...
+	fmt.Println(" |-> AWS Certificate ID", certId)
+
+	c := exec.Command("/usr/bin/env", "aws", "iot", "update-ca-certificate",
+		"--certificate-id", certId,
+		"--new-status", "ACTIVE",
+		"--new-auto-registration-status", "ENABLE",
+	)
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+	subcommands.DieNotNil(c.Run())
```
vkhoroz, line 61: These lines look almost the same as the `run` routine below.

I'd say we need these two functions:

```
jsonOut = readJsonDict(subcommands.RunCmd("/usr/bin/env", "aws", "iot", "register-ca-certificate", ...))
plainOut = subcommands.RunCmd("/usr/bin/env", "aws", "iot", "update-ca-certificate", ...)
```

The `RunCmd` seems enough generic to get into `common.go`; while `readJsonDict` looks specific to this module.

subcommands/el2g/devices.go
```diff
@@ -31,3 +45,22 @@ func doDevices(cmd *cobra.Command, args []string) {
...
+	factory := viper.GetString("factory")
+	deviceId := args[0]
+
+	info, err := api.El2gProductInfo(factory, deviceId)
+	subcommands.DieNotNil(err)
+	fmt.Println("Hardware Type:", info.Type)
+	fmt.Println("Hardware 12NC:", info.Nc12)
+
+	fmt.Println("Secure Objects:")
+	objects, err := api.El2gSecureObjectProvisionings(factory, deviceId)
```
vkhoroz, line 93: It seems strange that we need to call 2 API endpoints to get info about the single device.
Is it possible to have a single API for that?

subcommands/el2g/devices.go
```diff
@@ -26,9 +28,18 @@ func init() {
 fioctl el2g devices 
 
 # Show details of a device
-fioctl el2g devices  <device-id>
+fioctl el2g devices <device-id>
```
vkhoroz, line 36: This typo correction seems to belong to the previous commit.

subcommands/el2g/gateway.go
doanac: 1480471b31a089b4680713e7780f8c951b089f9b

subcommands/el2g/devices.go
```diff
@@ -26,9 +28,18 @@ func init() {
...
+# Add a device with an se050 (product ID: 935389312472)
+# Device ID can be found on a device by running:
+#  $ ssscli se05x uid | grep "Unique ID:" | cut -d: -f2
+#  ssscli se05x uid | grep "Unique ID:" | cut -d: -f2
+#  04005001eee3ba1ee96e60047e57da0f6880
+# This ID is hexadecimal and must be prefixed in the CLI with 0x. For example:
+fioctl el2g devices --add 935389312472 0x04005001eee3ba1ee96e60047e57da0f6880
 `,
 	}
+	devicesCmd.Flags().StringVarP(&add, "add", "", "", "Whitelist the device for the given nc12 product id")
```
vkhoroz, line 50: I don't understand from the help message what this command really does.

It also looks strange that it is implemented as a flag `--add` and not as a subcommand `add`.

subcommands/el2g/gateway.go
doanac: a113da2478dc2e125b26a58eb9e573e60b245bc4

client/foundries_el2g.go
doanac: good catch b7d34016ff64f205b967e242b2f8d6c4824526f0

subcommands/el2g/gateway.go
doanac: yes. this has been tested for both ci and prod devices

subcommands/el2g/devices.go
```diff
@@ -37,14 +41,22 @@ fioctl el2g devices <device-id>
 #  04005001eee3ba1ee96e60047e57da0f6880
 # This ID is hexadecimal and must be prefixed in the CLI with 0x. For example:
 fioctl el2g devices --add 935389312472 0x04005001eee3ba1ee96e60047e57da0f6880
+
+# Delete a device with an se050
+fioctl el2g devices --del 935389312472 <device-id>
 `,
 	}
 	devicesCmd.Flags().StringVarP(&add, "add", "", "", "Whitelist the device for the given nc12 product id")
+	devicesCmd.Flags().StringVarP(&del, "del", "", "", "Unclaim device for the given nc12 product id")
```
vkhoroz, line 51: IMHO this creates the same confusion as the `--add`; besides, implementing these as mutually-exclusive flags makes the implementation more complex to understand or extend.

client/foundries_el2g.go
```diff
@@ -69,3 +69,23 @@ func (a *Api) El2gConfigAws(factory string, awsRegistrationCode string) (El2gAWS
...
 	return cert, err
 }
+
+type El2gDevice struct {
+	DeviceGroup    string      `json:"device-group"`
+	Id             json.Number `json:"id"`
+	LastConnection string      `json:"last-connection"`
+}
+
+// TODO paginate
```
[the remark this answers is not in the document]
doanac: i think i'll do this after this PR to help keep the scope narrow

subcommands/el2g/status.go
```diff
@@ -0,0 +1,63 @@
...
+	cas, err := api.El2gIntermediateCas(factory)
+	subcommands.DieNotNil(err)
+	for i, ca := range cas {
+		if i > 0 {
+			fmt.Println()
+		}
+		fmt.Println("Name:", ca.Name)
+		fmt.Println("Algorithm:", ca.Algorithm)
+		fmt.Println(ca.Value)
+	}
```
vkhoroz, line 63: In general I'm not a big fan of one command calling multiple endpoints to gather all information it needs; but in this specific use case it seems OK.

subcommands/el2g/devices.go
doanac: their API is really limited in this regard. i've hidden some of these multi-round trip APIs in our backend where its hard to execute them correctly and in the proper order. But I decided this one didn't warrant it.

subcommands/keys/ca_show.go
```diff
@@ -26,6 +29,7 @@ func init() {
 	}
 	caCmd.AddCommand(cmd)
 	cmd.Flags().BoolVarP(&prettyFormat, "pretty", "", false, "Display human readable output of each certificate")
+	cmd.Flags().BoolVarP(&justRoot, "just-root", "", false, "Only show the Factory root CA certificate")
```
vkhoroz, line 32: Then we need `just-tls` and `just-device-ca`; otherwise, it is unfair.

client/foundries_el2g.go
```diff
@@ -106,33 +106,35 @@ func (a *Api) El2gDevices(factory string) ([]El2gDevice, error) {
...
+	body, err := json.Marshal(Req{prodId, devices, production})
 	if err != nil {
 		return err
 	}
 	_, err = a.Post(url, body)
 	return err
 }
 
-func (a *Api) El2gDelDevice(factory, prodId, deviceUuid string) error {
+func (a *Api) El2gDelDevice(factory, prodId, deviceUuid string, production bool) error {
```
vkhoroz, line 121: Why do we need to pass the `production` flag into the `delete` operation?

doanac: its a different device group in their product. w/o that flag i don't know where to find the device i need to delete. We can try to hide it. then you make 3 api calls, "find this device here? find it here?, okay delete that one".

subcommands/el2g/devices.go
doanac: i've grown sick of the `cmd list|add|show|del`. but i guess i should re-work that to make it consistent.

vkhoroz: I think the `cmd list|add|show|del` is that case when it is 10% harder to write the initial code, but 90% easier to support it later.

client/foundries_el2g.go
vkhoroz: - On the back-end when we fetch the device from the database, doesn't we know if it is production or not?
- Can the non-production device belong to the production El2Go product?

I feel like I'm definitely missing something here.

doanac: this isn't our database. these commands are for managing the device inside el2g's product. The device might not even exist yet in our product.

subcommands/el2g/devices.go
vkhoroz: ah, that...

I think one day we could hack the `cobra` to default `fioctl devices` to `fioctl devices list`, and make that change for all commands where it is relevant.

subcommands/keys/ca_show.go
doanac: i'll split it out. I started to do this and I remember why i got lazy - mutually exclusive flags aren't released in Cobra yet - https://github.com/spf13/cobra/pull/1654/commits/c437f2ca77a308e612d1ace6295d42c0791c7ac7.

anway - i can do it by hand. just a little inconvenience.

client/foundries_el2g.go
vkhoroz: yay, that's more complex than I thought... just ignore it.

subcommands/el2g/aws.go
doanac: I've never had much luck trying to make generic run functions. In this case they differ in handling of stdout/stderr. the next caller will need to override the working dir or set an environment variable. You wind up just needing the os/exec interface.

PR #177, opened 2022-09-01
keys: Add ability to show subset of certs

doanac:
Cobra does not yet have mutually exclusive arguments, so I bolted this
into our common code.

### DIFF
--- a/subcommands/keys/ca_show.go
+++ b/subcommands/keys/ca_show.go
@@ -16,7 +16,10 @@ import (
 	"github.com/foundriesio/fioctl/subcommands"
 )
 
-var prettyFormat bool
+var (
+	prettyFormat  bool
+	justShowFlags subcommands.MutuallyExclusiveFlags
+)
 
 func init() {
 	cmd := &cobra.Command{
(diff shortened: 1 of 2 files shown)


subcommands/keys/ca_show.go
```diff
@@ -26,6 +29,9 @@ func init() {
 	}
 	caCmd.AddCommand(cmd)
 	cmd.Flags().BoolVarP(&prettyFormat, "pretty", "", false, "Display human readable output of each certificate")
+	justShowFlags.Add(cmd, "just-root", "Only show the Factory root CA certificate")
+	justShowFlags.Add(cmd, "just-tls", "Only show the device-gateway TLS certificate")
+	justShowFlags.Add(cmd, "just-cas", "Only show device authenticate certificates trusted by the device-gateway")
```
vkhoroz, line 34: maybe, `just-ca` or even better `just-device-ca`?

vkhoroz (approved):
LGTM

subcommands/keys/ca_show.go
doanac: i was on the fence. including `device` kind of make sense. It's technically plural, so it felt weird not doing that and then seeing multiple certs. Let me add `device` because that does make it more consistent.

doanac: updated in force-push

PR #184, opened 2022-09-05
Introduce the `fioctl keys tuf` command section and restructure TUF keys management commands

vkhoroz:


vkhoroz:
Please, see the https://github.com/foundriesio/rftm/tree/main/rfds/0011 for the rationale.

vkhoroz:
I still need to run more tests for this locally to cover all code paths.

For some commands it is a simple code move; but the `rotate-offline-key` command deserved more attention.

mike-sul:
Maybe subdirectory `tuf` instead of `keys/tuf_*`?

subcommands/keys/resign_root.go
```diff
@@ -16,11 +16,14 @@ var dryRun bool
...
-		Args:  cobra.ExactArgs(1),
-		Long:  "The root metadata, root.json expires. Re-signing bumps the expiriation date.",
+		Use:    "resign-root <offline key archive>",
+		Short:  "[DEPRECATED] Re-sign the Factory's TUF root metadata",
+		Hidden: true,
+		Run:    doResignRoot,
+		Args:   cobra.ExactArgs(1),
+		Long: `Please, use a more secure way to keep your root role fresh - rotate the root key.
+
+Ability to refresh expiration date using the resign-root will be removed in upcoming releases...`,
```
mike-sul, line 29: Why not remove it now?

subcommands/keys/cmd.go
```diff
@@ -27,8 +27,17 @@ to own the PKI infrastructure of their factory. This infrastructure is used
 to manage mutual TLS between your devices and the Foundries.io device gateway.`,
 }
 
+var tufCmd = &cobra.Command{
+	Use:   "tuf",
+	Short: "Manage The Update Framework Keys for your factory",
+	Long: `Every factory uses TUF to provision OTA updates to devices.
+This command allows customers to own TUF PKI of their factory in order to
```
mike-sul, line 34: `PKI` term is bit confusing in this context.

subcommands/keys/tuf_show_root.go
```diff
@@ -21,7 +21,16 @@ func init() {
 		Run:   doShowRoot,
 	}
 	show.Flags().BoolVarP(&showProd, "prod", "", false, "Show the production version")
-	cmd.AddCommand(show)
+	tufCmd.AddCommand(show)
+
+	legacyShow := &cobra.Command{
```
mike-sul, line 26: Maybe it's not worth warning about the command deprecation it at all.

subcommands/keys/resign_root.go
vkhoroz: There may be customer scripts which still use it; a good habit is to produce the deprecation warning for some reasonable period like e.g. 3-6 months.

subcommands/keys/cmd.go
vkhoroz: This command manages _TUF private keys_; I thought _TUF PKI_ is enough sinonymous with that. I can change it to the former if you like it more; or use another your suggestion.

subcommands/common.go
```diff
@@ -132,6 +132,30 @@ func Tabby(indent int, columns ...interface{}) *tabby.Tabby {
 	return tab
 }
 
+func AssertWritable(path string) {
+	st, err := os.Stat(path)
+	DieNotNil(err)
+	f, err := os.OpenFile(path, os.O_APPEND|os.O_WRONLY, st.Mode())
```
mike-sul, line 138: Interesting way to check whether a file is writable. I guess you checked `st.Mode()` approach.

subcommands/common.go
```diff
@@ -132,6 +132,30 @@ func Tabby(indent int, columns ...interface{}) *tabby.Tabby {
...
+	if err != nil {
+		fmt.Println("ERROR: File is not writeable:", path)
+		os.Exit(1)
+	}
+	f.Close()
+}
+
+func IsSliceSetEqual(first, second []string) bool {
```
mike-sul, line 146: Will it work if the slices have different number of the same value, e.g. []{'foo', 'foo'} and []{'foo'}? I guess we assume that elements are unique in the slices.

mike-sul: Also, it won't work if the first slice in longer, e.g. []{'foo', 'bar'} and []{'foo'}, the function considers them equal.

mike-sul: oh, this is just code reshuffling so I suppose it works for our use-cases.

subcommands/keys/tuf_crypto.go
```diff
@@ -27,12 +27,26 @@ type tufKeyTypeEd25519 struct{}
 
 const (
 	// These are case insensitive
+	tufRoleNameRoot = "Root"
```
mike-sul, line 30: I wonder if introducing of the `TUF role` type would be helpful and improve readability.

```
type TufRole string

const (
   Root TufRole = "Root"
   Targets = "Targets"
)
```

subcommands/keys/tuf_show_root.go
vkhoroz: It is worth - we're going to remove these legacy commands (already hidden now).

BTW. I reworked the deprecation logic to use the `cmd.Deprecated` attribute; this warrants that the deprecation warning will appear not only in the help message, but also every time the command is invoked by the user.

subcommands/common.go
vkhoroz: This is not mine invention, I just moved the @doanac 's code to a new place; and it has proven to work well for over a year.

subcommands/common.go
vkhoroz: I was just reshuffling; but based on your comments I have to look for bugs now.

subcommands/keys/tuf_crypto.go
vkhoroz: I would have to cast `TufRole <-> string` in a few places; but I'm lazy and don't want to do that.
The Golang typing models teached me to be lazy the hard way.

vkhoroz: IMHO `tufRoleNameRoot` is more verbose than just `Root`; I'm tired of one-word const names all over the codebase, which lead to subtle errors over time.

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -0,0 +1,380 @@
...
+		Short: "Rotate the offline TUF signing key for the Factory",
+		Long: `Rotate the TUF root or TUF targets offline signing key for the Factory.
+
+The new signing key is rotated in both CI and production TUF root transactionally.
+
+When you rotate the TUF targets offline signing key:
+- if there are any production targets in your factory, they are re-signed using the new key.
+- if there is an active wave in your factory, the TUF targets rotation is not allowed.`,
+		Run:         doRotateOfflineKey,
+		Annotations: map[string]string{tufCmdAnnotation: tufCmdRotateOfflineKey},
```
vkhoroz, line 39: @mike-sul I'm sorry for this change in the middle of your review; the use of annotations helped to simplify the Run attribute usage quite some.

vkhoroz:
> Maybe subdirectory `tuf` instead of `keys/tuf_*`?

I just followed our usual pattern for the subcommands directories.
There is, however, more than just that: if I make it a subdirectory - I would have to make the `cmd` variable of the `subcommands/keys/cmd.go` module public, resulting into more (unnecessary) changes.

subcommands/common.go
vkhoroz: Actually, it was a good catch; thanks @mike-sul.
https://github.com/foundriesio/fioctl/pull/184/commits/de529e85e568c9fb268cc3ba45dba6d1950ee5f7

subcommands/keys/cmd.go
doanac: I'd leave PKI out just because its started to be used exclusively for device-gateway PKI. You also might double check elsewhere to see if we do Factory or factory in usage statements, I don't remember

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -0,0 +1,380 @@
...
+			subcommands.DieNotNil(errors.New("First time TUF key rotation must be performed for the root role."))
+		}
+	case tufCmdRotateRootLegacy:
+		roleName = tufRoleNameRoot
+		credsFile = args[0]
+		firstTime, _ = cmd.Flags().GetBool("initial")
+	case tufCmdRotateTargetsLegacy:
+		roleName = tufRoleNameTargets
+		credsFile = args[0]
+		firstTime = false
```
doanac, line 180: seems like a `default: panic` might not be a bad idea here.

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -108,10 +113,17 @@ func doRotateOfflineKey(cmd *cobra.Command, args []string) {
 		if firstTime && roleName != tufRoleNameRoot {
 			subcommands.DieNotNil(errors.New("First time TUF key rotation must be performed for the root role."))
 		}
+		switch roleName {
```
doanac, line 166: all these switch statements are starting to give off a C code vibe where `#defines` is getting out of hand. i have a feeling there might be a cleaner way to do this so that the reader's eyes are so focused on these details and can instead focus on the broader rotation logic

subcommands/keys/resign_root.go
```diff
@@ -82,5 +82,5 @@ func doResignRoot(cmd *cobra.Command, args []string) {
 		os.Exit(1)
```
doanac, line 86: this is on the commit message. Can you explain what's unsafe about copy-targets?

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -97,22 +99,32 @@ Instead, please, use the "fioctl keys tuf rotate-offline-key --role=targets --cr
...
 	factory := viper.GetString("factory")
 	credsFile, _ := cmd.Flags().GetString("creds")
+	targetsCredsFile, _ := cmd.Flags().GetString("targets-creds")
 	root, err := api.TufRootGet(factory)
 	subcommands.DieNotNil(err)
 	creds, err := GetOfflineCreds(credsFile)
 	subcommands.DieNotNil(err)
-	subcommands.DieNotNil(syncProdRoot(factory, root, creds))
+	if targetsCredsFile == "" {
+		subcommands.DieNotNil(syncProdRoot(factory, root, creds, creds))
```
doanac, line 130: shouldn't this be `subcommands.DieNotNil(syncProdRoot(factory, root, creds, nil))`?

subcommands/keys/tuf_copy_targets.go
```diff
@@ -13,14 +13,20 @@ import (
...
+		Args:   cobra.ExactArgs(2),
 		Long: `This command extracts the target signing credentials required for initializing 
 waves into a new tarball so that the offline key archive isn't required for
 rolling out production updates. This should be run after each target key
 rotation and distributed to the operator in charge of production OTAs.`,
+		Deprecated: `it will be removed in the future.
+Please, use a more secure way to rotate the targets key into a separate file:
+	fioctl keys tuf rotate-offline-key --role=targets
+		--creds=<root-creds.tar.gz> --targets-creds=<targets-creds.tar.gz>
+`,
```
doanac, line 29: but what if i want to export the target key w/o having to do a rotation?

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -35,6 +35,20 @@ The new signing key is rotated in both CI and production TUF root transactionall
...
+		Example: `
+  # Take ownership of TUF root and targets keys for a new factory, keep them in separate files:
+  fioctl keys tuf rotate-offline-key --role=root \
+    --creds=tuf-root-keys.tar.gz --first-time
+  fioctl keys tuf rotate-offline-key --role=targets \
+    --creds=tuf-root-keys.tar.gz --targets-creds=tuf-targets-keys.tar.gz
+
+  # Rotate offline TUF root key and set the root role expiration to 2 years (the default is 1 year):
+  fioctl keys tuf rotate-offline-key --role=root \
+    --creds=tuf-root-keys.tar.gz --expires-days=730
```
doanac, line 47: I almost objected to the previous commit that allows you to customize these dates for this very reason. Let's just not even let people know they can do bad things.

subcommands/keys/tuf_utils.go
```diff
@@ -103,6 +104,54 @@ func SignMeta(metaBytes []byte, signers ...TufSigner) ([]tuf.Signature, error) {
 	return signatures, nil
 }
 
+func SignRoot(root *client.AtsTufRoot, signers ...TufSigner) error {
```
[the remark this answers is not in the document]
vkhoroz: The majority of functions in this module are already used outside the `subcommands/keys` namespace (e.g. SignMeta, GetOfflineCreds` etc. Although some are only used internally, when I looked at this module I noticed that all its function members _that can potentially be used outside_ are public. And I found that neat, especially when I look at the module history which predicts a 90% probability for every function to be used externally in the future.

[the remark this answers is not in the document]
vkhoroz: The majority of functions in this module are already used outside the `subcommands/keys` namespace (e.g. SignMeta, GetOfflineCreds` etc. Although some are only used internally, when I looked at this module I noticed that all its function members _that can potentially be used outside_ are public. And I found that neat, especially when I look at the module history which predicts a 90% probability for every function to be used externally in the future.

subcommands/keys/tuf_show_root.go
```diff
@@ -21,7 +21,19 @@ func init() {
...
+		Use:   "show-root",
+		Short: "Show the Factory's TUF root metadata",
+		Deprecated: `it has moved to a new place.
+Instead, please, use the "fioctl keys tuf show-root".
+`,
```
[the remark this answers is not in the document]
vkhoroz: An example help message:
```
Command "show-root" is deprecated, it has moved to a new place.
Instead, please, use the "fioctl keys tuf show-root".

Show the Factory's TUF root metadata

Usage:
  fioctl keys show-root [flags]
```

The `cobra` package generates the first part of the sentence (including the comma); so I had to spend some time to craft a grammatically valid continuation.

But it provide a nice feature that the same header is printed on every deprecated command invokation.

subcommands/keys/resign_root.go
```diff
@@ -12,7 +12,10 @@ import (
 	"github.com/spf13/viper"
```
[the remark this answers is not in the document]
vkhoroz: I will try to reword these in the next force-push.

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -0,0 +1,380 @@
...
+		Annotations: map[string]string{tufCmdAnnotation: tufCmdRotateOfflineKey},
+	}
+	rotate.Flags().StringP("role", "r", "", "TUF role name, supported: Root, Targets.")
+	_ = rotate.MarkFlagRequired("role")
+	rotate.Flags().StringP("creds", "o", "", "Offline creds file containing root keys.")
+	_ = rotate.MarkFlagRequired("creds")
+	_ = rotate.MarkFlagFilename("creds")
+	rotate.Flags().BoolP("first-time", "", false, "Used for the first customer rotation. The command will download the initial root key.")
+	rotate.Flags().StringP("key-type", "k", tufKeyTypeNameRSA, "Key type, supported: Ed25519, RSA (default).")
+	rotate.Flags().StringP("changelog", "m", "", "Reason for doing rotation. Saved in root metadata for tracking change history.")
```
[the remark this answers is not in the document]
vkhoroz: It wasn't required before; and actually the code generates a sensible default message; so I left it almost unmodified (only added the new key ID into the generated message to give it a feeling of some valuable information besides just generic blabla).

subcommands/common.go
mike-sul: It should work if elements in the slices are unique, otherwise `int`/counter should be used instead of boolean.

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -206,12 +223,29 @@ func doRotateOfflineKey(cmd *cobra.Command, args []string) {
 		// 2. Generate a new key.
 		// 3. Set these keys in root.json.
```
mike-sul, line 238: I guess it should be `add this key to` in the step #3 since the step #2 generates a key not keys?

subcommands/keys/tuf_rotate_offline_key.go
vkhoroz: It should be like present in the code.
The user might choose to keep both root and targets keys in one tarball; actually in the past it was the only way to do this.
Thus, we pass `creds, creds` so that the code have where to search for targets keys if needed (to re-sign prod targets).

subcommands/keys/tuf_rotate_offline_key.go
vkhoroz: There are two reasons why I made it configurable:
1. This piece of code/logic is a copy-paste from `wave init` where we already allow to customize the expiration (i.e. cross-feature parity).
2. The default expiration period for TUF root is quite controversial: notary suggests 10 years, uptane sets it to 1 year; while garage-sign has no defaults - the user must choose.

Thus far, can you state at least one reason why you think it is "bad thing"?

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -239,9 +255,7 @@ func swapRootKey(
 ) (*TufSigner, OfflineCreds) {
 	kp := GenKeyPair(keyType)
 	root.Signed.Keys[kp.signer.Id] = kp.atsPub
-	root.Signed.Expires = time.Now().AddDate(1, 0, 0).UTC().Round(time.Second) // 1 year validity
```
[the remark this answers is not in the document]
vkhoroz: The expires field is now set outside this function (in `doRotateOfflineKey`), and it makes sure to always set that field.

subcommands/keys/tuf_utils.go
```diff
@@ -47,6 +49,34 @@ func ParseTufRoleNameOffline(s string) string {
 	return r
 }
 
+func ReadExpiration(cmd *cobra.Command, defaultDays int) (expires time.Time) {
+	var err error
+	now := time.Now().UTC()
+	if cmd.Flags().Changed("expires-at") {
+		at, _ := cmd.Flags().GetString("expires-at")
+		expires, err = time.Parse(time.RFC3339, at)
+		subcommands.DieNotNil(err, "Invalid expires-at value:")
```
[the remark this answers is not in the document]
vkhoroz: The `err` contains the exact detailed error message.

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -206,12 +223,29 @@ func doRotateOfflineKey(cmd *cobra.Command, args []string) {
...
+		}
+
 		onlineTargetsId, err := findOnlineTargetsId(factory, *root)
 		subcommands.DieNotNil(err)
-		newKey, newCreds = replaceOfflineTargetsKey(root, onlineTargetsId, creds, keyType)
+		newKey, newCreds = replaceOfflineTargetsKey(root, onlineTargetsId, targetsCreds, keyType)
 		fmt.Println("= New target keyid:", newKey.Id)
 		fmt.Println("= Resigning prod targets")
-		subcommands.DieNotNil(resignProdTargets(factory, root, onlineTargetsId, creds))
+		subcommands.DieNotNil(resignProdTargets(factory, root, onlineTargetsId, newCreds))
```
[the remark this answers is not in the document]
vkhoroz: The exact order of actions is this:
1. `tufRootPost` for the CI TUF root;
2. if 1 succeeds - `resignProdTargets ` on the client side;
3. if 2 succeeds - `tufRootPost` for the prod TUF root together with prod targets payload, saved atomically;

Hence, there is no time period when prod targets signatures are not valid.

subcommands/common.go
vkhoroz: This is not `IsSliceEqual`, it is `IsSliceSetEqual` i.e. it compares sets, not samples; and in a set logic `set("foo, foo") == set("foo")`.

That is exactly what we want from a function named that way.

subcommands/keys/cmd.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/184/commits/bc1eeb2de75e2bbf2ee84951fbcae0f109d01cc4 (squash later)

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -0,0 +1,380 @@
...
+	keyType := ParseTufKeyType(keyTypeStr)
+	changeLog, _ := cmd.Flags().GetString("changelog")
+	switch cmd.Annotations[tufCmdAnnotation] {
+	case tufCmdRotateOfflineKey:
+		roleName, _ = cmd.Flags().GetString("role")
+		roleName = ParseTufRoleNameOffline(roleName)
+		credsFile, _ = cmd.Flags().GetString("creds")
+		firstTime, _ = cmd.Flags().GetBool("first-time")
+		if firstTime && roleName != tufRoleNameRoot {
+			subcommands.DieNotNil(errors.New("First time TUF key rotation must be performed for the root role."))
```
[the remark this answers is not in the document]
vkhoroz: https://github.com/foundriesio/fioctl/pull/184/commits/bc1eeb2de75e2bbf2ee84951fbcae0f109d01cc4 (squash later)

subcommands/keys/tuf_rotate_offline_key.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/184/commits/f9cf0788a7ebebc98982ec462e35d3d500931b00 (squash later)

subcommands/keys/tuf_rotate_offline_key.go
mike-sul: Ah, OK, when reviewing the commit change the overall function is not visible.

subcommands/keys/tuf_rotate_offline_key.go
vkhoroz: Used the default expiration map: https://github.com/foundriesio/fioctl/pull/184/commits/bc1eeb2de75e2bbf2ee84951fbcae0f109d01cc4 (squash later).

The other switch statements seem sensible to me and really helped to minimize the amount of code.

I know someone would say we need a proper class heirarchy with clearly defined interfaces to avoid the switch and goto statements, because they are religiously impure. But IMHO where a simple switch makes the job we don't need anything else.

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -0,0 +1,380 @@
...
+func init() {
+	rotate := &cobra.Command{
+		Use:   "rotate-offline-key --role root|targets --creds=<offline-creds.tgz>",
+		Short: "Rotate the offline TUF signing key for the Factory",
+		Long: `Rotate the TUF root or TUF targets offline signing key for the Factory.
+
+The new signing key is rotated in both CI and production TUF root transactionally.
+
+When you rotate the TUF targets offline signing key:
+- if there are any production targets in your factory, they are re-signed using the new key.
```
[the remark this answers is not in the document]
vkhoroz: https://github.com/foundriesio/fioctl/pull/184/commits/c87616db4b3d6f3a54fbbedba2a07e0931f9d2a5 (squash later)

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -108,10 +113,17 @@ func doRotateOfflineKey(cmd *cobra.Command, args []string) {
 		if firstTime && roleName != tufRoleNameRoot {
 			subcommands.DieNotNil(errors.New("First time TUF key rotation must be performed for the root role."))
 		}
+		switch roleName {
+		case tufRoleNameRoot:
+			expires = ReadExpiration(cmd, 365)
+		case tufRoleNameTargets:
+			expires = ReadExpiration(cmd, 0)
```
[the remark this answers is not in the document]
vkhoroz: It means "leave the expiration as is from the previous root role".
This is how the targets key rotation behaves now, and it is a sensible default.

I added a one-line comment in https://github.com/foundriesio/fioctl/pull/184/commits/bc1eeb2de75e2bbf2ee84951fbcae0f109d01cc4 for this.

mike-sul (approved):

subcommands/keys/tuf_copy_targets.go
vkhoroz: We should be very careful about all the _what if customer wants_ statements, especially in security related topics like this.

IMHO the most secure way to work with TUF keys is to keep the root and targets private keys separately. This becomes especially relevant when we start looking at the HSM support. With the HSM enabled the factory _root_ admin might even have no access to the _targets_ key, if that responsibility is delegated to a different person. That person would generate the _targets_ key on their HSM and let the _root_ admin sign a new _root_ with a new targets public part.

As you see, above there is no any `creds.tar.gz` at all. But even without the HSM that workflow can be generated if we keep the root and targets keys in two separate files.

Now back to you question. We will always have a group of customers want to take it easy and don't bother with security. For them, it might look like a quick start to keep root and targets keys in one tarball, so I keep that functionality. But, there should be no _today I want this, tomorrow I want that_ with regards to security. Even if the customer wants to switch from keeping-it-all-in-one-file to a more secure way to keep the targets key separately - they should be forced to generate a new targets key. I would allow nothing like _keep it here and keep it there_; redundancy is better achieved by different means.

vkhoroz: @doanac the above paragraph looks to contain _strong opinions_, however, regarding the `copy-targets` I don't really have _that strong opinion_. So, if you have any reasonable argument why we _do need_ it - I'm fine, and will simply move it under the `fioctl keys tuf` section.

When I think about e.g. the HSM support, the customer will need to somehow migrate existing TUF keys into the HSM. Depending onto the HSM model, there are two options:
- if the HSM supports inserting the crypto key into it - the `copy-targets` might come handy.
- if it doesn't - we'll have to generate a new key anyway.

IMHO, when we start talking about HSMs, the private key should never cross its boundary (i.e. no in, no out); at least that's what was the requirement back when I was working on FIPS-140-2 compliance. Thus, (again IMHO) we should generate the new keys when migrating the customers to HSM anyway. Other than that, I simply see no use for `copy-targets`. So it's like _less commands means less maintenance_ thing for me.

vkhoroz:
@doanac this was now fully tested; I will squash all fixups and shorten commit messages once you have your review done.

subcommands/keys/tuf_utils.go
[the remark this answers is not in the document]
doanac: if you passionate about it - go for it. i'd generally recommend we don't make it public until there's a real need. But its not a big deal.

subcommands/keys/tuf_show_root.go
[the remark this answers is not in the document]
doanac: ah. thanks!

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -0,0 +1,380 @@
...
+		Short:   "Rotate the offline target signing key for the Factory",
+		Deprecated: `it has moved to a new place.
+Instead, please, use the "fioctl keys tuf rotate-offline-key --role=root --creds=<tuf-creds.tar.gz>".
+`,
+		Hidden:      true,
+		Run:         doRotateOfflineKey,
+		Annotations: map[string]string{tufCmdAnnotation: tufCmdRotateRootLegacy},
+		Args:        cobra.ExactArgs(1),
+	}
+	legacyRotateRoot.Flags().BoolP("sync-prod", "", false, "Make sure production root.json is up-to-date and exit")
```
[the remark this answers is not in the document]
doanac: yep. i caught up to that at the of the PR. commits were just not ordered that way. don't bother re-ordering, i'm guessing that will cause conflicts

subcommands/keys/tuf_rotate_offline_key.go
doanac: 10 years is bad. Ricardo has posted links where someone like NIST has decent guidance. 1 year is a good value. Giving options just takes something a user is scared about and gives them something else to fear and worry about. Then they start thinking "should i use expires-days or expires-at? i wonder why there are two options, do they do the same thing?". Then 30 minutes later they exit the rabbit hole.

subcommands/keys/tuf_rotate_offline_key.go
```diff
@@ -0,0 +1,467 @@
...
+	cmdName := cmd.Annotations[tufCmdAnnotation]
+	switch cmdName {
+	case tufCmdRotateOfflineKey:
+		roleName, _ = cmd.Flags().GetString("role")
+		roleName = ParseTufRoleNameOffline(roleName)
+		credsFile, _ = cmd.Flags().GetString("creds")
+		targetsCredsFile, _ = cmd.Flags().GetString("targets-creds")
+		firstTime, _ = cmd.Flags().GetBool("first-time")
+		if firstTime && roleName != tufRoleNameRoot {
+			subcommands.DieNotNil(errors.New("The --first-time is only valid for the first TUF root key rotation."))
```
doanac, line 162:
```suggestion
			subcommands.DieNotNil(errors.New("The --first-time option is only valid for the first TUF root key rotation."))
```

subcommands/keys/tuf_rotate_offline_key.go
doanac: i don't want you to go crazy on abstractions. you have a feel for the code. If you don't think this will be error prone making future changes where you hit things like "damn i updated the special switch cases for the first 2 swtiches but forgot the minor change for the third".

subcommands/keys/cmd.go
```diff
@@ -27,8 +27,17 @@ to own the PKI infrastructure of their factory. This infrastructure is used
 to manage mutual TLS between your devices and the Foundries.io device gateway.`,
 }
 
+var tufCmd = &cobra.Command{
+	Use:   "tuf",
+	Short: "Manage The Update Framework Keys for your factory",
+	Long: `Every factory uses TUF to provision OTA updates to devices.
+This command allows customers to own TUF private keys of their factory in order to
+warrant that only customer's personnel can crypto-sign device software updates.`,
```
doanac, line 35: It might be easier to just drop the `Long` attribute so we don't get into a grammatical hole. But perhaps we could state it this way:
```
These sub-commands allow you to manage your Factory's TUF private keys and ensure you are in complete control of your OTA metadata.
```

Actually - i don't think there's a good short answer. Each subcommand's message does a decent job explaining what/why. So leaving this out might be best.

subcommands/keys/tuf_copy_targets.go
doanac: maybe just make the statement less opinionated then. Rather than "Please, use a more secure way to rotate" just say something like "Please do it this way".

subcommands/keys/tuf_utils.go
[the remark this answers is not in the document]
vkhoroz: @doanac in general I agree with what you say. Here I'm just trying to find a balance between the _convenience of linter seeing unused functions_ and the _convenience of not having to switch between private/public every so often_. Both perks are driven more by laziness than necessity; so the latter laziness just won for me in this specific case.

subcommands/keys/tuf_rotate_offline_key.go
vkhoroz: I agree that 1 year is a sensible value; that's why it is the default. My guts feeling is that customers will ask for this rather sooner than later: some will want it more aggressive like 3-6 years; some will be more happy with 2-3 years. In general I tend to agree on minimalistic command designs (preferably with less than 5 options), but it is a hard game about clarity versus convenience. E.g. I miss some options in `pip install` despite it having 30 options I don't use.

That said, if you insist I will cut this commit; it is fairly easy to add it at any time later.

doanac: i think we should go for the apple approach here and just say "no, you can't do over 1 year because its not a good idea". and - if someone absolutely insists. there are options:
* the api allows it already
* they can fork fioctl
* we can add this in
Either way - making it painful to do something against recommendations isn't the worst thing

vkhoroz:
@doanac I have one specific question on this:
If you noticed, I've made all the deprecated commands hidden, as I see it is what we usually do in PRs like this.
However, I'm not sure this is a good idea, as we also need to update the documentation referencing these commands.

So, there are 3 options:
1. Deprecate, but don't hide for now, later update documentation, and then hide.
2. Deprecate and hide now, but don't cut a new fioctl release until the documentation change is ready, then release both in sync.
3. Deprecate and hide now, don't bother about documentation, it is OK if documentation lags by some reasonable time.

In any case, these commands are not removed, a user will be able to use them, they will just produce a warning and stay hidden.

doanac:
@vkhoroz - how about an option 2.5: Deprecate and hide now. Update the docs. Make sure we release fioctl before the next update (this almost always happens and will certainly do so because of this functionality that's landing).

subcommands/keys/resign_root.go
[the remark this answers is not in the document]
vkhoroz: fixed in force-push

subcommands/keys/resign_root.go
vkhoroz: changed in force-push to be less opinionated

subcommands/keys/tuf_copy_targets.go
vkhoroz: changed in force-push to be less opinionated

subcommands/keys/tuf_rotate_offline_key.go
vkhoroz: Remove this commit from the final force-push

subcommands/keys/tuf_rotate_offline_key.go
vkhoroz: fixed in the latest force-push

subcommands/keys/cmd.go
vkhoroz: Actually, I like your suggested help message (only changed two words).
We need _something_ here; otherwise `keys tuf` looks poor compared to `keys ca`.

fixed in the latest force-push

doanac (approved):

vkhoroz:
@doanac this should be ready for the final review

I re-read all help messages and updated them to have the same look-and-feel (primarily the examples). And also modified the option names (`--creds -> --keys`) to be conformant to the `waves init` option names.

doanac:
> I re-read all help messages and updated them to have the same look-and-feel (primarily the examples). And also modified the option names (--creds -> --keys) to be conformant to the waves init option names.

yep - i noticed that. looks good to me now. merge when you are comfortable. You might do this as a merge commit, will probably make it eaiser for me to comb through changes when doing release notes.

PR #185, opened 2022-09-09
Add a new hidden "post" command

doanac:
Much like `fioctl get`, this command can be handy for people trying to interact/experiment directly with our API.

### DIFF
--- a/subcommands/post.go
+++ b/subcommands/post.go
@@ -1,0 +1,52 @@
+package subcommands
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+func NewPostCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "post https://api.foundries.io/ota.... [header=val..]",
+		Short:  "Do an authenticated HTTP POST",
+		Hidden: true,
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			api = Login(cmd)
+		},
+		Run:  doPost,
+		Args: cobra.MinimumNArgs(1),
+	}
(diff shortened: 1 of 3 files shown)


subcommands/post.go
```diff
@@ -0,0 +1,52 @@
...
+		Short:  "Do an authenticated HTTP POST",
+		Hidden: true,
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			api = Login(cmd)
+		},
+		Run:  doPost,
+		Args: cobra.MinimumNArgs(1),
+	}
+	cmd.PersistentFlags().StringP("token", "t", "", "API token from https://app.foundries.io/settings/tokens/")
+	cmd.Flags().StringP("data", "d", "", "HTTP POST data")
```
vkhoroz, line 24: I would love if this also supports one or both of the two options:
- an input piping so I can conveniently do magic like `fioctl get smth | jq .some.data | fioctl post smth-else`;
- a file argument, so I can do smth like `fioctl post smth --file data.json`.

I know those can be implemented via the `--data` attribute, but one can easily mess with string escaping when doing tricks like `fioctl post smth --data="$(fioctl get smth | jq .some.data)"` or `fioctl post smth --data="$(cat data.json)"`.

vkhoroz (approved):
This looks scary, but I was looking for a command like this several times.

Thanks!

subcommands/post.go
doanac: good idea. I sort of did it like curl does: https://github.com/foundriesio/fioctl/pull/185/commits/9621994ba4ff4de797d1a06a3610e6110e951a61

vkhoroz:
LGTM

PR #186, opened 2022-09-13
Introduce `targets show sbom` command

doanac:


subcommands/targets/show.go
```diff
@@ -42,6 +44,31 @@ func init() {
...
+  fioctl targets show sboms 42 --download /tmp
+
+  # Show an overview of an SBOM from version 42:
+  # File with .spdx.tar.zst are produced by Yocto. Yocto produces an SPDX file
+  # per package. This will go through all the files and show the packages.
+  fioctl targets show sboms 42 42/raspberrypi4-64 initramfs-ostree-lmp-image-raspberrypi4-64.spdx.tar.zst
+  
+  # Show an overview of a container's SBOM from version 42:
+  # Files like this are from container builds where Syft was used
+  fioctl targets show sboms 42 41/build-armhf alpine:latest/arm.spdx.json`,
```
quaresmajose, line 67: This `41/build-armhf` seams to be a typo and on 41 it should be 42

doanac: no. 42 in this case was Platform build that inherited its containers from build 41

quaresmajose: right, thanks and sorry for the noise

subcommands/targets/show.go
```diff
@@ -51,6 +52,7 @@ func init() {
 		Args:  cobra.ExactArgs(1),
 	}
 	showCmd.AddCommand(sbomCmd)
+	sbomCmd.Flags().String("download", "", "Download all sboms to given path")
```
vkhoroz, line 70: Should we state somewhere explicitly that it downloads to a given _directory_ (not just any path).

subcommands/targets/show.go
```diff
@@ -238,17 +240,32 @@ func doShowSboms(cmd *cobra.Command, args []string) {
...
 		sboms, err := api.TargetSboms(factory, name)
 		subcommands.DieNotNil(err)
-		t := tabby.New()
-		t.AddHeader("BUILD", "ARTIFACT")
-		for _, sbom := range sboms {
-			t.AddLine(sbom.CiBuild+"/"+sbom.CiRun, sbom.Artifact)
+
+		if len(downloadPath) > 0 {
+			st, err := os.Stat(downloadPath)
+			subcommands.DieNotNil(err)
```
vkhoroz, line 294: Below we assume that it is a directory; so I would check for that here explicitly.
Also, it might make sense to check that directory is empty (so that we don't accidentally overwrite its contents).

client/foundries_sboms.go
```diff
@@ -44,3 +52,32 @@ func (s Sbom) Download(dst string) error {
...
+	return doc.Packages, nil
+}
+
+func (p SpdxPackage) License() string {
+	if p.LicenseConcluded == p.LicenseDeclared {
+		return p.LicenseConcluded
+	} else if p.LicenseConcluded == "NOASSERTION" {
+		return p.LicenseDeclared
+	}
+	panic(fmt.Errorf("Unknown license configuration for package: %s - Concluded(%s) Declared(%s)", p.Name, p.LicenseConcluded, p.LicenseDeclared))
```
vkhoroz, line 91: I'm not sure we need to panic here.
This means that in case of one broken item in SBOM customer cannot view anything.
I'd rather return here something sensible (e.g. a given error message).

subcommands/targets/show.go
```diff
@@ -243,6 +258,25 @@ func doShowSboms(cmd *cobra.Command, args []string) {
...
+			t := tabby.New()
+			t.AddHeader("NAME", "VERSION", "LICENSE")
+			for _, pkg := range packages {
+				t.AddLine(pkg.Name, pkg.VersionInfo, pkg.License())
+			}
+			t.Print()
+			// all targets come from the same CI build with the same SBOM return
+			return
+		}
```
vkhoroz, line 282: Does it make sense for a user to use the `--download` option in this code path?

- If yes - please, implement it.
- If no - please, show an error message if it was provided.

client/foundries_sboms.go
```diff
@@ -39,18 +40,26 @@ func (a *Api) TargetSboms(factory string, targetName string) ([]Sbom, error) {
...
+		buf, err := s.api.TargetSbomPackagesCsv(s.factory, targetName, path)
+		if err != nil {
+			return err
+		}
+		return os.WriteFile(dst, buf, 0o744)
+	}
 	logrus.Debugf("Downloading %s", s.Uri)
-
-	buff, err := s.api.Get(s.Uri)
+	buf, err := s.api.Get(s.Uri)
```
vkhoroz, line 58: This var correction belogs to a different commit.

client/foundries_sboms.go
```diff
@@ -39,18 +40,26 @@ func (a *Api) TargetSboms(factory string, targetName string) ([]Sbom, error) {
...
-func (s Sbom) Download(dst string) error {
+func (s Sbom) Download(targetName, dst string, asCsv bool) error {
+	if asCsv {
+		path := fmt.Sprintf("%s/%s/%s", s.CiBuild, s.CiRun, s.Artifact)
+		buf, err := s.api.TargetSbomPackagesCsv(s.factory, targetName, path)
+		if err != nil {
+			return err
+		}
+		return os.WriteFile(dst, buf, 0o744)
+	}
```
vkhoroz, line 56: IMHO this thing becomes hard to understand.

For example, I don't understand why we download from `s.Uri` in one case, and craft a different Uri in another one; I had to look into the API implementation to understand that.

So, two comments here:

1. It would be nice to have a comment why we download from Uri in one case, and do Api call in another case.
2. I think a cleaner solution would be to use the `api.Get(sbom.URI), api.TargetSbomPackagesCsv(factory, targetName, sbom.ApiPath())` directly inside the `--download` option handler; that would keep complexity in just one place.

subcommands/post.go
```diff
@@ -55,7 +54,7 @@ func doPost(cmd *cobra.Command, args []string) {
 	data, _ := cmd.Flags().GetString("data")
 	if data == "-" {
 		logrus.Debug("Reading post data from stdin")
-		dataBytes, err = ioutil.ReadAll(os.Stdin)
+		dataBytes, err = io.ReadAll(os.Stdin)
```
vkhoroz, line 57: Thanks

doanac:
I really didn't like my first attempt. This is a 2nd attempt that's pretty much a redo.

doanac:
ci-retest

subcommands/targets/show.go
```diff
@@ -45,19 +54,51 @@ func init() {
...
+		contentType: "application/spdx.json",
+	})
+	sbomFormats = append(sbomFormats, sbomFormat{
+		name:        "cyclonedx",
+		contentType: "application/cyclone.json",
+	})
+	sbomFormats = append(sbomFormats, sbomFormat{
+		name:        "csv",
+		contentType: "text/csv",
+	})
```
vkhoroz, line 104: Can you make this a map?
I think it could even be a literal map like below:

```
var sbomFormats = map[string]string{
    "spdx": "application/spdx.json",
    ...
}

func parseSbomFormat(format string) (res sbomFormat) {
    if ct, ok := sbomFormats[format]; ok {
        res = sbomFormat{format, ct}
    } else {
        DieNotNil(fmt.Errorf("Unknown --format: %s", format))
    }
    return
}
```

That should save quite a few lines of code and make it more clear.

vkhoroz:
LGTM

subcommands/targets/show.go
doanac: Thanks. I'm finding a struct a little easier still because there's special handling for the "table" output. However, by setting a default CLI option and making this a helper function, I've got the code a lot easier:

 https://github.com/foundriesio/fioctl/pull/186/commits/70d47e90517ee5996b47a21df0e4d4de0a1d5f1d

doanac:
force push with fix commit squashed

subcommands/targets/show.go
```diff
@@ -45,19 +54,51 @@ func init() {
...
+
+	allowed := ""
+	for i, fmt := range sbomFormats {
+		if i > 0 {
```
vkhoroz, line 108: I'd recommend to check for `f.name != "table"` here explicitly.

doanac: how about even simpler and not use any kind of "needle"?
```
        allowed := ""
	for _, fmt := range sbomFormats {
		if len(allowed) > 0 {
			allowed += ","
		}
		allowed += fmt.name
	}
```
doesn't see like we should make comma joining logic be dependent on the order of the inputs?

subcommands/targets/show.go
vkhoroz: It is still possible to use the struct with a map just like I proposed in my example (see `res = sbomFormat{format, ct}`) or in a different way.
The whole idea was to use a map instead of a list as it is easier to work with in this case (doesn't require an append or a for loop to search).

I'm fine if you like the current approach more.

doanac: its just gets weird because the map doesn't give you control over the order you display the format options. It's not huge, but I was trying to present them in a certain order.

vkhoroz: That's reasonable.

Now, what about just hard-coding a list of options for the help message?
It's just 4 simple comma-separated words; do we need all this hussle for the CLI help message?

vkhoroz: Note, that I usually prefer a natural language in help messages like `allowed formats: foo, bar, or baz`.
Now go write a function which does that (I know it's possible, but we'd need to support it, and add localization at 2025).

doanac: i've rewritten as an array and hard-coded the logic:

ebc402aa36a05fdd81f602410e1ee2671f6a1bb3

vkhoroz: Looks much simpler now, thanks.

You've got my LGTM earlier anyway, so ship whatever way you like more.

PR #188, opened 2022-10-11
fix(post): Post subcommand index error

StealthyCoder:
There was an indexing error because we check for '@' first but if you give an empty body `-d ""` it will fail with an index[0] lookup error.


doanac (approved):
good catch. feel free to merge.

PR #190, opened 2022-10-28
Revert the TUF keyid field calculation to support garage-sign

vkhoroz:
We still use the garage-sign to sign targets; some of our customers also use it. So, we need to support the garage-sign required algorithm to generate the TUF key IDs. Which by the way differs a lot from the TUF specification.

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -2,7 +2,6 @@ package client
 
 import (
 	"bytes"
-	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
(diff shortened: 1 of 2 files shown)


vkhoroz:
@mike-sul I'm sorry for being nasty about this, but... can you check that this works as expected with garage-sign?

I wrote this change as a side-project from my primary work (that was easy), but verifying this requires much more time.

doanac (approved):
Looks good to me assuming it interoperates with garage-sign

mike-sul:
> can you check that this works as expected with garage-sign?

sure, a bit later today.

mike-sul:
@vkhoroz It works for me.

PR #194, opened 2022-11-21
Git/support oauth

StealthyCoder:
Added git credential helper. 

It contains the same logic as the docker one.

### DIFF
--- a/subcommands/git/cmd.go
+++ b/subcommands/git/cmd.go
@@ -1,0 +1,114 @@
+package git
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"os/exec"
+	"os/user"
+	"path/filepath"
+
+	"github.com/foundriesio/fioctl/subcommands"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const GIT_CREDS_HELPER = "git-credential-fio"
+
+var (
+	helperPath string
+)
+
+func NewCommand() *cobra.Command {
+	path, err := exec.LookPath("git")
+	if err == nil {
+		helperPath = filepath.Dir(path)
+	}
+
+	cmd := &cobra.Command{
+		Use:   "configure-git",
+		Short: "Configure a source.foundries.io Git credential helper",
+		Long: `Configure a Git credential helper that allows Git to access
+source.foundries.io.
+
+This command will likely need to be run as root. It creates a symlink,
+git-credential-fio, in the same directory as the git client binary.
+
+NOTE: The credentials will need the "source:read-update" scope to work with Git`,
+		Run: doGitCreds,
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -39,6 +39,7 @@ require (
 	github.com/mattn/go-colorable v0.1.12 // indirect
 	github.com/mattn/go-isatty v0.0.14 // indirect
 	github.com/mitchellh/mapstructure v1.4.3 // indirect
+	github.com/muja/goconfig v0.0.0-20180417074348-0a635507dddc
```
doanac, line 42: this should go up in the first requires stanza as its it a direct dependency. However, I wonder:

1) Will github.com/pelletier/go-toml work? we already depend on that, so one less dep to vet.
2) Should we do this directly or use `git config` commands. Its a pretty advanced command so I worry we might mess something up. It also might let you drop some code that's trying to discover where gitconfig is located

StealthyCoder: Removed the dep entirely and just used two commands.

subcommands/git/cmd.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	return filepath.Clean(self)
+}
+
+func doGitCreds(cmd *cobra.Command, args []string) {
+	self := findSelf()
+
+	c := exec.Command("git", "config", "--global", "credential.https://source.foundries.io.username", "fio-oauth2")
+	subcommands.DieNotNil(c.Run())
+	c = exec.Command("git", "config", "--global", "credential.https://source.foundries.io.helper", "fio")
```
doanac, line 64: might want to have stderr,stdout go to a pipe(the same pipe is probably fine), so that you can include that in an error message.

subcommands/git/cmd.go
```diff
@@ -0,0 +1,90 @@
...
+	subcommands.DieNotNil(c.Run())
+
+	dst := filepath.Join(helperPath, GIT_CREDS_HELPER)
+	fmt.Println("Symlinking", self, "to", dst)
+	subcommands.DieNotNil(os.Symlink(self, dst))
```
doanac, line 69: this would need to run as root, right? so this means you'd be configuring the helper for the root user and not the user that ran `sudo ....`?

subcommands/git/cmd.go
StealthyCoder: Gotcha. will hook those up.

doanac (approved):
tested on my machine and it works

doanac:
I just merged my  cert rotation work which doesn't have docs with the following rationale - i'm going to flag "unreleased/undocumented" features as hidden commands with Cobra. Then we can finish up our work and also start to loop in folks like QA. I think we can do the same with this if you are ready to merge.

PR #196, opened 2022-11-21
Refactoring to prepare for the TUF updates feature

vkhoroz:
* extract TUF root client routines into a separate file;
* rename TUF related routines to hold Tuf prefix or suffix where applicable;
* make private all TUF related functions (which are not re-used outside the `subcommands/keys` package;
* use canonical MarshalIndent to output user-friendly JSON data in all commands;
* some other minor non-functional code cleanups.

### DIFF
--- a/subcommands/common.go
+++ b/subcommands/common.go
@@ -1,6 +1,7 @@
 package subcommands
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"os"
@@ -8,6 +9,7 @@ import (
 	"text/tabwriter"
 
 	"github.com/cheynewallace/tabby"
+	canonical "github.com/docker/go/canonical/json"
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/shurcooL/go/indentwriter"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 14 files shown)


mike-sul (approved):

vkhoroz:
CC: @doanac 

This part was tested locally.

vkhoroz:
The last force push adds a cosmetic change to print key id instead of pubkey when the key is not found.
That makes looking for the problem easier as key ID is a part of the key filename.

subcommands/common.go
```diff
@@ -132,6 +134,20 @@ func Tabby(indent int, columns ...interface{}) *tabby.Tabby {
 	return tab
 }
 
+// Copied from canonical.MarshalIndent, but replaced the Marshal call with MarshalCanonical.
+func MarshalIndent(v interface{}, prefix, indent string) ([]byte, error) {
```
doanac, line 138: There is a bug in canonical.MarshalIndent. I forget the exact issue, but I think it happens if you ever call it more than once. I *think* I hit it while playing on this commit:

 https://github.com/doanac/fioctl/commit/2b0256c65d1d49710fa8ab4dd45e63b205be04a1

and thus - still had to use `json.Marshal` to get the correct content

mike-sul (approved):

subcommands/common.go
vkhoroz: I suspect that due to the _premature optimizations_ here https://github.com/docker/go/blob/master/canonical/json/encode.go#L264 and here https://github.com/docker/go/blob/master/canonical/json/encode.go#L332

On practice, after this commit there is only one place where we call `canonical.Marshal` twice: `fioctl keys resign-root`.
I tested it and it produces the correct result.

IIRC the problem appeared when using a mix of `canonical.MarshalCanonical` and `canonical.MarshalIndent` inter-changeably. But I rewrote our version of `MarshalIndent` in such a way that it is always _canonical_. So, as long as we never use `canonical/json` to produce non-canonical output - we should be fine.

doanac (approved):

PR #197, opened 2022-11-25
Improve: return well formatted errors where API allows

vkhoroz:
Most of our APIs return the well-formatted JSON error for 4xx status codes. Print these errors in a user-friendly output.

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -542,10 +542,58 @@ func readResponse(res *http.Response, log logrus.FieldLogger) (*[]byte, error) {
 		// Still return a body, a caller might need it, but also return an error
 		msg := fmt.Sprintf("HTTP error during %s '%s': %s",
 			res.Request.Method, res.Request.URL.String(), res.Status)
-		if len(body) < PRINT_LIMIT {
-			// return an error response body up to a meaningful limit - if it spans beyond a few
-			// lines, need to find a more appropriate message.
-			msg = fmt.Sprintf("%s\n=%s", msg, body)
+
+		// Some APIs return well-formatted errors, try to use them
+		var (
+			useGenericError bool
+			listErrors      struct {
+				Msg     string   `json:"msg,omitempty"`
+				Message string   `json:"message,omitempty"`
+				Errors  []string `json:"errors,omitempty"`
+			}
+			dictErrors struct {
+				Msg     string            `json:"msg,omitempty"`
+				Message string            `json:"message,omitempty"`
+				Errors  map[string]string `json:"errors,omitempty"`
+			}
+		)
+
+		if merr := json.Unmarshal(body, &listErrors); merr == nil {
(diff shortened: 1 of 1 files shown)


doanac:
this is a welcome change I've wanted to introduce forever. you mind sharing a couple examples of the new output in the commit message?

client/foundries.go
```diff
@@ -542,10 +542,58 @@ func readResponse(res *http.Response, log logrus.FieldLogger) (*[]byte, error) {
...
+			}
+			dictErrors struct {
+				Msg     string            `json:"msg,omitempty"`
+				Message string            `json:"message,omitempty"`
+				Errors  map[string]string `json:"errors,omitempty"`
+			}
+		)
+
+		if merr := json.Unmarshal(body, &listErrors); merr == nil {
+			if listErrors.Msg != "" {
```
mike-sul, line 562: Just really not important comment. The first part for these two types of Errors is the same so maybe the code can be unified.

mike-sul (approved):

client/foundries.go
mike-sul: My first thought was about doing it in a functional way :).

client/foundries.go
```diff
@@ -542,10 +542,58 @@ func readResponse(res *http.Response, log logrus.FieldLogger) (*[]byte, error) {
...
+				msg += "\n=" + dictErrors.Msg
+			} else if dictErrors.Message != "" {
+				msg += "\n=" + dictErrors.Message
+			} else {
+				useGenericError = true
+			}
+			if !useGenericError && dictErrors.Errors != nil {
+				for field, emsg := range dictErrors.Errors {
+					msg += fmt.Sprintf("\n * %s: %s", field, emsg)
+				}
```
doanac, line 585: It's kind of nitpicking, but the `=<message` felt a little weird to me. e.g.:
```
ERROR: HTTP error during PATCH 'https://api.foundries.io/ota/devices/andy-quitt/config/?factory=andy-corp': 400 BAD REQUEST
=Given tag 'foooo' is not present in targets.json
```

I wonder if we should use the asterisk approach for all of this and make the message be:
```
ERROR: HTTP error during PATCH 'https://api.foundries.io/ota/devices/andy-quitt/config/?factory=andy-corp': 400 BAD REQUEST
* Given tag 'foooo' is not present in targets.json
```

Code seems fine though.

vkhoroz: We use `= ` in fioctl in other locations like this; so I preserve the same style here.
The ` *` is actually a _new way_ which is only used here, and only to represent a list of items... message is printed the same as elsewhere.

That said, there is a small problem - I need to ad the _space after =_ (it was missing in the initial commit too, but is present in other locations).

vkhoroz: changed in force-push

PR #201, opened 2022-12-05
Do not require the txid for the user who initiated TUF updates

vkhoroz:


vkhoroz:
tested with the latest API changes

doanac (approved):

PR #203, opened 2022-12-08
feat(wave): prune multiple targets

StealthyCoder:
This should support pruning old targets as part of starting a new wave.

### DIFF
--- a/subcommands/waves/init.go
+++ b/subcommands/waves/init.go
@@ -3,12 +3,14 @@ package waves
 import (
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/docker/go/canonical/json"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"golang.org/x/exp/slices"
 
 	"github.com/foundriesio/fioctl/client"
 	"github.com/foundriesio/fioctl/subcommands"
@@ -28,6 +30,14 @@ Use a "fioctl wave complete <wave>" to update all devices (make it globally avai
 Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).`,
 		Run:  doInitWave,
 		Args: cobra.ExactArgs(3),
+		Example: `
+Start a new wave for the target version 4 and the 'production' device tag:
+$ fioctl wave init -k ~/path/to/keys/targets.only.key.tgz wave-name 4 production
+
+Start a new wave for the target version 16 and also prune old production versions 1,2,3 and 4 in this case:
+$ fioctl wave init -k ~/path/to/keys/targets.only.key.tgz wave-name 16 production --prune 1,2,3,4
+
+`,
 	}
 	cmd.AddCommand(initCmd)
 	initCmd.Flags().IntP("expires-days", "e", 0, `Role expiration in days; default 365.
@@ -37,6 +47,8 @@ The same expiration will be used for production targets when a wave is complete.
 When set this value overrides an 'expires-days' argument.
 Example: 2020-01-01T00:00:00Z`)
 	initCmd.Flags().BoolP("dry-run", "d", false, "Don't create a wave, print it to standard output.")
+	initCmd.Flags().StringSlice("prune", []string{}, `Prune old unused Target(s) from the production metadata.
+Example: 1,2,3`)
 	initCmd.Flags().StringP("keys", "k", "", "Path to <offline-creds.tgz> used to sign wave targets.")
 	initCmd.Flags().StringP("source-tag", "", "", "Match this tag when looking for target versions. Certain advanced tagging configurations may require this argument.")
 	_ = initCmd.MarkFlagRequired("keys")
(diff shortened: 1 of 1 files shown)


subcommands/waves/init.go
```diff
@@ -126,6 +132,32 @@ func doInitWave(cmd *cobra.Command, args []string) {
...
+		}
+	}
+	for i, v := range found {
+		if !v {
+			subcommands.DieNotNil(fmt.Errorf("Could not find targets to prune with version = %s", versions[i]))
+		}
+	}
+	for _, name := range keys {
+		delete(currentTargets.Targets, name)
+	}
```
doanac, line 156: It looks like you are going to the trouble of denoting all missing versions in a data structure, but then you just error out with the first missing one. I think you could simplify this logic quite a bit with something more like:
```
var missing []string
for i, version := range versions {
....
    found := false
    for name, file := range currentTargets.Targets {
        if custom.Version == version {
             delete(targets.Targets, name)
             found = true
        } 
    }
    if !found {
        missing = append(missing, version)
    } 
}
if missing != nil && len(missing) > 0 {
    die with strings.Join(missing)
}
```

subcommands/waves/init.go
```diff
@@ -37,6 +37,7 @@ The same expiration will be used for production targets when a wave is complete.
 When set this value overrides an 'expires-days' argument.
 Example: 2020-01-01T00:00:00Z`)
 	initCmd.Flags().BoolP("dry-run", "d", false, "Don't create a wave, print it to standard output.")
+	initCmd.Flags().StringArray("prune", []string{}, "Prune old unused Target(s) from the production metadata.")
```
doanac, line 40: we should add a cli example of this in the usage.

subcommands/waves/init.go
```diff
@@ -49,6 +50,7 @@ func doInitWave(cmd *cobra.Command, args []string) {
 	subcommands.DieNotNil(err, "Version must be an integer")
 	expires := readExpiration(cmd)
 	dryRun, _ := cmd.Flags().GetBool("dry-run")
+	prune, _ := cmd.Flags().GetStringArray("prune")
```
vkhoroz, line 53: I guess we need to add a check against a fool that `version` is not in `prune`.

vkhoroz (commented):
IIUC we need to check the API as well, as it verifies that all prod versions are retained.

doanac:
> IIUC we need to modify the API as well, as it verifies that all prod versions are retained.

Yep - this will just get things to where our API will reject the request. There is backend work required before this could be merged

subcommands/waves/init.go
```diff
@@ -126,6 +148,29 @@ func doInitWave(cmd *cobra.Command, args []string) {
...
 }
 
+func pruneTargets(currentTargets *client.AtsTargetsMeta, versions []string) client.AtsTargetsMeta {
+	var missing []string
+	for _, version := range versions {
+		found := false
+		for name, file := range currentTargets.Targets {
+			custom, err := api.TargetCustom(file)
+			subcommands.DieNotNil(err)
+			if custom.Version == version {
```
vkhoroz, line 158: Using `slices.Contains(versions, custom.Version)` would allow to simplify it to just one loop here.

subcommands/waves/init.go
```diff
@@ -100,6 +113,15 @@ func doInitWave(cmd *cobra.Command, args []string) {
 		targets.Targets[name] = file
 	}
 
+	if len(prune) > 0 {
+		for _, v := range prune {
+			if v == version {
+				subcommands.DieNotNil(fmt.Errorf("Cannot prune current version"))
+			}
+		}
```
vkhoroz, line 121: Maybe, `if slices.Contains(prune, version)`?

subcommands/waves/init.go
```diff
@@ -28,6 +29,14 @@ Use a "fioctl wave complete <wave>" to update all devices (make it globally avai
 Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).`,
 		Run:  doInitWave,
 		Args: cobra.ExactArgs(3),
+		Example: `
+To simply initialize a wave run the following:
+$ fioctl wave init -k ~/path/to/keys/targets.only.key.tgz wave-name 4 production
```
vkhoroz, line 34: I'd paraphrase to a more instructive language here: `Start a new wave for the target version 4 and the 'production' device tag:`

subcommands/waves/init.go
```diff
@@ -28,6 +29,14 @@ Use a "fioctl wave complete <wave>" to update all devices (make it globally avai
...
+
+To initialize a wave and prune some old targets at the same time:
+$ fioctl wave init -k ~/path/to/keys/targets.only.key.tgz wave-name 16 production --prune 1,2,3,4
```
vkhoroz, line 37: Maybe, `Start a new wave for the target version 16 and also prune some old production versions:`?

vkhoroz (approved):
LGTM

subcommands/waves/init.go
doanac: good suggestion. I think katrina has been trying to get us to stop using works like "simply".

StealthyCoder: I adjusted it. :+1:

subcommands/waves/init.go
StealthyCoder: yes we are on high enough golang version.

doanac (approved):

PR #205, opened 2022-12-21
Feature: command to rotate online TUF keys

vkhoroz:

### DIFF
--- a/subcommands/keys/tuf_rotate_all_keys.go
+++ b/subcommands/keys/tuf_rotate_all_keys.go
@@ -1,0 +1,87 @@
+package keys
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/foundriesio/fioctl/subcommands"
+)
+
+func init() {
+	rotate := &cobra.Command{
+		Use:   "rotate-all-keys --keys=<offline-creds.tgz>",
+		Short: "Rotate all online and offline TUF signing keys for the Factory",
+		Long: `Rotate the following TUF keys for the Factory:
+- offline root signing key;
+- offline targets signing key;
+- online targets signing key;
+- online snapshot signing key;
+- online timestamp signing key.
+
+The new signing keys are rotated in both CI and production TUF root transactionally.
+
+When you rotate all TUF signing leys:
+- if there are CI or production targets in your factory, they are re-signed using the new keys.
+- if there is an active wave in your factory, this command is not allowed.
+- new CI targets upload is temporarily disabled for the duration of transaction.`,
+		Example: `
+Migrate an old factory to use Ed25519 key type for all TUF signing keys (online and offline):
+  fioctl keys tuf rotate-all-keys --key-type=ed25519 \
+    --keys=offline-tuf-root-keys.tgz --targets-keys=offline-tuf-targets-keys.tgz`,
+		Run: doRotateAllKeys,
+	}
+	rotate.Flags().StringP("keys", "k", "", "Path to <offline-creds.tgz> used to sign TUF root.")
+	_ = rotate.MarkFlagRequired("keys")
+	_ = rotate.MarkFlagFilename("keys")
+	rotate.Flags().StringP("targets-keys", "K", "", "Path to <offline-targets-creds.tgz> used to sign prod & wave TUF targets.")
+	_ = rotate.MarkFlagFilename("targets-keys")
+	rotate.Flags().BoolP("first-time", "", false, "Used for the first customer rotation. The command will download the initial root key.")
+	rotate.Flags().StringP("key-type", "y", tufKeyTypeNameRSA, "Key type, supported: Ed25519, RSA (default).")
+	rotate.Flags().StringP("changelog", "m", "", "Reason for doing rotation. Saved in root metadata for tracking change history.")
+	tufCmd.AddCommand(rotate)
+}
+
+func doRotateAllKeys(cmd *cobra.Command, unusedArgs []string) {
+	isTufUpdatesShortcut = true
+	keyType, _ := cmd.Flags().GetString("key-type")
+	ParseTufKeyType(keyType) // fails on error
+	changelog, _ := cmd.Flags().GetString("changelog")
+	if changelog == "" {
+		changelog = "Rotate all TUF root signing keys"
+	}
+	firstTime, _ := cmd.Flags().GetBool("first-time")
+	credsFile, _ := cmd.Flags().GetString("keys")
+	targetsCredsFile, _ := cmd.Flags().GetString("targets-keys")
+	if targetsCredsFile == "" {
+		targetsCredsFile = credsFile
+	}
+
+	// Below the `tuf updates` subcommands are chained in a correct order.
+	// Detach from the parent, so that command calls below use correct args.
+	tufCmd.RemoveCommand(tufUpdatesCmd)
+
+	fmt.Println("= Creating new TUF updates transaction")
+	args := []string{"init", "-m", changelog}
+	if firstTime {
+		args = append(args, "--first-time", "-k", credsFile)
+	}
+	tufUpdatesCmd.SetArgs(args)
+	subcommands.DieNotNil(tufUpdatesCmd.Execute())
+
+	args = []string{"rotate-online-key", "-r", "targets,snapshot,timestamp", "-y", keyType}
+	tufUpdatesCmd.SetArgs(args)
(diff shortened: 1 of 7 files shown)


vkhoroz:
@doanac @mike-sul this is also ready for review now

subcommands/keys/tuf_rotate_all_keys.go
```diff
@@ -0,0 +1,87 @@
...
+    --keys=offline-tuf-root-keys.tgz --targets-keys=offline-tuf-targets-keys.tgz`,
+		Run: doRotateAllKeys,
+	}
+	rotate.Flags().StringP("keys", "k", "", "Path to <offline-creds.tgz> used to sign TUF root.")
+	_ = rotate.MarkFlagRequired("keys")
+	_ = rotate.MarkFlagFilename("keys")
+	rotate.Flags().StringP("targets-keys", "K", "", "Path to <offline-targets-creds.tgz> used to sign prod & wave TUF targets.")
+	_ = rotate.MarkFlagFilename("targets-keys")
+	rotate.Flags().BoolP("first-time", "", false, "Used for the first customer rotation. The command will download the initial root key.")
+	rotate.Flags().StringP("key-type", "y", tufKeyTypeNameRSA, "Key type, supported: Ed25519, RSA (default).")
```
StealthyCoder, line 40: I wish we could have something like an Enum-type Flag out of the box. So it is easier to write and make it less likely someone will forget to check the valid values down the road in the actual command execution. :slightly_smiling_face:

vkhoroz: There is this thing: https://pkg.go.dev/github.com/spf13/pflag#Var but I think we could easily go down the rabbit hole with it in Golang, even with the help of generics available in 1.19.

subcommands/keys/tuf_updates_rotate_online_key.go
```diff
@@ -0,0 +1,94 @@
...
+When you rotate the TUF online signing key:
+- if there are CI or production targets in your factory, they are re-signed using the new key.
+- if there is an active wave in your factory, the TUF online key rotation is not allowed.
+- the new wave cannot be created until you apply the online keys rotation.
+
+When you apply the online key rotation, these features are temporarily disabled until it succeeds:
+- new CI targets upload (including the targets upload during CI builds).
+- automatic re-signing of expired TUF roles using online keys (both CI and production targets).
+
+Please, plan the TUF online key rotation timing accurately in order to prevent service disruption.`,
```
doanac, line 31: You may want to elaborate on what you mean here. I sort of read this, got scared and didn't know of anything actionable I could do.

vkhoroz: I think I should just remove the last sentense.

It is the summary for the above two paragraphs: I wanted to put something string enough to let people re-read those two paragraphs.

But if this instead creates the confusion, it is definitely not the desired effect.

vkhoroz: The worst thing that could happen during online key rotation is a failed CI build. That's not bad enough to scare the user.

doanac (approved):
Its working for me. Feel free to merge once you've cleaned up the help text and are satisfied.

subcommands/keys/tuf_updates_apply.go
```diff
@@ -34,6 +35,13 @@ func doTufUpdatesApply(cmd *cobra.Command, args []string) {
 		msg := "Failed to apply staged TUF root updates:\n%w\n"
 		var isNonFatal bool
 		if herr := client.AsHttpError(err); herr != nil {
+			if herr.Response.StatusCode == 404 {
+				// Double check: if there are no TUF updates - fail clean; otherwise, fatal error.
+				updates, err1 := api.TufRootUpdatesGet(factory)
+				if err1 == nil && updates.Status == client.TufRootUpdatesStatusNone {
+					subcommands.DieNotNil(errors.New("There are no TUF root updates in progress."))
+				}
+			}
```
mike-sul, line 44: If `err1` is not nil then the command will print 
```
This is a critical error: Staged TUF root updates may be partially applied to your Factory.
Please, re-run the "fioctl keys tuf updates apply" again after some time, but as soon as possible.
If the error persists, please, contact the customer support.`
```
It looks like too scary message for 404.

subcommands/keys/tuf_updates_rotate_online_key.go
```diff
@@ -0,0 +1,87 @@
...
+		Example: `
+- Rotate online TUF targets key and re-sign the new TUF root:
+  fioctl keys tuf updates rotate-online-key \
+    --txid=abc --role=targets --keys=tuf-root-keys.tgz --sign
```
mike-sul, line 35: Why the `--sign` option is needed? I thought that root meta signing is part of the rotation process.
Or this is for the case if a user would like to rotate a few key types (a.g. targets and snapshot) and then sign once.

mike-sul: And there is `sign` sub-command.

mike-sul: And the `apply` command doesn't do automatic signing.

mike-sul: I would rather do signing for each operation or automatically sign on `apply`. I think a user will always forget to run the `sign` command before running `apply`.

doanac: I think Voldodya's going to say: "We went through this in the design review and all agreed on this flow". I think that's fair. Also - this is *expert* mode. I kind of liked the granularity when I tested this out. I could do each step and also do the `review` subcommand to observe.

vkhoroz: Thanks @doanac ! You answered 80% of the question.

The other 20% is that for offline targets key rotation, the root.json is normally not signed by the same person who rotates it. Because, from the security standpoint Root Officer versus Targets Officer should be two different persons.

Although it is technically possible that Root Officer rotates the key and hands it over to the Targets Officer - that's a backward workflow. Furthermore, in one of the future releases we'll introduce the HSM support, which will make this backwards behavior impossible.

So, for the `rotate-offline-key`, the signature is optional; but I added the `--sign` argument for the user's convenience, as 80% of users will do it this way. Once I do that, other commands like e.g. `rotate-online-key` have to follow the same pattern to share the common _look and feel_; and we need a standalone `sign` command.

subcommands/keys/tuf_updates_apply.go
vkhoroz: If we get 404 - we check if the updates transaction exist:
- if it doesn't - that's regular situation, and we show a simple regular error.
- if it exists, and there was 404 - only the universe knows why it happened, so it is critical.

mike-sul: The `api.TufRootUpdatesGet(factory)` can fail so `err1` is not nil if there is no networking connection e.g. And the command will print that scary message then, instead of connection error?

vkhoroz: There are 3 use cases:
- `err == nil & there is no transtaction` - we know why 404 happened, and this is a regular situation.
- `err == nil & there is transtaction` - we don't know why 404 happened, this is bad. So, we show critical error.
- `err != nil` - we don't know what we don't know, this is the worst situation possible. So, we show critical error _with the original error message_.

mike-sul: If `err1 != nil` we know the exact error, it's in the `err1` context. As I mentioned in the example above I would prefer to see "connection/timeout error" instead of that scary message.

Anyway, it's not really important.

vkhoroz: > As I mentioned in the example above I would prefer to see "connection/timeout error" instead of that scary message.

Here is the thing:
- we apply updates and get 404 error;
- we try to verify if it is critical or not... for that we see if the transaction exists;
- so, checking for transaction is a tool to evaluate the error severity; it has no other value beyond that.
- therefore, if we get an error while ~checking if transaction exists~ assessing the error severity - that severity gets a maximum level.
- ...because we know that something bad has happened, and we cannot assess what exactly.

subcommands/keys/tuf_updates_rotate_online_key.go
vkhoroz: Removed the last paragraph in the force-push

vkhoroz:
rebased on main and re-signed

PR #206, opened 2023-01-03
fix(git): Use UID instead of name

StealthyCoder:
Hopefully this will be more robust in finding the necessary users, rather than relying on names.


doanac:
i don't like this new approach because it would break my laptop where I don't have anything in $PATH that's writable by my user.

I think we need to step back on this for a moment and understand/reproduce the real issue before trying to fix something. This might not fix the actual problem.

StealthyCoder:
Closing this because of #219

PR #208, opened 2023-01-17
Improve the targets prune help by adding examples

vkhoroz:

### DIFF
--- a/subcommands/targets/prune.go
+++ b/subcommands/targets/prune.go
@@ -26,6 +26,21 @@ func init() {
 		Short: "Prune target(s)",
 		Run:   doPrune,
 		Args:  cobra.MinimumNArgs(1),
+		Example: `
+  # prune a single target by name:
+  fioctl targets prune intel-corei7-64-lmp-123
+
+  # prune all targets with the "custom.tags" set to ["devel"]:
+  fioctl targets prune --by-tag devel
+
+  # prune all targets with the "custom.tags" set to ["devel"] except for the most recent 10:
+  fioctl targets prune --by-tag devel --keep-last=10
+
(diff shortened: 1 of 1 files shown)


subcommands/targets/prune.go
```diff
@@ -26,6 +26,21 @@ func init() {
 		Short: "Prune target(s)",
 		Run:   doPrune,
 		Args:  cobra.MinimumNArgs(1),
+		Example: `
+  # prune a single target name its name:
```
vanmaegima, line 30:
```suggestion
  # prune a single target by its name:
```

doanac: i think other parts of fioctl talk about `target name` and `target version`. So maybe we do `prune a single target name`

subcommands/targets/prune.go
```diff
@@ -26,6 +26,21 @@ func init() {
...
+  fioctl targets prune intel-corei7-64-lmp-123
+
+  # prune all targets with the "custom.tags" set to ["devel"]:
+  fioctl targets prune --by-tag devel
+
+  # prune all targets with the "custom.tags" set to ["devel"] except for the last 10:
```
doanac, line 36: I like "last 10" but i wonder if "most recent 10" would make people more comfortable? I don't have a strong opinion.

subcommands/targets/prune.go
vkhoroz: This is what happens when one spends less then 5 minutes on the commit.
Fixed in force-push.

subcommands/targets/prune.go
vkhoroz: Changed in force-push.

doanac (approved):
a could usage statement is more valuable than 1000 lines of backend validation improvements :)

PR #211, opened 2023-01-25
Feature: ability to show tuf updates in human readable and diff formats

vkhoroz:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/docker/go v1.5.1-1.0.20160303222718-d30aec9fd63c
 	github.com/fatih/color v1.13.0
 	github.com/foundriesio/go-ecies v0.3.0
+	github.com/karrick/godiff v0.0.2
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/pelletier/go-toml v1.9.4
 	github.com/shurcooL/go v0.0.0-20200502201357-93f07166e636
--- a/subcommands/keys/tuf_updates_review.go
+++ b/subcommands/keys/tuf_updates_review.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/fatih/color"
 	"github.com/karrick/godiff"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -90,7 +91,15 @@ func doTufUpdatesReview(cmd *cobra.Command, args []string) {
 				strings.Split(string(before), "\n"),
 				strings.Split(string(after), "\n"),
 			)
-			fmt.Println(strings.Join(diff, "\n"))
+			for _, line := range diff {
+				if line[0] == '+' {
+					color.Green(line)
+				} else if line[0] == '-' {
+					color.Red(line)
+				} else {
+					fmt.Println(line)
+				}
+			}
 		}
 	} else {
 		fmt.Println("The following TUF root changes are staged for your factory:")
--- a/subcommands/keys/tuf_updates_review.go
+++ b/subcommands/keys/tuf_updates_review.go
@@ -2,6 +2,7 @@ package keys
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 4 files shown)


subcommands/keys/tuf_updates_review.go
```diff
@@ -17,35 +18,61 @@ func init() {
...
+	if showRaw {
+		var newProdRoot, rootToShow *client.AtsTufRoot
+		if updates.Updated.ProdRoot != "" {
+			subcommands.DieNotNil(
+				json.Unmarshal([]byte(updates.Updated.ProdRoot), &newProdRoot), "Updated prod root",
+			)
+		}
+		if newProdRoot == nil {
+			// No effective changes yet, but we know how the prod root would look like
+			newProdRoot = genProdTufRoot(newCiRoot)
```
doanac, line 56: I think I see your point here - just show the actual "raw" content. But you can't apply this - so there's really nothing to review. I wonder if the old approach of saying "There are no staged TUF updates." would cause less confusion.

subcommands/keys/tuf_updates_review.go
```diff
@@ -56,9 +61,37 @@ func doTufUpdatesReview(cmd *cobra.Command, args []string) {
...
+
+			before, err := subcommands.MarshalIndent(baseRootToShow, "", "  ")
+			subcommands.DieNotNil(err)
+			after, err := subcommands.MarshalIndent(rootToShow, "", "  ")
+			subcommands.DieNotNil(err)
+			diff := godiff.Strings(
+				strings.Split(string(before), "\n"),
+				strings.Split(string(after), "\n"),
+			)
+			fmt.Println(strings.Join(diff, "\n"))
```
doanac, line 93: The output is a little weird. It shows *everything*. I guess I was expecting classic unified diff where you only include a few context lines around the change. Still probably handy enough that someone will use it.

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/docker/go v1.5.1-1.0.20160303222718-d30aec9fd63c
 	github.com/fatih/color v1.13.0
 	github.com/foundriesio/go-ecies v0.3.0
+	github.com/karrick/godiff v0.0.2
```
doanac, line 11: i wonder if its worth carrying this whole external dependency? It might be easier to just copy the file with license/attribution comment at the top?

subcommands/keys/tuf_updates_review.go
```diff
@@ -56,9 +61,37 @@ func doTufUpdatesReview(cmd *cobra.Command, args []string) {
...
+		}
```
doanac, line 94: We already have the `github.com/fatih/color` module. You could probably look at `line[0] == '+'` and make them green and `line[0] == '-'` and make them red.

subcommands/keys/tuf_updates_review.go
```diff
@@ -17,35 +18,61 @@ func init() {
...
+			rootToShow = newProdRoot
+		} else {
+			rootToShow = newCiRoot
+		}
+		bytes, err := subcommands.MarshalIndent(rootToShow, "", "  ")
+		subcommands.DieNotNil(err)
+		fmt.Println(string(bytes))
 	} else {
-		rootToShow = newCiRoot
+		fmt.Println("The following TUF root changes are staged for your factory:")
```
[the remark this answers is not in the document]
vkhoroz: The API always returns at least 2 amendments for a fresh `updates` object: version increment and new changelog.
So `len == 0` could only mean the API error here; so I'd better let the user complain if the list is empty.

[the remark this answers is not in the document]
doanac: not if you run `updates init` then `updates review`:
```
$ fioctl keys tuf updates review 
The following TUF root changes are staged for your factory:
```

subcommands/keys/tuf_updates_review.go
vkhoroz: I think the initial implementation was underthought.
There is something to show here: the version increment and the new changelog message.

In case of human readable format the user would see who created the transaction (even if there are no other changes yet).
I find that information valuable; regardless of the fact that it cannot be applied yet.

subcommands/keys/tuf_updates_review.go
vkhoroz: I was trying to find a Golang library which does it exactly like the diff utility; but it was only a part of the `golang.org/x/tools/internal/lsp/diff` which is long gone. So, maybe, I should call the `diff` commandline tool instead, I was just not sure if it is present on every system.

At the end, I like how this minimalistic tool gives the full context; the root.json is not that large.
I could write a small function which filters lines far away from the diff. I just wanted to keep the scope manageable.

subcommands/keys/tuf_updates_review.go
vkhoroz: Makes sense. I just wanted to keep the scope manageable; so waited for someone to make a comment on this.

subcommands/keys/tuf_updates_review.go
doanac: let's not overthink it. this is good enough.

go.mod
vkhoroz: We already carry more over 100 dependencies; one more makes no difference. Copying would mean that fioctl will have a larger codebase; not sure it is worth to do that.

FYI the `proxy.golang.org` package manager gives a promise that no package can ever be removed; even after the github project was removed. So, we're safe here.

subcommands/keys/tuf_updates_review.go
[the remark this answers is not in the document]
vkhoroz: You've probably tested it in production, but the API changes are not deployed there yet.

This is what it gives me in staging:
```
The following TUF root changes are staged for your factory:
 - User <Volodymyr Khoroz (id=5f7b2f1a17c799bafeacfa52)> at 2023-01-26T15:56:27Z set change message to: test amendments in stag2
 - Changed version from 6 to 7

There are no effective changes to your TUF root except the version and changelog.
These changes cannot be applied before some of the key rotation commands are performed.
Please, run the 'fioctl keys tuf updates --help' for details.
```

doanac (commented):
Looks good. Merge this or add color - not a big deal to me.

subcommands/keys/tuf_updates_review.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/211/commits/57b0f5532cbd1efaecb77649ef119aa7cded93ce

PR #214, opened 2023-02-02
feat(waves): Implement pagination

StealthyCoder:

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -410,6 +410,12 @@ type WaveCreate struct {
 	Targets tuf.Signed `json:"targets"`
 }
 
+type WaveList struct {
+	Waves []Wave  `json:"waves"`
+	Total int     `json:"total"`
+	Next  *string `json:"next"`
+}
+
 type WaveRolloutOptions struct {
 	Group string `json:"group"`
 }
@@ -1649,8 +1655,8 @@ func (a *Api) FactoryCreateWave(factory string, wave *WaveCreate) error {
 	return err
 }
 
-func (a *Api) FactoryListWaves(factory string, limit uint64) ([]Wave, error) {
-	url := a.serverUrl + "/ota/factories/" + factory + "/waves/?limit=" + strconv.FormatUint(limit, 10)
+func (a *Api) FactoryListWaves(factory string, limit uint64, page int) (*WaveList, error) {
+	url := a.serverUrl + "/ota/factories/" + factory + "/waves/?limit=" + strconv.FormatUint(limit, 10) + "&page=" + strconv.Itoa(page)
 	logrus.Debugf("Listing factory waves %s", url)
 
 	body, err := a.Get(url)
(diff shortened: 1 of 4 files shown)


subcommands/waves/list.go
```diff
@@ -44,4 +49,25 @@ func doListWaves(cmd *cobra.Command, args []string) {
...
+				found = true
+				i++
+			} else {
+				fmt.Print(os.Args[i], " ")
+			}
+		}
+		if !found {
+			fmt.Print("-p", showPage+1)
+		}
+		fmt.Println()
```
vkhoroz, line 71: I've already seen an almost identical code somewhere. It should be feasible to extract the common part into a function in e.g. `subcommands/common.go` and reuse in both places.

client/foundries.go
```diff
@@ -1648,18 +1654,18 @@ func (a *Api) FactoryCreateWave(factory string, wave *WaveCreate) error {
...
 
 	body, err := a.Get(url)
 	if err != nil {
 		return nil, err
 	}
 
-	var resp []Wave
-	err = json.Unmarshal(*body, &resp)
-	return resp, err
+	waves := WaveList{}
```
vkhoroz, line 1666: Nitpick: `var waves WaveList` is how we currently do it. I'm not sure about the tiny nuances of either way.

Visually, the biggest difference is that in case of an error the `var` way will return `nil`, but the `:=` way will return a partially initialized object.

StealthyCoder: I do the exact same as for DeviceList to be consistent. But if you want we can change it for this one.

vkhoroz: That was a nipick, so you can ignore it.

subcommands/waves/list.go
vkhoroz: Thanks

vkhoroz:
LGTM.

Please, make sure to finish and merge the API changes _before_ merging this.

subcommands/waves/list.go
```diff
@@ -18,15 +20,14 @@ func init() {
...
 
 func doListWaves(cmd *cobra.Command, args []string) {
 	factory := viper.GetString("factory")
 	limit, _ := cmd.Flags().GetUint64("limit")
 	showPage, _ := cmd.Flags().GetInt("page")
-	if showAll, _ := cmd.Flags().GetBool("all"); showAll {
-		limit = 0
+	if limit > 100500 || limit <= 0 {
+		subcommands.DieNotNil(errors.New("Limit needs to be between 1 and 100500 inclusive."))
 	}
```
vkhoroz, line 31: I'm not sure we need this _easter egg_ here.

In other places we only allow _specific limit values_; but here it give no benefit.
So, I'd just remove `--all` and allow any limit value.

subcommands/waves/list.go
```diff
@@ -1,6 +1,8 @@
 package waves
 
 import (
+	"errors"
+
```
vkhoroz, line 5: This is now unused. I guess this is why it fails.

vkhoroz (approved):
LGTM once you fix imports.

PR #219, opened 2023-02-25
Fixes the git credential helper install logic

doanac:
This addresses some issues found on Windows and MacOS

### DIFF
--- a/subcommands/git/cmd_unix.go
+++ b/subcommands/git/cmd_unix.go
@@ -1,0 +1,10 @@
+//go:build !windows
+// +build !windows
+
+package git
+
+import "golang.org/x/sys/unix"
+
+func isWritable(dir string) bool {
+	return unix.Access(dir, unix.W_OK) == nil
+}
--- a/subcommands/git/cmd_windows.go
+++ b/subcommands/git/cmd_windows.go
@@ -1,0 +1,16 @@
+//go:build windows
+// +build windows
+
+package git
+
+import (
+	"syscall"
+)
+
+func isWritable(dir string) bool {
+	if hwnd, err := syscall.CreateFile(syscall.StringToUTF16Ptr(dir), 2, 0, nil, syscall.OPEN_EXISTING, syscall.FILE_FLAG_BACKUP_SEMANTICS|syscall.FILE_FLAG_OPEN_REPARSE_POINT, 0); err == nil {
+		_ = syscall.CloseHandle(hwnd)
+		return true
+	}
+	return false
+}
--- a/subcommands/git/cmd.go
+++ b/subcommands/git/cmd.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"os/user"
 	"path/filepath"
+	"runtime"
 
 	"github.com/foundriesio/fioctl/subcommands"
 	"github.com/sirupsen/logrus"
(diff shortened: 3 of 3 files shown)


doanac:
@munoz0raul - these two changes get things working on my Mac and Windows systems.

subcommands/git/cmd_unix.go
```diff
@@ -0,0 +1,10 @@
+//go:build !windows
+// +build !windows
+
+package git
+
+import "golang.org/x/sys/unix"
+
+func isWritable(dir string) bool {
+	return unix.Access(dir, unix.W_OK) == nil
```
StealthyCoder, line 9: I think this is the same as in my PR #206 where I use `syscall.Access(p, syscall.O_RDWR)` .

subcommands/git/cmd_windows.go
```diff
@@ -0,0 +1,23 @@
...
+		return false
+	}
+	if !info.IsDir() {
+		return false
+	}
+	// Check if the user bit is enabled in file permission
+	if info.Mode().Perm()&(1<<(uint(7))) == 0 {
+		return false
+	}
+	return true
```
StealthyCoder, line 22: I tried this approach too as I found it online some where, then I thoroughly tested it and found out it just always will work. Regardless if you can actually write to the dir or not. In PR #206 I made the following and that is direct translation of what happens in Unix:

```go
if hwnd, err := syscall.CreateFile(syscall.StringToUTF16Ptr(p), FILE_APPEND_FILE, FILE_LOCK, nil, syscall.OPEN_EXISTING, syscall.FILE_FLAG_BACKUP_SEMANTICS|syscall.FILE_FLAG_OPEN_REPARSE_POINT, 0); err == nil {
				candidates = append(candidates, p)
				if err = syscall.CloseHandle(hwnd); err != nil {
					return "", err
				}
			}

```

Where p is the path to be checked. Can we please use this implementation?

StealthyCoder (changes requested):
If we tighten up the windows implementation, this is good to go.

subcommands/git/cmd_windows.go
doanac: sorry - the title of that PR was misleading so i hadn't looked at it for doing this. This is quite a complicated API call, can you explain what it's doing? Almost looks like a really fancy file open/close might work?

doanac: actually - i'm testing my logic and it works. It's kind of hard to make a directory read-only or not writable to a user, but I've created a director `c:\test` owned by user "foo". My administrator account can't write to it and it returns false. There is some insane issue in Windows where my *bash* shell can still write to the folder but windows cmd.exe can't and my program from either shell will say its not writable.

StealthyCoder: Sorry for the misleading title. So I wrote the following test program: 

```go
var p = "C:\\Program Files\\"
        if hwnd, err := syscall.CreateFile(syscall.StringToUTF16Ptr(p), 2, 0, nil, syscall.OPEN_EXISTING, syscall.FILE_FLAG_BACKUP_SEMANTICS|syscall.FILE_FLAG_OPEN_REPARSE_POINT, 0); err == nil {
                if err = syscall.CloseHandle(hwnd); err != nil {
                        fmt.Println(err)
                }
                fmt.Println("Create files/dir: Yes")
        } else {
                fmt.Println("Create files/dir: No")
        }

        info, _ := os.Stat(p)
        if info.Mode().Perm()&(1<<(uint(7))) == 0 {
                fmt.Println("Create files/dir: No")
        } else {
                fmt.Println("Create files/dir: Yes")
        }
```

I run it on my Windows 10 Intel NUC machine I have here, so not in a VM (maybe that makes a difference). The outcome of it is that both in PowerShell and in Command Prompt (CMD.exe) from the Windows Terminal app I get the same response. For my normal user I cannot write to the "C:\Program Files" directory. Which is correct. However for my Administrator elevated prompts I also get that I cannot write to the directory. Which is incorrect. 

Normal User

Elevated User

I will explain the API call in separate comment.

StealthyCoder: `syscall.CreateFile(syscall.StringToUTF16Ptr(p), 2, 0, nil, syscall.OPEN_EXISTING, syscall.FILE_FLAG_BACKUP_SEMANTICS|syscall.FILE_FLAG_OPEN_REPARSE_POINT, 0)` 

I will try to break this down as best I can. So `CreateFile` calls Win32 `CreateFileW` under the hood. 
> To obtain a handle to an existing directory, call the [CreateFile](https://learn.microsoft.com/en-us/windows/desktop/api/FileAPI/nf-fileapi-createfilea) function with the FILE_FLAG_BACKUP_SEMANTICS flag.

The parameters are as follows:

1. string pointer to the path
2. file access level , 2 means FILE_ADD_FILE, For a directory, the right to create a file in the directory.
3. share mode level, 0 means Prevents subsequent open operations on a file or device if they request delete, read, or write access. 
4. optional security attributes
5. creation disposition, OPEN_EXISTING means Opens a file or device, only if it exists. otherwise fail
6. flags and attributes
    FILE_FLAG_BACKUP_SEMANTICS needed because opening a handle to a directory
    FILE_FLAG_OPEN_REPARSE_POINT is needed because normal reparse point processing will not occur and if it is not a reparse point it  will ignore this. This has to do with mounted volumes amongst other things
7. optional file template so therefore 0.

This actually goes through the Win32 API and I am assuming does all the correct checks like you would expect syscalls to the linux kernel to do.

doanac: thanks for the explanation. fixed in force-push

doanac:
@StealthyCoder - i'm going to create a release for this. Do you think you can test the resulting binaries on Windows to help confirm?

StealthyCoder:
> @StealthyCoder - i'm going to create a release for this. Do you think you can test the resulting binaries on Windows to help confirm?

Of course. :+1:

StealthyCoder:
@doanac Just tested the latest release. If git is not installed I get a nice error stating git is not installed. Then for a normal user it places the symlink in AppData\Local and if I am an elevated user it places it where `git` executable is located. These screens are just here for you to confirm that is expected behaviour. 

Git not Found

Normal User Symlink

Elevated User

PR #220, opened 2023-02-27
fix(docker): Same improvements as git helper

StealthyCoder:
Just added the same improvements as the git helper to make the DX better across all platforms.

### DIFF
--- a/subcommands/docker/cmd.go
+++ b/subcommands/docker/cmd.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"os/user"
 	"path/filepath"
+	"runtime"
 
 	"github.com/foundriesio/fioctl/subcommands"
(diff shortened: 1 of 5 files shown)


subcommands/docker/cmd.go
```diff
@@ -130,3 +140,31 @@ func RunCredsHelper() int {
 	os.Stdout.Write(bytes)
 	return 0
 }
+
+// Find an entry in the PATH we can write to. For example, on MacOS git is
+// installed under /usr/bin but even root can't write to that because of
+// filesystem protection logic they have.
+func findWritableDirInPath(gitPath string) string {
```
doanac, line 147: i would have done it this way also, but after a little thinking - I think we should move this to subcommands/common*.go and have both docker/git use the same thing.

doanac (approved):

PR #222, opened 2023-02-27
fix(docker): Safeguard config.json not being found

StealthyCoder:
This PR just adds a way to safeguard against the potential situation that someone just installs docker but never runs it yet. In that case there is no `.docker` dir created and no `config.json`.

In addition the reason the error checking is changed is because of an [internal comment of golang itself](https://pkg.go.dev/os#IsNotExist).

### DIFF
--- a/subcommands/common.go
+++ b/subcommands/common.go
@@ -233,7 +233,7 @@ func FindWritableDirInPath(helperPath string) string {
 	// Give preference to helper executable location if its in PATH
 	if len(helperPath) > 0 {
 		if _, ok := paths[helperPath]; ok {
-			if isWritable(helperPath) {
+			if IsWritable(helperPath) {
 				return helperPath
 			}
 		}
@@ -241,7 +241,7 @@ func FindWritableDirInPath(helperPath string) string {
 
 	// Now try everything
 	for _, path := range filepath.SplitList(path) {
-		if isWritable(path) {
+		if IsWritable(path) {
 			return path
 		}
 	}
--- a/subcommands/docker/cmd.go
+++ b/subcommands/docker/cmd.go
@@ -29,6 +29,9 @@ func NewCommand() *cobra.Command {
 	}
 	helperPath = subcommands.FindWritableDirInPath(helperPath)
 	dockerConfigFile = dockerConfigPath()
+	if !subcommands.IsWritable(dockerConfigFile) {
+		dockerConfigFile = ""
+	}
 
 	cmd := &cobra.Command{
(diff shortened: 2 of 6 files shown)


subcommands/common.go
```diff
@@ -247,3 +247,10 @@ func FindWritableDirInPath(helperPath string) string {
 	}
 	return ""
 }
+
+func IsWritablePath(path string) string {
```
doanac, line 251: An `Is*` function returning a string rather than a boolean is a little awkward. Looks like you are trying to save a couple of lines of code below, but I'd just make `isWritable` a public function and use that.

subcommands/docker/cmd.go
```diff
@@ -86,9 +88,9 @@ func doDockerCreds(cmd *cobra.Command, args []string) {
 
 	var config map[string]interface{}
 	bytes, err := os.ReadFile(dockerConfigFile)
-	if os.IsNotExist(err) {
+	if errors.Is(err, fs.ErrNotExist) {
 		dockerConfig := filepath.Dir(dockerConfigFile)
-		if _, err := os.Stat(filepath.Dir(dockerConfig)); os.IsNotExist(err) {
+		if _, err := os.Stat(dockerConfig); errors.Is(err, fs.ErrNotExist) {
```
doanac, line 93: Is this really a 2nd change? i.e. the *fix* to your problem is checking if the path is writable. If so, `dockerConfigFile` becomes empty and a required flag. And this this is modernizing our error checking?

subcommands/common.go
vkhoroz: I concur on this. Other than this your PR LGTM.

subcommands/docker/cmd.go
```diff
@@ -29,6 +29,9 @@ func NewCommand() *cobra.Command {
 	}
 	helperPath = subcommands.FindWritableDirInPath(helperPath)
 	dockerConfigFile = dockerConfigPath()
+	if !subcommands.IsWritable(dockerConfigFile) {
+		dockerConfigFile = ""
+	}
```
vkhoroz, line 36: So, there are 3 modernizr commits; and then this one line of code which _fixes the thing_, right?
Can you extract it into a separate commit with a loud message like _it fixes xyz_?

vkhoroz (approved):
LGTM

I think this is the case when _one merge commit_ to main is better than _3-4 commits created by rebase_.
Please, remember to hit the correct button when closing this PR (`Squash and merge` in GitHub terms).

PR #223, opened 2023-02-28
refactor(errors): Refactoring file not exist error

StealthyCoder:
Making sure all error checking is done according to the new preferred way according to golang itself.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,7 +1,9 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -104,7 +106,7 @@ func getConfigDir() string {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	if _, err := os.Stat(config); os.IsNotExist(err) {
+	if _, err := os.Stat(config); errors.Is(err, fs.ErrNotExist) {
```
vkhoroz, line 109: I've seen this warning in new Golang documentation:
```
This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist). 
```
So, in our case:
- Is there any functional change in behavior? Did you spot a related bug?
- Or, is this just a response to the mentioned Golang warning?

This is just a question to have more context.

StealthyCoder: That last part. I just happened to see the warning and thought as part of trying to fix a bug in #222  I came across this warning. I was trying to apply the [boy scout rule](https://matheus.ro/2017/12/11/clean-code-boy-scout-rule/).

StealthyCoder:
I am actually closing this one, as it became a part of #222 after some feedback from @doanac .

PR #224, opened 2023-03-03
Feature: filter device list by target name pattern

vkhoroz:

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -784,15 +784,17 @@ func (a *Api) DeviceGet(factory, device string) (*Device, error) {
 	return &d, nil
 }
 
-func (a *Api) DeviceList(mine bool, matchTag, byFactory, byGroup, nameIlike, uuid string, page, limit int) (*DeviceList, error) {
+func (a *Api) DeviceList(
+	mine bool, matchTag, byFactory, byGroup, nameIlike, uuid, byTarget string, page, limit int,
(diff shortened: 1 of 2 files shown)


client/foundries.go
```diff
@@ -784,15 +784,17 @@ func (a *Api) DeviceGet(factory, device string) (*Device, error) {
...
 	mineInt := 0
 	if mine {
 		mineInt = 1
 	}
 	url := a.serverUrl + "/ota/devices/?"
 	url += fmt.Sprintf(
-		"mine=%d&match_tag=%s&name_ilike=%s&factory=%s&uuid=%s&group=%s&page=%d&limit=%d",
-		mineInt, matchTag, nameIlike, byFactory, uuid, byGroup, page, limit)
+		"mine=%d&match_tag=%s&name_ilike=%s&factory=%s&uuid=%s&group=%s&target_name=%s&page=%d&limit=%d",
+		mineInt, matchTag, nameIlike, byFactory, uuid, byGroup, byTarget, page, limit)
```
doanac, line 797: we are almost to the point of the `WithOption` type pattern the Golang people use for unwieldy functions. This is an achievement :)

doanac (approved):

client/foundries.go
vkhoroz: yes, we are... but someone also needs to fight the lazyness pattern before that :D

PR #225, opened 2023-03-03
Feature: allow sorting by last-seen and/or name asc/desc

vkhoroz:

### DIFF
--- a/subcommands/devices/list.go
+++ b/subcommands/devices/list.go
@@ -103,6 +103,27 @@ func addPaginationFlags(cmd *cobra.Command) {
 	cmd.Flags().IntVarP(&paginationLimit, "limit", "n", 500, "Number of devices to paginate by. Allowed values: "+limitsStr)
 }
 
+func addSortFlag(cmd *cobra.Command, flag, short, help string) {
+	// Allows `--flag` to specify ascending sort order or a regular `--flag=asc`, `--flag=desc`.
+	cmd.Flags().StringP(flag, short, "", help)
+	cmd.Flags().Lookup(flag).NoOptDefVal = "asc"
+}
+
+func appendSortFlagValue(sortBy []string, cmd *cobra.Command, flag string, field string) []string {
+	val, _ := cmd.Flags().GetString(flag)
+	switch val {
+	case "":
+		break
+	case "asc":
+		sortBy = append(sortBy, field)
+	case "desc":
+		sortBy = append(sortBy, "-"+field)
+	default:
+		subcommands.DieNotNil(fmt.Errorf("Only 'asc, desc' values are allowed for %s but received: %s", flag, val))
+	}
+	return sortBy
+}
+
 func init() {
 	var defCols = []string{
 		"name", "target", "status", "apps", "up-to-date", "is-prod",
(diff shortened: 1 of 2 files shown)


subcommands/devices/list.go
```diff
@@ -103,6 +103,27 @@ func addPaginationFlags(cmd *cobra.Command) {
 	cmd.Flags().IntVarP(&paginationLimit, "limit", "n", 500, "Number of devices to paginate by. Allowed values: "+limitsStr)
 }
 
+func addSortFlag(cmd *cobra.Command, flag, short, help string) {
+	// Allows `--flag` to specify ascending sort order or a regular `--flag=asc`, `--flag=desc`.
+	cmd.Flags().StringP(flag, short, "", help)
```
doanac, line 108: couldn't you put "asc" here instead of an empty string so that it defaults that way and is also visible in help text as to the default value?

vkhoroz: Default is to sort by `owner,name` - the way it was before.

The "" default means to not use this flag.

doanac: okay. so if i run `fioctl devices list --sort-by-last-seen=asc` - this is purely by the last-seen attribute and totally drops the "owner/name" sorting?

I also get confused with doing both new sort options and knowing which gets priority in sorting. I wonder if we need it more like: `--sort-by=-name,last-seen` or `--sort-by=name:ASC,last-seen:DESC`

vkhoroz: My initial idea was to do `--sort-by=-name,last-seen`, but then I realized that we have another UI problem:
- either support esoteric sort options like `name,owner,last-seen` which make no sense, because a name is unique;
- or have hard time explaining to the user which options make sense and why.

So, I came up with an idea that the user will only ever need 3 modes:
- old way to group by owner, and then sort by name (default);
- sort by name: `--sort-by-name | --sort-by-name=asc` (alphabetically) or `--sort-by-name=desc` (reverse).
- sort by last-seen: `--sort-by-last-seen | --sort-by-last-seen=asc` (inactive first) or `--sort-by-last-seen=desc` (active first), and then sort duplicates by name.

I theory, a user may try `--sort-by-last-seen --sort-by-name`, but it is technically equivalent to just `--sort-by-last-seen`.
I just didn't want to bother the user about the exclusivity warning.
To me, it seemed obvious that using these two flags it only makes sense to sort by `last-name` first, because the `name` field is unique.

vkhoroz: @doanac Maybe, we can try this to be as simple as this:
```
--sort-by=owner  A field to sort devices, one of owner (default), name, -name, last_seen, -last_seen.
```

I still see subtle UI problems with this:
- need to explain to the user what `-name,-last_seen` means.
- A `-` sign will confuse the argument parser, e.g. when used like `--sort-by -name`, it may interpret `-name` as a `-n` option with argument `ame`. So, maybe we need `name_descending`, which brings me back to the confusing UI.

doanac: maybe we just flag the two sort-by options you've denoted as mutually-exclusive. Otherwise someone will try it and ping customer support.

vkhoroz: > as mutually-exclusive

I didn't want that; but it actually makes no big difference. I cannot imagine the user using both options together.

So, I'll do it to save our souls from the high entropy.

vkhoroz: added exclusive flag in force-push

doanac (approved):

PR #227, opened 2023-03-06
Bump golangci-ling to 1.51.2

vkhoroz:
I'm helping with LmP upgrade of Golang to 1.20 for fioconfig. So, my system-wide Golang version was upgraded to version 1.20. But, for some weird reason, older golangci-lint stopped working after that regardless of the go1.19 flag in go.mod.
It looks like it uses some of the Golang internals. The easiest solution is to upgrade golangci-lint to a newer version, which support Golang 1.17+.

Looking ahead, this is a good addition anyway to stay tuned.

### DIFF
--- a/.jobserv.yml
+++ b/.jobserv.yml
@@ -4,7 +4,7 @@ triggers:
     type: github_pr
     runs:
       - name: unit-test
-        container: golangci/golangci-lint:v1.49
+        container: golangci/golangci-lint:v1.51.2
         host-tag: amd64
         script: unit-test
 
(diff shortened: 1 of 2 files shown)


vkhoroz:
Maybe, these tickets help:
- https://github.com/golangci/golangci-lint/issues/2374
- https://github.com/golangci/golangci-lint/pull/3037
- https://github.com/golangci/golangci-lint/issues/3420

In short: we need to upgrade golint every time Golang releases a new major version and we plan to support it.

doanac (approved):

PR #228, opened 2023-03-17
wireguard: Don't allow a .0 device ip

doanac:
This is not a valid value

### DIFF
--- a/subcommands/devices/config_wireguard.go
+++ b/subcommands/devices/config_wireguard.go
@@ -117,7 +117,7 @@ func findVpnAddress(factory string) string {
 
 	ips := factoryIps(factory)
 	for ip := serverIp + 1; ip < serverIp+10000; ip++ {
-		if _, ok := ips[ip]; !ok {
+		if _, ok := ips[ip]; !ok && byte(ip) != 0 {
 			logrus.Debugf("Found unique ip: %d", ip)
 			return fmt.Sprintf("%d.%d.%d.%d", byte(ip>>24), byte(ip>>16), byte(ip>>8), byte(ip))
 		}


mike-sul (approved):

subcommands/devices/config_wireguard.go
```diff
@@ -117,7 +117,7 @@ func findVpnAddress(factory string) string {
 
 	ips := factoryIps(factory)
 	for ip := serverIp + 1; ip < serverIp+10000; ip++ {
-		if _, ok := ips[ip]; !ok {
+		if _, ok := ips[ip]; !ok && byte(ip) != 0 {
```
mike-sul, line 120: Do we allow `.1` device IP?

MrCry0: Which network mask do we use for VPN? Shouldn't we test the host part of the address, not just the last byte?

doanac: wireguard isn't putting devices on a subnet. they each just have a unique IP and route to one other in a mesh.

I was a little surprised, but .1 worked fine. It was just a .0 that tripped things up

MrCry0: Oh, it's a mesh. Thanks, it's interesting, need to read more about WG :)

doanac:
@Tim-Anderson - would you like to test a little more or would you rather me merge now?

doanac:
@Tim-Anderson - another thing we might point out to users - this is just how *fioctl* assigns IPs through an API. A customer that had more control over their devices' MAC address could probably do something completely deterministic. (i.e if the low 16 bits of the MAC are unique for their fleet - map that onto a 10.0.x.x/16 subnet).

PR #234, opened 2023-05-09
Feature: add verbosity and dry-run flags to wave rollout

vkhoroz:
This adds 3 options: --dry-run, --print-names, --print-uuids. Below is an example output with all 3 options enabled:
```
$ fioctl-linux-amd64 waves rollout test-wave --limit=2 --dry-run --print-uuids --print-names
A wave would be rolled out to 2 devices in your factory.
Below is a full list of these devices:
a4419ffb-b8ce-492c-8e50-4bce9a2398a1 (prod-dev1), 7a733e81-f948-43a9-a358-56f3deb5f184 (prod-dev2)
```

### DIFF
--- a/subcommands/waves/rollout.go
+++ b/subcommands/waves/rollout.go
@@ -74,6 +74,12 @@ A maximum number of devices rolled out using this flag cannot exceed 10000.`,
 Also accepts a filename containing a comma-separated list via "--uuids=@path/to/file.name".
 A maximum number of devices rolled out using this flag cannot exceed 10000.`,
 	)
+	rollout.Flags().BoolP("dry-run", "", false,
+		"Only show what would happen without an actual rollout. Most useful with --print-xxx flags.")
+	rollout.Flags().BoolP("print-uuids", "", false,
+		"Print UUIDs of devices to which a wave was rolled out (would be rolled out with --dry-run).")
+	rollout.Flags().BoolP("print-names", "", false,
+		"Print names of devices to which a wave was rolled out (would be rolled out with --dry-run).")
 	cmd.AddCommand(rollout)
 }
 
@@ -83,8 +89,13 @@ func doRolloutWave(cmd *cobra.Command, args []string) {
 	group := readGroup(cmd, args)
 	limit, percentage := readLimit(cmd)
 	uuids := readUuids(cmd)
+	dryRun, _ := cmd.Flags().GetBool("dry-run")
+	printUuids, _ := cmd.Flags().GetBool("print-uuids")
+	printNames, _ := cmd.Flags().GetBool("print-names")
 
-	if len(group) == 0 && len(uuids) == 0 && limit == 0 && percentage == 0 {
+	isFullRollout := len(uuids) == 0 && limit == 0 && percentage == 0
+
+	if len(group) == 0 && isFullRollout {
 		subcommands.DieNotNil(errors.New(
 			"One of the following flags must be set: group, limit, percentage, uuids\n" + cmd.UsageString(),
 		))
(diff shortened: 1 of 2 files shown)


subcommands/waves/rollout.go
```diff
@@ -74,6 +74,12 @@ A maximum number of devices rolled out using this flag cannot exceed 10000.`,
 Also accepts a filename containing a comma-separated list via "--uuids=@path/to/file.name".
 A maximum number of devices rolled out using this flag cannot exceed 10000.`,
 	)
+	rollout.Flags().BoolP("dry-run", "", false,
+		"Only show what would happen without an actual rollout. Most useful with --print-xxx flags.")
+	rollout.Flags().BoolP("print-uuids", "", false,
+		"Print the device UUIDs that were rolled out to (would be rolled out to with --dry-run).")
```
mike-sul, line 80: `device UUIDs that were rolled out` - it sounds a bit odd to me. I think these are UUIDs of devices that a wave update was rolled out to.
I defer it to @kprosise .

subcommands/waves/rollout.go
```diff
@@ -98,8 +109,52 @@ func doRolloutWave(cmd *cobra.Command, args []string) {
...
+	if (printUuids && len(res.DeviceUuids) == 0) || (printNames && len(res.DeviceNames) == 0) {
+		// This can happen when rolling out to a large group without any limit specified.
+		fmt.Println("A full list of these devices exceeds 10000, so it cannot be printed.")
+	} else if printUuids || printNames {
+		fmt.Println("Below is a full list of these devices:")
+		for i := 0; i < res.DeviceNum; i++ {
+			if i > 0 {
+				fmt.Print(", ")
+			}
+			if printUuids {
```
mike-sul, line 147: It is just nitpicking. Doing `if printUuids {...} else if printNames {...}` seems a redundant since  `printUuids` and `printNames` are constant within the loop.
I would do this `if else` before the loop and define a function variable in it, then just call the defined func var in the loop.

mike-sul (approved):
LGTM. Just two minor comments.

subcommands/waves/rollout.go
```diff
@@ -83,8 +89,13 @@ func doRolloutWave(cmd *cobra.Command, args []string) {
...
+	dryRun, _ := cmd.Flags().GetBool("dry-run")
+	printUuids, _ := cmd.Flags().GetBool("print-uuids")
+	printNames, _ := cmd.Flags().GetBool("print-names")
 
-	if len(group) == 0 && len(uuids) == 0 && limit == 0 && percentage == 0 {
+	isFullRollout := len(uuids) == 0 && limit == 0 && percentage == 0
+
+	if len(group) == 0 && isFullRollout {
 		subcommands.DieNotNil(errors.New(
 			"One of the following flags must be set: group, limit, percentage, uuids\n" + cmd.UsageString(),
```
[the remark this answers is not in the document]
vkhoroz: The explaination is already a part of the help message, which is BTW printed here.
There is no need to double-tell this to the user.

subcommands/waves/rollout.go
doanac: It seems like `--print-uuids` and `--print-names` aren't mutually exclusive? It might be handy this way - but then we need 3 pieces of logic for printing out the content.

subcommands/waves/rollout.go
vkhoroz: A full sentence is `device UUIDs that were rolled out to`... The last preposition is important ;) .

subcommands/waves/rollout.go
vkhoroz: yes, they are not exclusive and @doanac is right about the 3 pieces of logic.

mike-sul: It's not about optimization, this is about code readability.

vkhoroz: Finally, I _believe_ you propose to define 3 functions + if..else outside the loop, which would consume 15 lines of code (13 lines at the least with some hacking), in order to save 8 lines of code inside the loop.
I cannot imagine how it helps to simplify things.

subcommands/waves/rollout.go
```diff
@@ -98,8 +109,52 @@ func doRolloutWave(cmd *cobra.Command, args []string) {
...
+		fmt.Println("Below is a full list of these devices:")
+		for i := 0; i < res.DeviceNum; i++ {
+			if i > 0 {
+				fmt.Print(", ")
+			}
+			if printUuids {
+				fmt.Print(res.DeviceUuids[i])
+				if printNames {
+					fmt.Printf(" (%s)", res.DeviceNames[i])
+				}
```
doanac, line 151: I missed this earlier - you are handling all 3 cases. I thought about trying to simplify this logic but it wound up using `text/template` and defining a struct. So this is probably *almost* as short - and probably *much* more efficient.

subcommands/waves/rollout.go
vkhoroz: We roll out a wave to devices... at least that's the terminology we use in all help messages, including this file.  The word "[target]" gives no added context to this.

I'll defer this to @kprosise...

The full form would be:
`Print a list of UUIDs of devices to which a wave was rolled out (would be rolled out with --dry-run)`.
But I _believe_ that within a context of the full help message this can be truncated to the present sentence, and still be grammatically and contextually correct, while I find it neater.

subcommands/waves/rollout.go
mike-sul: It's not about optimizing a number of code lines, this is about code readability :).
Those `if` statements in the loop makes me think that the way info is output depends on a device number.

subcommands/waves/rollout.go
mike-sul: `"Print the device UUIDs that wave was rolled out to..."` looks fine to me. (just three more symbols :)).

subcommands/waves/rollout.go
vkhoroz: Yes, we basically have 3 options:
1. A loop with 3 ifs.
2. 3 ifs, each with its own loop.
3. A loop with a static template containing 3 ifs.

What @mike-sul suggests above is practically equivalent to the template implemented as a set of functions (i.e. option 3).

All 3 choices are technically equivalent, but as per my personal taste the first option is more readable.
From the speed perspective 2 is probably marginally faster than 1, yet both win option 3 by a decent margin.

subcommands/waves/rollout.go
vkhoroz: A code readability is dependent onto a personal taste. As per my taste a loop with 3 ifs is more readable than declaring 3 functions.

So there should be a stronger argument. Speed of execution could be that argument, but then we'd end up with 3 if statements, each containing its own loop. That'd make the code less readable IMHO.

subcommands/waves/rollout.go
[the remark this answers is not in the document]
mike-sul: 1. The explanation how to use is printed, not a reason of the command failure. Instead it says "error: do something".

It looks like I have to replace all error messages in aklite. So, instead of "err: invalid DG url" aklite should print "err: set proper DG url is sota.toml". :).

subcommands/waves/rollout.go
mike-sul: >A code readability is dependent onto a personal taste.

It's not, there are best practices.

subcommands/waves/rollout.go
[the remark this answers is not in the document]
vkhoroz: fioctl is not aklite - they can use different semantics.

Here are some of the fioctl errors generated by cobra for this use case:
- `Error: required flag(s) "role" not set`;
- `Error: if any flags in the group [first-time keys] are set they must all be set; missing [keys]`
- `Error: if any flags in the group [raw diff] are set none of the others can be; [diff raw] were all set`
- `ERROR: One of the following flags must be set: group, limit, percentage, uuids`

The last one is from this PR.  An error here is that the user did not specify one of the required flags. The most I can do to keep it the same look and feel is to add `; none was set` at the end, although I find it redundant.

Also, maybe, we need to change `ERROR:` to `Error:` everywhere; but this is a separate topic.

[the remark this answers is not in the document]
vkhoroz: I'm not going to change this. This message goes from the previous PR and is good enough.

doanac:
I think it looks good. Would be nice to get @kprosise  input on messages though.

PR #236, opened 2023-06-12
feat(device): Managing device config

StealthyCoder:
A way to manage config on devices to set apps and tags to either:

- Inherit apps from parent
- Give special character `,` to select all known apps in tag
- Give `""` to not run any apps at all

- Inherit tag from parent
- Give either special character `,` or `""` to empty the tag

### DIFF
--- a/subcommands/common_config.go
+++ b/subcommands/common_config.go
@@ -7,12 +7,14 @@ import (
 	"io"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/fatih/color"
 	toml "github.com/pelletier/go-toml"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/exp/slices"
 
 	"github.com/foundriesio/fioctl/client"
 )
@@ -176,16 +178,20 @@ func (o RotateCertOptions) AsConfig() client.ConfigCreateRequest {
 }
 
 type SetUpdatesConfigOptions struct {
-	UpdateTag  string
-	UpdateApps string
-	IsDryRun   bool
-	IsForced   bool
-	Device     *client.Device
-	ListFunc   func() (*client.DeviceConfigList, error)
-	SetFunc    func(client.ConfigCreateRequest, bool) error
+	UpdateTag       string
+	UpdateTagGiven  bool
+	InheritTag      bool
+	UpdateApps      string
+	UpdateAppsGiven bool
+	InheritApps     bool
+	IsDryRun        bool
+	IsForced        bool
+	Device          *client.Device
+	ListFunc        func() (*client.DeviceConfigList, error)
+	SetFunc         func(client.ConfigCreateRequest, bool) error
 }
 
-func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reportedApps []string) {
+func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reportedApps []string, factory string, api *client.Api) {
 	DieNotNil(validateUpdateArgs(opts))
 
 	dcl, err := opts.ListFunc()
@@ -196,8 +202,7 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
 	if !opts.IsForced {
 		DieNotNil(err, "Invalid FIO toml file (override with --force):")
 	}
-
-	if opts.UpdateApps == "" && opts.UpdateTag == "" {
+	if opts.UpdateApps == "" && opts.UpdateTag == "" && !opts.InheritApps && !opts.InheritTag {
 		if opts.Device != nil {
 			fmt.Println("= Reporting to server with")
 			fmt.Println(" Tag: ", opts.Device.Tag)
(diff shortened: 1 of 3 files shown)


subcommands/devices/config_updates.go
```diff
@@ -32,7 +32,9 @@ currently configured and reporting.`,
 	configCmd.AddCommand(configUpdatesCmd)
 	configUpdatesCmd.Flags().StringP("tag", "", "", "Target tag for device to follow")
 	configUpdatesCmd.Flags().StringP("tags", "", "", "Target tag for device to follow")
+	configUpdatesCmd.Flags().BoolP("inherit-tag", "", false, "Toggle if tag should be inherited from parent")
 	configUpdatesCmd.Flags().StringP("apps", "", "", "comma,separate,list")
+	configUpdatesCmd.Flags().BoolP("inherit-apps", "", false, "Toggle if apps should be inherited from parent")
```
doanac, line 37: I'm not sure parent is the correct word for this. Also - should `apps` and `inherit-apps` be mutually exclusive? Same thing with tags as well.

subcommands/common_config.go
```diff
@@ -220,22 +225,62 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
...
+						}
+						if v, err := strconv.Atoi(custom.Version); err == nil {
+							if v > version {
+								final_apps = apps
+								version = v
+							}
+						}
+					}
+				}
+				opts.UpdateApps = strings.Join(final_apps, ",")
```
vkhoroz, line 263: IIUC this is a huge change in behavior.

1. In the past `--apps=","` was used to tell the device to run no apps, and this PR changes it to the opposite: run app apps.

2. Furthermore, it does that in an IMHO wrong way:
- a true way to tell to run all known apps is to set apps to an empty string, this way the device will always run all applications known at a time of the update.
- with this change, the device will be stick to run the apps known at a time of config change; when the list of apps change with the future updates - the device would not update to a new apps list automatically.

So, I _believe_ that:
- for 1 we must revert the original behavior to be backward compatible, i.e. `--apps=,` should result into "no apps".
- for 2 we should find a way to set the `pacman.docker_apps=""` and verify if it overrides the upper level `sota.toml` values properly. If not, maybe your idea to set the device to all _currently known_ apps is OK, but we must use a different _special value_ than a comma for this; maybe even a separate flag like `--all-apps`.

subcommands/common_config.go
```diff
@@ -176,16 +178,20 @@ func (o RotateCertOptions) AsConfig() client.ConfigCreateRequest {
...
-	IsForced   bool
-	Device     *client.Device
-	ListFunc   func() (*client.DeviceConfigList, error)
-	SetFunc    func(client.ConfigCreateRequest, bool) error
+	UpdateTag       string
+	UpdateTagGiven  bool
+	InheritTag      bool
+	UpdateApps      string
+	UpdateAppsGiven bool
+	InheritApps     bool
```
vkhoroz, line 186: I'm drawn in all these flags. Although at a UI level it makes sense to have separate flags for this (for the user convenience), I think that for `SetUpdatesConfigOptions` flags we can do it more neat.

For example, an easy way to eliminate a necessity for `UpdateAppsGiven` flag is to use a `*string` type for `UpdateApps`. The code would be roughly the same, but the amount of code friction is smaller.

subcommands/devices/config_updates.go
```diff
@@ -57,17 +63,21 @@ func doConfigUpdates(cmd *cobra.Command, args []string) {
...
-		IsForced:   isForced,
-		Device:     device,
+		UpdateApps:      updateApps,
+		UpdateAppsGiven: updateAppsGiven,
+		InheritApps:     inheritApps,
+		UpdateTag:       updateTag,
+		UpdateTagGiven:  updateTagGiven,
+		InheritTag:      inheritTag,
+		IsDryRun:        isDryRun,
+		IsForced:        isForced,
```
vkhoroz, line 73: This boilerplate (as well as option parsing above) is _exactly the same_ in all 3 config handlers.

I think, as the amount of options (and underlying logic) increased quite some, it makes sence to make a common function like `ParseConfigOptions(cmd *cobra.Command) SetUpdatesConfigOptions`, which parses these _known_ options in a uniform way.

subcommands/common_config.go
vkhoroz: @StealthyCoder for now, please, ignore my comment before @mike-sul gives a definite answer.
I might make some _too early_ conclusions.

StealthyCoder: @vkhoroz 
> But, what will happen if sota.toml has pacman.compose_apps=,? Will it run "no apps" or "all apps"?

As far as I can tell through testing, it does the run no apps. Same for "" or " ", but I have not tried leaving it open. `pacman.compose_apps=` . So I will try that. 

I am not sure how to send that through via Golang but that is something different.

subcommands/devices/config_updates.go
StealthyCoder: Gotcha, I will try and implement that.

subcommands/common_config.go
```diff
@@ -220,22 +225,62 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
...
-			opts.UpdateTag = ""
+	if opts.UpdateTagGiven || opts.InheritTag {
+		if opts.InheritTag && sota.Has("pacman.tags") {
+			fmt.Printf("Changing tag from: [%s] -> inherit from parent\n", configuredTag)
+			DieNotNil(sota.Delete("pacman.tags"))
+			changed = true
+		} else if configuredTag != opts.UpdateTag {
+			if strings.TrimSpace(opts.UpdateTag) == "," {
+				opts.UpdateTag = ""
+			}
```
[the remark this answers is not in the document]
vkhoroz: > I just copied this logic from existing implementation. I can get rid of it through this PR no worries.

After I wrote a comment, I had a feeling that you copied it _as is_ and I understood it wrong.
This is not your fault, it is the complexity of _all variants we need to support_ that drives seeds the confusion.

subcommands/common_config.go
vkhoroz: As per the offline email thread, let's just remove this logic to select "all apps".
It is a new "potential" feature, which is rather ambiguous, as some users would expect it to include "all future apps".

The previous behavior for "," value was indeed to set "no apps".
So, let's just preserve this behavior _as is_.

An option to introduce an explicit `--no-apps` flag (and prohibit a "," here) looks appealing, but it would require too much effort to update docs/blogs/etc.

subcommands/common_config.go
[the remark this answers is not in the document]
vkhoroz: @doanac do we want to continue supporting the "unset tag" option?
I _believe_ we don't want to allow devices with no tag.

@StealthyCoder if Andy gives a green light to this, please, make it as a separate commit in this PR, explaining the rationale.

subcommands/common_config.go
vkhoroz: @doanac I _guess_ you should agree to this _being conservative_ proposal.

subcommands/common_config.go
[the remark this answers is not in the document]
doanac: Its safe to "unset" because a `sota.toml` will have a tag - lmp-device-register forces that. This could be handy for something like:
* A device in a device-group is misbehaving
* Move device to a "debug" type tag
* After debugging, "unset" to get it back to device-group.

PR #179, opened 2022-09-01
Add el2g device management commands

doanac:

PR #238, opened 2023-06-20
Support multi-key TUF root in key rotation and signing commands

vkhoroz:

### DIFF
--- a/subcommands/keys/tuf_updates_rotate_offline_key.go
+++ b/subcommands/keys/tuf_updates_rotate_offline_key.go
@@ -199,10 +199,14 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
 func replaceOfflineRootKey(
 	root *client.AtsTufRoot, creds OfflineCreds, keyType TufKeyType,
 ) (*TufSigner, OfflineCreds) {
+	oldKey, err := FindOneTufSigner(root, creds, root.Signed.Roles["root"].KeyIDs)
+	subcommands.DieNotNil(err)
+	newKids := subcommands.SliceRemove(root.Signed.Roles["root"].KeyIDs, oldKey.Id)
+
 	kp := genTufKeyPair(keyType)
 	root.Signed.Keys[kp.signer.Id] = kp.atsPub
 	root.Signed.Expires = time.Now().AddDate(1, 0, 0).UTC().Round(time.Second) // 1 year validity
-	root.Signed.Roles["root"].KeyIDs = []string{kp.signer.Id}
+	root.Signed.Roles["root"].KeyIDs = append(newKids, kp.signer.Id)
 
 	base := "tufrepo/keys/fioctl-root-" + kp.signer.Id
 	creds[base+".pub"] = kp.atsPubBytes
@@ -213,9 +217,14 @@ func replaceOfflineRootKey(
 func replaceOfflineTargetsKey(
 	root *client.AtsTufRoot, onlineTargetsId string, creds OfflineCreds, keyType TufKeyType,
 ) (*TufSigner, OfflineCreds) {
+	oldKey, err := FindOneTufSigner(root, creds,
+		subcommands.SliceRemove(root.Signed.Roles["targets"].KeyIDs, onlineTargetsId))
+	subcommands.DieNotNil(err)
+	newKids := subcommands.SliceRemove(root.Signed.Roles["targets"].KeyIDs, oldKey.Id)
+
 	kp := genTufKeyPair(keyType)
 	root.Signed.Keys[kp.signer.Id] = kp.atsPub
-	root.Signed.Roles["targets"].KeyIDs = []string{onlineTargetsId, kp.signer.Id}
+	root.Signed.Roles["targets"].KeyIDs = append(newKids, kp.signer.Id)
 	root.Signed.Roles["targets"].Threshold = 1
 
 	base := "tufrepo/keys/fioctl-targets-" + kp.signer.Id
--- a/subcommands/keys/tuf_updates_rotate_offline_key.go
+++ b/subcommands/keys/tuf_updates_rotate_offline_key.go
@@ -138,10 +138,7 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
 		subcommands.DieNotNil(errors.New("The --keys option is required to sign the new TUF root."))
 	}
 
-	var (
-		creds, targetsCreds OfflineCreds
-		err                 error
-	)
+	var creds, targetsCreds OfflineCreds
 	if _, err := os.Stat(targetsKeysFile); err == nil {
 		targetsCreds, err = GetOfflineCreds(targetsKeysFile)
 		subcommands.DieNotNil(err)
@@ -157,6 +154,7 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
 		if keysFile == targetsKeysFile {
 			creds = targetsCreds
 		} else {
+			var err error
 			creds, err = GetOfflineCreds(keysFile)
 			subcommands.DieNotNil(err)
 		}
@@ -185,7 +183,7 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
 	newProdRoot := genProdTufRoot(newCiRoot)
 
 	fmt.Println("= Re-signing prod targets")
-	newTargetsSigs, err := resignProdTargets(factory, newCiRoot, onlineTargetsId, newCreds)
+	newTargetsSigs, err := resignProdTargets(factory, newKey)
 	subcommands.DieNotNil(err)
 
 	if shouldSign {
@@ -226,27 +224,12 @@ func replaceOfflineTargetsKey(
 	return &kp.signer, creds
 }
 
-func resignProdTargets(
-	factory string, root *client.AtsTufRoot, onlineTargetsId string, creds OfflineCreds,
-) (map[string][]tuf.Signature, error) {
+func resignProdTargets(factory string, signer *TufSigner) (map[string][]tuf.Signature, error) {
 	targetsMap, err := api.ProdTargetsList(factory, false)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to fetch production targets: %w", err)
 	} else if targetsMap == nil {
 		return nil, nil
-	}
-
-	var signers []TufSigner
-	for _, kid := range root.Signed.Roles["targets"].KeyIDs {
-		if kid == onlineTargetsId {
-			continue
-		}
-		pub := root.Signed.Keys[kid].KeyValue.Public
-		signer, err := FindTufSigner(kid, pub, creds)
-		if err != nil {
-			return nil, fmt.Errorf("Failed to find private key for %s: %w", kid, err)
-		}
-		signers = append(signers, *signer)
 	}
 
 	signatureMap := make(map[string][]tuf.Signature)
@@ -255,7 +238,7 @@ func resignProdTargets(
 		if err != nil {
 			return nil, fmt.Errorf("Failed to marshal targets for tag %s: %w", tag, err)
 		}
-		signatures, err := SignTufMeta(bytes, signers...)
+		signatures, err := SignTufMeta(bytes, *signer)
 		if err != nil {
 			return nil, fmt.Errorf("Failed to re-sign targets for tag %s: %w", tag, err)
 		}
--- a/subcommands/keys/tuf_updates_rotate_offline_key.go
+++ b/subcommands/keys/tuf_updates_rotate_offline_key.go
@@ -198,7 +198,7 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
 
 func replaceOfflineRootKey(
 	root *client.AtsTufRoot, creds OfflineCreds, keyType TufKeyType,
-) (*TufSigner, OfflineCreds) {
+) (TufSigner, OfflineCreds) {
 	oldKey, err := FindOneTufSigner(root, creds, root.Signed.Roles["root"].KeyIDs)
 	subcommands.DieNotNil(err)
 	newKids := subcommands.SliceRemove(root.Signed.Roles["root"].KeyIDs, oldKey.Id)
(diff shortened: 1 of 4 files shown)


vkhoroz:
@doanac @mike-sul @StealthyCoder this is a draft change to support multi-key rotation.

I am extensively testing these changes, while you have a chance to take a quick look at them.

vkhoroz:
One pending commit is an optimization:
- currently, some commands can crawl through the creds file several times (3 times at most).
- ideally, that needs to be done just once, but I'm not sure if that wouldn't be a pre-mature optimization.

So, if I encounter during testing that for a file with e.g. 50 keys, there is an unreasonable slowdown - I'll consider doing it.
Otherwise, there is no value in that effort.

vkhoroz:
@doanac @mike-sul I finished my tests.

These changes behave exactly the same with the current API.
An extension to support multi-key TUF roles will be tested when I implement ability to add more keys, in subsequent PRs.

From the performance perspective, after 100 root & targets signing key rotations (each), I found no significant correlation between the size of an archive and the rotation command time.
In all cases between 0 and 100 keys, both commands took 1-2 seconds, regardless of the file size.
Thus, there is no need to implement any speed optimizations for the command.

subcommands/keys/tuf_utils.go
```diff
@@ -212,44 +212,81 @@ func GetOfflineCreds(credsFile string) (OfflineCreds, error) {
...
-				}, nil
-			}
+func FindOneTufSigner(root *client.AtsTufRoot, creds OfflineCreds, keyids []string) (signer TufSigner, err error) {
+	var signers []TufSigner
+	if signers, err = findTufSigners(root, creds, keyids); err == nil {
+		if len(signers) == 0 {
+			err = fmt.Errorf("Found no active signing key for: %v.", keyids)
+		} else if len(signers) > 1 {
+			err = fmt.Errorf(`Found more than one active signing key for: %v.
+This is an unsupported and insecure way to store private keys.
```
doanac, line 222: This is a functional change in behavior, right? not against it, but probably need to mention this in the next release notes.

vkhoroz: An answer to your question is a bit more complex:
- From the current implementation perspective, this is a forward-looking non-functional change. We only allow to have 1 active offline TUF root key and 1 active offline TUF targets key at a time. Thus, there is no way this condition can be true.
- At the same time, this is a change defines a future function of the multi-key TUF support, in accordance to our agreed design.

So, I'd say that this is a regular _new feature_, which is fully backward-compatible.
Therefore it needs not be mentioned explicitly in release notes for the next fioctl version, other than a part of a sentence about the _multi-key TUF support_.

doanac (approved):

subcommands/keys/tuf_updates_rotate_offline_key.go
```diff
@@ -185,7 +183,7 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
 	newProdRoot := genProdTufRoot(newCiRoot)
 
 	fmt.Println("= Re-signing prod targets")
-	newTargetsSigs, err := resignProdTargets(factory, newCiRoot, onlineTargetsId, newCreds)
+	newTargetsSigs, err := resignProdTargets(factory, newKey)
```
mike-sul, line 186: Will anything bad happen if the `fioctl` process is killed/interrupted at the moment between #179 (key replaced and stored) and #186?
I suppose no, a user will need to run fioctl command again to resign targets with the new targets key?

I wonder whether it's not an issue for the root key rotation when two signatures are required (previous and new root role key signatures).

mike-sul: `newKey` is a bit confusing name since it has `*TufSigner` type.

subcommands/keys/tuf_updates_rotate_offline_key.go
```diff
@@ -226,36 +224,21 @@ func replaceOfflineTargetsKey(
...
-	}
-
 	signatureMap := make(map[string][]tuf.Signature)
 	for tag, targets := range targetsMap {
 		bytes, err := canonical.MarshalCanonical(targets.Signed)
 		if err != nil {
 			return nil, fmt.Errorf("Failed to marshal targets for tag %s: %w", tag, err)
 		}
-		signatures, err := SignTufMeta(bytes, signers...)
+		signatures, err := SignTufMeta(bytes, *signer)
```
mike-sul, line 250: Are we sure that `signer` is not nil at this point?

subcommands/keys/tuf_utils.go
```diff
@@ -212,44 +212,81 @@ func GetOfflineCreds(credsFile string) (OfflineCreds, error) {
...
-func findTufRootSigner(root *client.AtsTufRoot, creds OfflineCreds) (*TufSigner, error) {
-	kid := root.Signed.Roles["root"].KeyIDs[0]
-	pub := root.Signed.Keys[kid].KeyValue.Public
-	return FindTufSigner(kid, pub, creds)
+func findTufSigners(root *client.AtsTufRoot, creds OfflineCreds, keyids []string) ([]TufSigner, error) {
+	// Look in creds for each candidate from keyids and return all private keys that match
+	matchPubKeys := make(map[string]client.AtsKey, len(keyids))
+	for _, kid := range keyids {
+		if pk, ok := root.Signed.Keys[kid]; ok {
+			pk.KeyValue.Public = strings.TrimSpace(pk.KeyValue.Public)
```
mike-sul, line 236: What this trimming is need for? Does it mean that we store pubkey not quite correctly?

subcommands/keys/tuf_utils.go
```diff
@@ -212,44 +212,81 @@ func GetOfflineCreds(credsFile string) (OfflineCreds, error) {
...
 	}
-	return nil, fmt.Errorf("Can not find private key for: %s", keyid)
+	return
 }
 
-func findTufRootSigner(root *client.AtsTufRoot, creds OfflineCreds) (*TufSigner, error) {
-	kid := root.Signed.Roles["root"].KeyIDs[0]
-	pub := root.Signed.Keys[kid].KeyValue.Public
-	return FindTufSigner(kid, pub, creds)
+func findTufSigners(root *client.AtsTufRoot, creds OfflineCreds, keyids []string) ([]TufSigner, error) {
```
mike-sul, line 231: At first glance it looks like passing both `client.AtsTufRoot` and `keyids` and then checking if all keys in `keyids` are present in the root is redundant taking into account how `findTufSigners` is used.

`FindOneTufSigner(curCiRoot, creds, curCiRoot.Signed.Roles["root"].KeyIDs) -> findTufSigners(root, creds, keyids)`.  Effectively, `keyids` refers to the `root`.

Maybe `findTufSigners` is used in different way in some other code places.

subcommands/keys/tuf_updates_rotate_offline_key.go
```diff
@@ -199,10 +199,14 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
 func replaceOfflineRootKey(
 	root *client.AtsTufRoot, creds OfflineCreds, keyType TufKeyType,
 ) (*TufSigner, OfflineCreds) {
+	oldKey, err := FindOneTufSigner(root, creds, root.Signed.Roles["root"].KeyIDs)
```
mike-sul, line 202: So, we are replacing the first found key?

subcommands/keys/tuf_updates_rotate_offline_key.go
```diff
@@ -213,9 +217,14 @@ func replaceOfflineRootKey(
 func replaceOfflineTargetsKey(
 	root *client.AtsTufRoot, onlineTargetsId string, creds OfflineCreds, keyType TufKeyType,
 ) (*TufSigner, OfflineCreds) {
+	oldKey, err := FindOneTufSigner(root, creds,
+		subcommands.SliceRemove(root.Signed.Roles["targets"].KeyIDs, onlineTargetsId))
```
mike-sul, line 221: Why are we replacing the online key? We may wanna rotate the exiting offline targets key too?

subcommands/keys/tuf_updates_rotate_offline_key.go
vkhoroz: > Will anything bad happen if the fioctl process is killed/interrupted ...

There is no chance anything bad can happen until the user runs `fioctl keys tuf apply` command. All these actions are performed upon a temporal metadata, which does not affect devices. So, even if some unknown unknown bug is burrowed here, it only affects that temporal metadata. In any case, at the apply stage, that temporal data is verified for integrity before the commit.

> newKey is a bit confusing name since it has *TufSigner type.

This is how we are used to call this thing in other places like this.
In general, I care little about the correlation between a variable name and a type name. In essence, this variable holds a reference to the new key. A fact that it is wrapped into a convenient intermediary type is secondary here.

subcommands/keys/tuf_utils.go
vkhoroz: I don't remember the exact details... IIRC in the past I've seen that public key to be prepended by a newline or smth like that in some edge cases. Note: these key files are not required to be in a canonical form, although all our code tries to generate canonical JSONs. In essence, an RSA key in PEM format with and without a white space is exactly the same. In theory, it can be even split into lines in any way; so, ideally we should split by white space and join without it to compare the keys. But, luckily I never observed a difference in presentations to go that far.

subcommands/keys/tuf_utils.go
vkhoroz: It is redundant in some cases, and not redundant in other cases.
For example, for TUF targets we need to exclude the online signing key from the list.

subcommands/keys/tuf_updates_rotate_offline_key.go
vkhoroz: We are replacing one and the only found key.
In case if there are more than 1 matching keys - we error out.

Please, see the reference design for a reasoning why it is not secure to keep two or more actual signing keys in one file.

subcommands/keys/tuf_updates_rotate_offline_key.go
vkhoroz: I exclude an online key from a list of keys to search for.
This warrants that the online key is never replaced, even if a user manually inserted it into the keys file.

subcommands/keys/tuf_updates_rotate_offline_key.go
vkhoroz: Thanks for the comment.

I need one more commit which makes `replaceOfflineRootKey` and `replaceOfflineTargetsKey` to return value types, so that I can make `resignProdTargets` accept a value type too.

This value can never be `nil`, so there is no reason to make it a reference type.

TL;DR
IIRC this is a historic baggage from the times some of us believed that having a reference makes the program faster because of not having to do the copy-on-passing-argument for the value types. It turns out the effect is quite opposite, as Golang leverages CPU registers a lot to pass small value types, making that faster. So, we unintentionally made some places of our code nil-poisoned while striving to achieve a nonexisting benefit.

subcommands/keys/tuf_updates_rotate_offline_key.go
mike-sul: >Please, see the reference design for a reasoning why it is not secure to keep two or more actual signing keys in one file.

That's not related to my question.

If we have more than one signing key (doesn't matter if they are in a file, files or DB) how we know that we are replacing the key that customers would like to rotate.

mike-sul: I think `findTufSigners` searches through many files, so it can return many "keys/signers", and `FindOneTufSigner` just choose the first one.

mike-sul: https://github.com/foundriesio/fioctl/blob/6a0a259b3d8010d0b5450a8020ab7719e292a7ed/subcommands/keys/tuf_utils.go#L225

subcommands/keys/tuf_updates_rotate_offline_key.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/238/commits/6c1e20c1712a2ce8d4b640cdd48c07387378531a

subcommands/keys/tuf_utils.go
mike-sul: ok

subcommands/keys/tuf_updates_rotate_offline_key.go
vkhoroz: The `FindOneTufSigner` makes sure that `signers` list contains exactly one item before returning the first one.
Please, re-read its implementation.

subcommands/keys/tuf_updates_rotate_offline_key.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/238/commits/6c1e20c1712a2ce8d4b640cdd48c07387378531a

subcommands/keys/tuf_utils.go
mike-sul: My point is that if we store the pub key, then we can store it correctly so the code that reads do not need to do the trimming. Otherwise, it looks like we don't trust ourself.

Anyway, I think the trimming doesn't do any bad.

subcommands/keys/tuf_updates_rotate_offline_key.go
mike-sul: I though that the key replaced and stored in a file at #179, but it's not so.

vkhoroz: A new key is stored in a file, correct.
But an old key is _not_ removed from it - it remains in a file for the case if something goes wrong or the user changes their mind.

mike-sul: ok, early you said that two keys in a single file is not good :).

mike-sul: I got it now. thx.

subcommands/keys/tuf_updates_rotate_offline_key.go
mike-sul: OK, so `creds` that contains a key file data identifies the key to be replaced.

mike-sul (approved):

PR #239, opened 2023-06-23
Meds updates

doanac:
These are the changes that were required to allow fioctl to fully integrate with a MEDs instance.

### DIFF
--- a/client/oauth.go
+++ b/client/oauth.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -28,7 +29,8 @@ type OAuthConfig struct {
 }
 
 type ClientCredentials struct {
-	Config OAuthConfig
+	Config      OAuthConfig
+	InsecureSSL bool
 }
 
 type Org struct {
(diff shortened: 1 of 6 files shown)


subcommands/common.go
```diff
@@ -55,6 +52,10 @@ func Login(cmd *cobra.Command) *client.Api {
 		DieNotNil(fmt.Errorf("Required flag \"factory\" not set"))
 	}
 	creds := client.NewClientCredentials(Config.ClientCredentials)
+	if viper.GetBool("server.insecure_skip_verify") {
+		Config.InsecureSkipVerify = true
+		creds.InsecureSSL = true
+	}
```
vkhoroz, line 58: Can you put this block into the desired place in the [original commit](https://github.com/foundriesio/fioctl/pull/239/commits/f93f30446e9593b928410c69f813973f96551072), to make the diff cleaner?

client/oauth.go
```diff
@@ -71,7 +73,14 @@ func (c *ClientCredentials) updateConfig(r OAuthResponse) {
...
 // Perform a POST request.
 func (c *ClientCredentials) post(uri string, data url.Values) (*[]byte, error) {
-	res, err := http.PostForm(uri, data)
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: c.InsecureSSL,
+			},
+		},
+	}
```
vkhoroz, line 82: This looks like a continuation of [this commit](https://github.com/foundriesio/fioctl/pull/239/commits/f93f30446e9593b928410c69f813973f96551072).

I _believe_ there is some common part here that can be generalized in a nice way; so that we don't have to copy it over and over again.

vkhoroz (commented):
Looks good.

subcommands/common.go
doanac: fixed in force-push - you actually uncovered a bug i introduced here.

client/oauth.go
vkhoroz: If unification creates a problem for you; we're fine as is. We only have 2 places of a code like this.

vkhoroz (approved):

PR #241, opened 2023-06-23
Feature: ability to add more than one offline TUF signing key

vkhoroz:

### DIFF
--- a/subcommands/keys/tuf_updates_add_offline_key.go
+++ b/subcommands/keys/tuf_updates_add_offline_key.go
@@ -1,0 +1,122 @@
+package keys
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	tuf "github.com/theupdateframework/notary/tuf/data"
+
+	"github.com/foundriesio/fioctl/client"
+	"github.com/foundriesio/fioctl/subcommands"
+)
+
+func init() {
+	add := &cobra.Command{
+		Use:   "add-offline-key --role root|targets --txid=<txid> --keys=<tuf-root-keys.tgz>",
+		Short: "Stage addition of the offline TUF signing key for the Factory",
+		Long: `Stage addition of the offline TUF signing key for the Factory.
+
+The new offline signing key will be used in both CI and production TUF root.
+
+When you add a new TUF targets offline signing key, existing production targets are not signed by it.
+Please, use the sign-prod-targets subcommand if you want to sign existing production targets with a new key.`,
+		Example: `
+- Add offline TUF root key:
+  fioctl keys tuf updates add-offline-key \
+	 --txid=abc --role=root --keys=tuf-root-keys.tgz
+- Add offline TUF targets key, explicitly specifying new key type (and signing algorithm):
+  fioctl keys tuf updates add-offline-key \
+    --txid=abc --role=targets --keys=tuf-targets-keys.tgz --key-type=ed25519`,
+		Run: doTufUpdatesAddOfflineKey,
+	}
+	add.Flags().StringP("role", "r", "", "TUF role name, supported: Root, Targets.")
+	_ = add.MarkFlagRequired("role")
+	add.Flags().StringP("txid", "x", "", "TUF root updates transaction ID.")
+	add.Flags().StringP("keys", "k", "", `Path to <tuf-keys.tgz> where a new key should be created.
+For security reasons, it is disallowed to add several actual keys for the same TUF role into the same file.`)
+	_ = add.MarkFlagFilename("keys")
+	_ = add.MarkFlagRequired("keys")
+	add.Flags().StringP("key-type", "y", tufKeyTypeNameEd25519, "Key type, supported: Ed25519, RSA.")
+	tufUpdatesCmd.AddCommand(add)
+}
+
+func doTufUpdatesAddOfflineKey(cmd *cobra.Command, args []string) {
+	factory := viper.GetString("factory")
+	txid, _ := cmd.Flags().GetString("txid")
+	keyTypeStr, _ := cmd.Flags().GetString("key-type")
+	keyType := ParseTufKeyType(keyTypeStr)
+	keysFile, _ := cmd.Flags().GetString("keys")
+	roleName, _ := cmd.Flags().GetString("role")
+	roleName = ParseTufRoleNameOffline(roleName)
+
+	var creds OfflineCreds
+	if _, err := os.Stat(keysFile); err == nil {
+		creds, err = GetOfflineCreds(keysFile)
+		subcommands.DieNotNil(err)
+		subcommands.AssertWritable(keysFile)
+	} else if errors.Is(err, fs.ErrNotExist) {
+		creds = make(OfflineCreds, 0)
+		saveTufCreds(keysFile, creds)
+	} else {
+		subcommands.DieNotNil(err)
+	}
+
+	updates, err := api.TufRootUpdatesGet(factory)
+	subcommands.DieNotNil(err)
+
+	_, newCiRoot := checkTufRootUpdatesStatus(updates, true)
+
+	switch roleName {
+	case tufRoleNameRoot:
+		addOfflineRootKey(newCiRoot, creds, keyType)
+	case tufRoleNameTargets:
+		onlineTargetsId := updates.Updated.OnlineKeys["targets"]
+		if onlineTargetsId == "" {
+			subcommands.DieNotNil(errors.New("Unable to find online target key for factory"))
+		}
+		addOfflineTargetsKey(newCiRoot, creds, keyType, onlineTargetsId)
+	default:
+		panic(fmt.Errorf("Unexpected role name: %s", roleName))
+	}
+
+	newCiRoot.Signatures = make([]tuf.Signature, 0)
+	removeUnusedTufKeys(newCiRoot)
+	newProdRoot := genProdTufRoot(newCiRoot)
+
+	fmt.Println("= Uploading new TUF root")
+	tmpFile := saveTempTufCreds(keysFile, creds)
+	err = api.TufRootUpdatesPut(factory, txid, newCiRoot, newProdRoot, nil)
+	handleTufRootUpdatesUpload(tmpFile, keysFile, err)
+}
+
+func addOfflineRootKey(root *client.AtsTufRoot, creds OfflineCreds, keyType TufKeyType) {
+	subcommands.DieNotNil(checkNoTufSigner(root, creds, root.Signed.Roles["root"].KeyIDs))
+
+	kp := genTufKeyPair(keyType)
+	root.Signed.Keys[kp.signer.Id] = kp.atsPub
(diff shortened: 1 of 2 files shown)


subcommands/keys/tuf_updates_add_offline_key.go
```diff
@@ -0,0 +1,124 @@
...
+func init() {
+	add := &cobra.Command{
+		Use:   "add-offline-key --role root|targets --txid=<txid> --keys=<tuf-root-keys.tgz>",
+		Short: "Stage addition of the offline TUF signing key for the Factory",
+		Long: `Stage addition of the offline TUF signing key for the Factory.
+
+The new offline signing key will be used in both CI and production TUF root.
+
+When you add a new TUF targets offline signing key, existing production targets are not signed by it.
+Please, use the sign-prod-targets subcommand if you want to sign existing production targets with a new key.`,
```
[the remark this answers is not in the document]
mike-sul: >So, together these problems necessitate to add a separate command to sign prod targets.

makes sense to me.

subcommands/keys/tuf_updates_add_offline_key.go
```diff
@@ -0,0 +1,124 @@
...
+		subcommands.SliceRemove(root.Signed.Roles["targets"].KeyIDs, onlineTargetsId)))
+
+	kp := genTufKeyPair(keyType)
+	root.Signed.Keys[kp.signer.Id] = kp.atsPub
+	root.Signed.Roles["targets"].KeyIDs = append(root.Signed.Roles["targets"].KeyIDs, kp.signer.Id)
+
+	base := "tufrepo/keys/fioctl-targets-" + kp.signer.Id
+	creds[base+".pub"] = kp.atsPubBytes
+	creds[base+".sec"] = kp.atsPrivBytes
+	fmt.Println("= New targets keyid:", kp.signer.Id)
```
doanac, line 123: these two functions look a lot alike. I'm not against duplication if its easier or seemed cleaner to you. Just double checking.

vkhoroz: I copied a pattern from `rotateXXX` functions here.

At one point I was thinking about unification; but there are some tiny nuances that could become too vague during that process.  That's why I decided to do this _very explicitly_ and _maybe revisit later_.

subcommands/keys/tuf_updates_add_offline_key.go
```diff
@@ -0,0 +1,124 @@
...
+	case tufRoleNameRoot:
+		addOfflineRootKey(newCiRoot, creds, keyType)
+	case tufRoleNameTargets:
+		onlineTargetsId := updates.Updated.OnlineKeys["targets"]
+		if onlineTargetsId == "" {
+			subcommands.DieNotNil(errors.New("Unable to find online target key for factory"))
+		}
+		addOfflineTargetsKey(newCiRoot, creds, keyType, onlineTargetsId)
+	default:
+		panic(fmt.Errorf("Unexpected role name: %s", roleName))
```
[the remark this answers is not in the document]
vkhoroz: This is a copy-pasted pattern from `rotate` function as well. A `ParseTufRoleNameOffline` does not allow a wrong value here; so this `default panic` is here for the sake of completeness. IIRC a few months ago there was a comment from somebody to add this default in a similar situation; which made sense to me to prevent programming errors.

subcommands/keys/tuf_updates_add_offline_key.go
```diff
@@ -0,0 +1,124 @@
...
+		subcommands.AssertWritable(keysFile)
+	} else if errors.Is(err, fs.ErrNotExist) {
+		creds = make(OfflineCreds, 0)
+		saveTufCreds(keysFile, creds)
+	} else {
+		subcommands.DieNotNil(err)
+	}
+
+	updates, err := api.TufRootUpdatesGet(factory)
+	subcommands.DieNotNil(err)
```
mike-sul, line 71: Will it produce something like "No ongoing TUF meta update" or just some HTTP status code and message (404)?

subcommands/keys/tuf_updates_add_offline_key.go
```diff
@@ -0,0 +1,124 @@
...
+	factory := viper.GetString("factory")
+	txid, _ := cmd.Flags().GetString("txid")
+	keyTypeStr, _ := cmd.Flags().GetString("key-type")
+	keyType := ParseTufKeyType(keyTypeStr)
+	keysFile, _ := cmd.Flags().GetString("keys")
+	roleName, _ := cmd.Flags().GetString("role")
+	roleName = ParseTufRoleNameOffline(roleName)
+
+	var creds OfflineCreds
+	if _, err := os.Stat(keysFile); err == nil {
```
mike-sul, line 59: The lines 59-68 seems like a common pattern for any the tuf update sub-command that requires a key file, so maybe add some helper function?

subcommands/keys/tuf_updates_add_offline_key.go
```diff
@@ -0,0 +1,124 @@
...
+	err = api.TufRootUpdatesPut(factory, txid, newCiRoot, newProdRoot, nil)
+	handleTufRootUpdatesUpload(tmpFile, keysFile, err)
+}
+
+func addOfflineRootKey(root *client.AtsTufRoot, creds OfflineCreds, keyType TufKeyType) {
+	subcommands.DieNotNil(checkNoTufSigner(root, creds, root.Signed.Roles["root"].KeyIDs))
+
+	kp := genTufKeyPair(keyType)
+	root.Signed.Keys[kp.signer.Id] = kp.atsPub
+	root.Signed.Expires = time.Now().AddDate(1, 0, 0).UTC().Round(time.Second) // 1 year validity
```
mike-sul, line 103: Why the root meta is refreshed (expiration reset) if a root key is being added and we don't do it if a targets key is being added?

mike-sul (approved):
LGTM, mostly just questions.

mike-sul:
I know a key file can have many keys, one active and some old/rotated keys (not present in the current root). But if we allow adding a new key file into existing key file it means it will have more than one active key.

subcommands/keys/tuf_updates_add_offline_key.go
```diff
@@ -0,0 +1,124 @@
...
+
+The new offline signing key will be used in both CI and production TUF root.
+
+When you add a new TUF targets offline signing key, existing production targets are not signed by it.
+Please, use the sign-prod-targets subcommand if you want to sign existing production targets with a new key.`,
+		Example: `
+- Add offline TUF root key:
+  fioctl keys tuf updates add-offline-key \
+	 --txid=abc --role=root --keys=tuf-root-keys.tgz
+- Add offline TUF targets key, explicitly specifying new key type (and signing algorithm):
```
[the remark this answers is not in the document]
vkhoroz: There are several signing algorithms used to sign the TUF metadata: `rsassa-pss-sha256, ed25519, ecdsa-sha2-nistp256` (we support 2 first of them).
Each signing algorithm is tied to a specific key type: `RSA, ED25519, ECDSA`, correspondingly (although TUF spec says nothing about the letter case, `ota-community-edition` wants them to be upper case).

Thus, by selecting a key type, user also selects a corresponding signing algorithm.

(I basically copied this message from other, already existing Fioctl commands).

subcommands/keys/tuf_updates_add_offline_key.go
vkhoroz: They use a common pattern, but do a different thing inside that pattern.

There is already a sufficient enough generalizations in a form of `GetOfflineCreds, AssertWritable`.

subcommands/keys/tuf_updates_add_offline_key.go
mike-sul: ok

subcommands/keys/tuf_updates_add_offline_key.go
mike-sul: Not sure I get it. Adding a key doesn't mean key rotation. So, why we refresh root meta each time new root key is added to it is not clear to me.

subcommands/keys/tuf_updates_add_offline_key.go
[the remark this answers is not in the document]
mike-sul: I know it. My point is that a user specifies just a key type to this command. The sentence sounds me like the command accepts two params. Maybe it should be new key type (the key type determines corresponding signing algo).

subcommands/keys/tuf_updates_add_offline_key.go
vkhoroz: I thought it is not clear to you why targets key addition/rotation does not refresh expiration.
So, I answered that part of your question.

For a question why do we refresh expiration when a user adds a new key I have no clear answer.
The only argument is: because we do the same thing for the TUF root rotation, so we need a parity here.
If you think that _this is bad_ - I'm fine to remove this line, so that we only increase expires field at TUF root rotation.

In the future, I think we need to come out with a solution when _"root.json expires field equals to the minimum value of expiration of all its keys"_. But, TUF spec does not have a "standard" solution for that; so we need to invent our own wheel. That part is our of scope.

mike-sul: I don't know the answer myself. On the one side, it makes sense to refresh root meta expiration each time its updated (new root meta version), then if a new root key or targets key or any other role keys is added to room meta, then its expiration should be refreshed.

On the other hand, a user may wanna have a full control over root meta expiration, then we should not refresh at all.

So, I am fine with either refresh if anything changes in root meta or don't refresh on changes unless a user specifically says so. I am not sure I get the mix approach, if a root key is added we refresh if some other role key is added we don't.

Maybe, we should just add a flag `--refresh-root` to that command.

subcommands/keys/tuf_updates_add_offline_key.go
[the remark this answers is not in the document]
mike-sul: np.

subcommands/keys/tuf_updates_add_offline_key.go
mike-sul: I guess it's not critical atm.

vkhoroz: After some thought I think this is the correct fixup at our current stage: https://github.com/foundriesio/fioctl/pull/241/commits/c13c0c6142c7c4b1e7d1a9262163b42c93548e90

subcommands/keys/tuf_updates_add_offline_key.go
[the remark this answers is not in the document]
vkhoroz: The aforementioned changes require changes to the API, so that the user can view which production signatures were added to the transaction.
So, I'll do them as a separate PR.

vkhoroz:
@doanac @mike-sul if you are OK with the latest fixup I'm going to merge this, as it was tested and works with the current API.

mike-sul:
> @doanac @mike-sul if you are OK with the latest fixup I'm going to squash and merge this, as it was tested and works with the current API.

Sure, I am fine with it.

doanac (approved):

subcommands/keys/tuf_updates_add_offline_key.go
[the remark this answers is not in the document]
vkhoroz: https://github.com/foundriesio/fioctl/pull/244

PR #242, opened 2023-06-26
Feature: ability to delete offline TUF signing key

vkhoroz:

### DIFF
--- a/subcommands/keys/tuf_updates_delete_offline_key.go
+++ b/subcommands/keys/tuf_updates_delete_offline_key.go
@@ -1,0 +1,120 @@
+package keys
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"golang.org/x/exp/slices"
+
+	tuf "github.com/theupdateframework/notary/tuf/data"
+
+	"github.com/foundriesio/fioctl/subcommands"
+)
+
+func init() {
+	del := &cobra.Command{
+		Use:   "delete-offline-key --role root|targets --txid=<txid> --keys=<tuf-root-keys.tgz>|--key-id=<key-id>",
+		Short: "Stage deletion of the offline TUF signing key for the Factory",
+		Long: `Stage deletion of the offline TUF signing key for the Factory.
+
+There are two ways to delete the offline TUF signing key:
+- If you own the keys file - you can delete your key by providing your keys file.
+  Fioctl will search through your keys file for an appropriate key to delete.
+- You can also provide an exact key ID to delete.
+
+When you delete the TUF targets offline signing key:
+- if there are production targets in your factory, corresponding signatures are also deleted.
+  if any production targets lack enough signatures - you need to sign them using the "sign-prod-targets" command.
+- if there is an active wave in your factory, the TUF targets key deletion is not allowed.`,
+		Example: `
+- Delete offline TUF root key:
+  fioctl keys tuf updates delete-offline-key \
+    --txid=abc --role=root --keys=tuf-root-keys.tgz
+- Delete offline TUF targets key by its ID:
+  fioctl keys tuf updates delete-offline-key \
+    --txid=abc --role=targets
+	 --key-id=15bbb6e79c9ac73b2db7df73c96f3a4937a25d948c048ba0208e49e426e5888a`,
+		Run: doTufUpdatesDeleteOfflineKey,
+	}
+	del.Flags().StringP("role", "r", "", "TUF role name, supported: Root, Targets.")
+	_ = del.MarkFlagRequired("role")
+	del.Flags().StringP("txid", "x", "", "TUF root updates transaction ID.")
+	del.Flags().StringP("keys", "k", "", "Path to <tuf-root-keys.tgz> used to sign TUF root.")
+	_ = del.MarkFlagFilename("keys")
+	del.Flags().StringP("key-id", "i", "", "A key ID to delete, as specified in your TUF root.")
+	del.MarkFlagsMutuallyExclusive("keys", "key-id")
+	tufUpdatesCmd.AddCommand(del)
+}
+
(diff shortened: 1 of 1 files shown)


subcommands/keys/tuf_updates_delete_offline_key.go
```diff
@@ -0,0 +1,120 @@
...
+	var (
+		creds OfflineCreds
+		err   error
+	)
+
+	if keysFile != "" {
+		creds, err = GetOfflineCreds(keysFile)
+		subcommands.DieNotNil(err)
+	} else if keyId == "" {
+		subcommands.DieNotNil(errors.New(
```
doanac, line 68: In theory this can't happen right? (ie cobra prevents this).

subcommands/keys/tuf_updates_delete_offline_key.go
```diff
@@ -0,0 +1,120 @@
...
+			subcommands.DieNotNil(errors.New("Unable to find online target key for factory"))
+		}
+		if keyId != "" && keyId == onlineTargetsId {
+			subcommands.DieNotNil(fmt.Errorf(
+				"It is not allowed to delete an online TUF targets key: %s", onlineTargetsId,
+			))
+		}
+		validKeyIds = subcommands.SliceRemove(roleToUpdate.KeyIDs, onlineTargetsId)
+	default:
+		panic(fmt.Errorf("Unexpected role name: %s", roleName))
```
doanac, line 99: I don't think cobra prevents a bad value for this - so we need a subcommands.DieNotNil, right?

subcommands/keys/tuf_updates_delete_offline_key.go
vkhoroz: A cobra makes sure that _both --keys and --key-id_ cannot be set at once. But we also need to check that _at least --keys or --key-id_ is set. Cobra has no utility to check the second condition.

subcommands/keys/tuf_updates_delete_offline_key.go
```diff
@@ -0,0 +1,120 @@
...
+	roleToUpdate.KeyIDs = subcommands.SliceRemove(roleToUpdate.KeyIDs, keyId)
+
+	fmt.Println("= Delete keyid:", keyId)
+	newCiRoot.Signatures = make([]tuf.Signature, 0)
+	removeUnusedTufKeys(newCiRoot)
+	newProdRoot := genProdTufRoot(newCiRoot)
+
+	fmt.Println("= Uploading new TUF root")
+	subcommands.DieNotNil(api.TufRootUpdatesPut(factory, txid, newCiRoot, newProdRoot, nil))
+}
```
mike-sul, line 120: So, we don't delete the key file, just remove the key from root meta?

subcommands/keys/tuf_updates_delete_offline_key.go
```diff
@@ -0,0 +1,120 @@
...
+				"It is not allowed to delete an online TUF targets key: %s", onlineTargetsId,
+			))
+		}
+		validKeyIds = subcommands.SliceRemove(roleToUpdate.KeyIDs, onlineTargetsId)
+	default:
+		panic(fmt.Errorf("Unexpected role name: %s", roleName))
+	}
+
+	if keyId == "" {
+		oldKey, err := FindOneTufSigner(newCiRoot, creds, validKeyIds)
```
mike-sul, line 103: I have the same question again. I know that `FindOneTufSigner` returns the first and only key from a single key file. But, the previous PR with the "add-key" command allows adding a new key into existing file, it means that a single key file can have more than one key?

vkhoroz: The `add-offline-key` command only allows to add 1 key of the same role into the same file.
It fails if it notices that there are some actual keys already.
So, no, no way to have 2 actual keys in 1 file using fioctl; a user needs to do some hackery to achieve that.

subcommands/keys/tuf_updates_delete_offline_key.go
mike-sul: ok.
How a user can list IDs of keys located in a single key file? For example, if a user decides to delete one of the old keys, or actually add it again to root meta :). I think we forbid it for purpose since it sounds too hackish :).

subcommands/keys/tuf_updates_delete_offline_key.go
mike-sul: I see it now. Thx.

subcommands/keys/tuf_updates_delete_offline_key.go
vkhoroz: > How a user can list IDs of keys located in a single key file?

I do `vim keys.tar.gz` for that... only because I'm too lazy to remember a proper `tar -tf keys.tar.gz` (or what) command.
For offline keys, a key ID is a part of a file name.

mike-sul: >or offline keys, a key ID is a part of a file name.

I meant the case when a single key file has many keys, one active (present in root) and a few older/rotated.

If we delete a key with ID=`id1` from a key file `id1.tgz` and then add a new key with ID=`id2` then the fioctl command will create a new key file `id2.tgz` or rename `id1.tgz` to `id2.tgz`?

mike-sul (approved):

subcommands/keys/tuf_updates_delete_offline_key.go
vkhoroz: Sorry, you misread my comment because I gave not enought details.

A typical `keys.tgz` file has an inner folder structure, where the leaf nodes have names like `tuf-root-<id1>.pub, tuf-root-<id1>.sec, tuf-root-<id2>.pub, tuf-root-<id2>.sec, ...`. A simple way to get a list of key IDs is to list file names inside a tarball.

mike-sul: ok. This is a file name inside the TAR, not the *.tgz file.

PR #245, opened 2023-07-06
cmd: Allow config file to come from environment

doanac:
This allows you to set FIOCTL_CONFIG=/tmp/fioctl.yaml and then run normal fioctl commands from a shell. This is quite handy when working on multiple factories (without having to introduce a more complex config file format).

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -116,6 +116,10 @@ func getConfigDir() string {
 }
 
 func initConfig() {
+	cfgFileFromEnv := os.Getenv("FIOCTL_CONFIG")
+	if len(cfgFileFromEnv) > 0 {
+		cfgFile = cfgFileFromEnv
+	}
 	if cfgFile != "" {
 		// Use config file from the flag.
 		viper.SetConfigFile(cfgFile)


cmd/root.go
```diff
@@ -116,6 +116,10 @@ func getConfigDir() string {
 }
 
 func initConfig() {
+	cfgFileFromEnv := os.Getenv("FIOCTL_CONFIG")
+	if len(cfgFileFromEnv) > 0 {
+		cfgFile = cfgFileFromEnv
+	}
 	if cfgFile != "" {
 		// Use config file from the flag.
 		viper.SetConfigFile(cfgFile)
```
vkhoroz, line 125: I'm fine with it.

The only question is: why isn't the `-c <config>` parameter convenient enough?

vkhoroz (approved):
LGTM

cmd/root.go
doanac: when you start working with our creds helpers for docker and git. you can just set an env variable in the shell you are in and then have things work

PR #246, opened 2023-07-07
feat(config): Show CreatedBy user name

StealthyCoder:
Just added the information on who created the config.

### DIFF
--- a/subcommands/common_config.go
+++ b/subcommands/common_config.go
@@ -71,10 +71,19 @@ func SetConfig(opts *SetConfigOptions) {
 }
 
 type LogConfigsOptions struct {
+	UserLookup    map[string]client.FactoryUser
 	Limit         int
 	ShowAppliedAt bool
 	ListFunc      func() (*client.DeviceConfigList, error)
 	ListContFunc  func(string) (*client.DeviceConfigList, error)
+}
+
+func (opts *LogConfigsOptions) GetUsersLookupById(api *client.Api, factory string) {
+	list, err := api.UsersList(factory)
+	DieNotNil(err)
+	for _, user := range list {
+		opts.UserLookup[user.PolisId] = user
+	}
 }
 
 func LogConfigs(opts *LogConfigsOptions) {
@@ -93,6 +102,13 @@ func LogConfigs(opts *LogConfigsOptions) {
 		}
 		DieNotNil(err)
 		for _, cfg := range dcl.Configs {
+			if len(cfg.CreatedBy) > 0 {
+				if v, ok := opts.UserLookup[cfg.CreatedBy]; ok {
+					cfg.CreatedBy = fmt.Sprintf("%s / %s", v.PolisId, v.Name)
+				} else {
+					cfg.CreatedBy = fmt.Sprintf("%s / ?", cfg.CreatedBy)
+				}
+			}
 			PrintConfig(&cfg, opts.ShowAppliedAt, true, "")
 			if listLimit -= 1; listLimit == 0 {
 				return
@@ -128,6 +144,9 @@ func PrintConfig(cfg *client.DeviceConfig, showAppliedAt, highlightFirstLine boo
 		firstLine.Printf(indent+"Created At:    %s\n", cfg.CreatedAt)
 	} else {
 		printf("Created At:    %s\n", cfg.CreatedAt)
+	}
+	if len(cfg.CreatedBy) > 0 {
+		printf("Created By:    %s\n", cfg.CreatedBy)
 	}
 	if showAppliedAt {
 		printf("Applied At:    %s\n", cfg.AppliedAt)
--- a/subcommands/devices/show.go
+++ b/subcommands/devices/show.go
@@ -126,6 +126,14 @@ func doShow(cmd *cobra.Command, args []string) {
 	}
 	if device.ActiveConfig != nil {
 		fmt.Println("Active Config:")
+		if len(device.ActiveConfig.CreatedBy) > 0 {
+			user, err := api.UserAccessDetails(factory, device.ActiveConfig.CreatedBy)
+			if err != nil {
+				device.ActiveConfig.CreatedBy = fmt.Sprintf("%s / ?", device.ActiveConfig.CreatedBy)
+			} else {
+				device.ActiveConfig.CreatedBy = fmt.Sprintf("%s / %s", user.PolisId, user.Name)
+			}
+		}
 		subcommands.PrintConfig(device.ActiveConfig, true, false, "\t")
 	}
 	if len(device.PublicKey) > 0 {
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -50,6 +50,7 @@ type ConfigCreateRequest struct {
 type DeviceConfig struct {
 	CreatedAt string       `json:"created-at"`
 	AppliedAt string       `json:"applied-at"` // This is not present in factory config
+	CreatedBy string       `json:"created-by"`
 	Reason    string       `json:"reason"`
 	Files     []ConfigFile `json:"files"`
 }


StealthyCoder:
@doanac Can you help me test the other commands on this so that they display and work correctly still?

subcommands/config/log.go
```diff
@@ -30,18 +30,52 @@ func doConfigLog(cmd *cobra.Command, args []string) {
...
+			},
+			ListContFunc: func(url string) (*client.DeviceConfigList, error) {
+				list, err := api.FactoryListConfigCont(url)
+				for _, dc := range list.Configs {
+					if len(dc.CreatedBy) > 0 {
+						user, _ := api.UserAccessDetails(factory, dc.CreatedBy)
+						dc.CreatedBy = user.Name
+					}
+				}
+				return list, err
```
vkhoroz, line 51: These new lines of code seem to be the same for all functions, so they'd better go into the `LogConfigs`.

If you need to add one more parameter into the `LogConfigsOptions` to make this possible (e.g. `Factory`) - it's better to add that single parameter than copy paste the same 7 lines of code in 6 places.

subcommands/config/log.go
```diff
@@ -30,18 +30,52 @@ func doConfigLog(cmd *cobra.Command, args []string) {
...
+			},
+			ListContFunc: func(url string) (*client.DeviceConfigList, error) {
+				list, err := api.FactoryListConfigCont(url)
+				for _, dc := range list.Configs {
+					if len(dc.CreatedBy) > 0 {
+						user, _ := api.UserAccessDetails(factory, dc.CreatedBy)
+						dc.CreatedBy = user.Name
```
vkhoroz, line 48: What about the performance part of this?

IIRC we keep a maximum of 10 config log entries, so it should be _not so bad_ to make 10 serial user API calls.
But, we have plans to increase the config history limit to 20 or 50 in the future, as it is highly likely some customers might ask for that.
That would make performance implications much worse.

So, I think we can be smart here:
- Most of the time, `CreatedBy` will be the same for all config modifications (at most 3 unique values in 99% percentile).
  That allows to make a trivial optimization by grabbing all unique `CreatedBy` values, and only make one request for each of them.
- The above should be enough for now, so I will not propose to parallelize.
  But at least we need to add a comment like _this can be parallelized if config history limit grows above 10_.

subcommands/config/log.go
```diff
@@ -30,18 +30,52 @@ func doConfigLog(cmd *cobra.Command, args []string) {
 		subcommands.LogConfigs(&subcommands.LogConfigsOptions{
 			Limit: listLimit,
 			ListFunc: func() (*client.DeviceConfigList, error) {
-				return api.FactoryListConfig(factory)
+
+				list, err := api.FactoryListConfig(factory)
+				for _, dc := range list.Configs {
+					if len(dc.CreatedBy) > 0 {
+						user, _ := api.UserAccessDetails(factory, dc.CreatedBy)
+						dc.CreatedBy = user.Name
```
vkhoroz, line 38: I wonder if we _really_ want to show a user name?

In some places like this we show just user IDs, and that is good enough.
That way all my other _optimization_ comments become irrelevant.

But, if there is a good reason to show a user name - I'd also show a user ID and/or email.
Both attributes provide a way more valuable debugging information to a seasoned admin.

doanac: I think we can do this in a better way that addresses performance and also doesn't make users have to constantly look up uid's. We have an api `UsersList` which looks up all the users in a factory in one call returning an array like:
```
{
        "polis-id": "59db9c9a1c85010019e023cc",
        "name": "Andy Doan",
        "role": "owner",
        "email": "andy@foundries.io"
    },
```

There's a few ways to do it. One is we pass this array to the LogConfigOptions. Then that function can do a lookup *and* I think it removes a bunch of copy/paste we have in this PR

vkhoroz: @doanac isn't a list API paging?

I mean, as we grow - there would be no wonder to have a factory with 5000 users.
At that size, the list-all API becomes non-scalable; a get one-by-one is easier scale as we have an eternal hard limit here.

P.S. Even better would be to have an API which accepts a list of user IDs and returns objects.
But, at our current stage I see no need for that; this task can be done without API changes.

doanac: I don't believe the API paginates. Its one of the ones we've concluded can function fine even with large numbers.

vkhoroz: If no pagination - we can give it a try.

vkhoroz: Though I'm not sure if it will save a lot of code.

Duplication here is easily tackled; so it's a matter of _how many lines of code to have in one place_.

subcommands/config/log.go
StealthyCoder: Removed these in update. :smile:

subcommands/common_config.go
```diff
@@ -71,6 +72,8 @@ func SetConfig(opts *SetConfigOptions) {
 }
 
 type LogConfigsOptions struct {
+	Api           *client.Api
+	Factory       string
```
doanac, line 76: I guess this is okay. I suspect volodya will have stronger opinions though. In my mind I was thinkink you could just have a:
```
UserLookup map[string]User
```
This can then be nil so that volodya's request to make it optional stays possible in our API.

Then you could add a simple helper function to this file like like `GetUsersLookup(api, factory) map[string]User` and then the users of this struct have an easy way to populate it.

subcommands/common_config.go
```diff
@@ -93,6 +98,11 @@ func LogConfigs(opts *LogConfigsOptions) {
 		}
 		DieNotNil(err)
 		for _, cfg := range dcl.Configs {
+			if len(cfg.CreatedBy) > 0 {
+				if index := slices.IndexFunc(users, func(user client.FactoryUser) bool { return user.PolisId == cfg.CreatedBy }); index >= 0 {
+					cfg.CreatedBy = users[index].Name
```
doanac, line 103: we might format this like `<id> / <name>` so you get something like `59db9c9a1c85010019e023cc / Andy Doan`

subcommands/common_config.go
```diff
@@ -71,12 +71,21 @@ func SetConfig(opts *SetConfigOptions) {
...
 
 type LogConfigsOptions struct {
+	UserLookup    map[string]client.FactoryUser
 	Limit         int
 	ShowAppliedAt bool
 	ListFunc      func() (*client.DeviceConfigList, error)
 	ListContFunc  func(string) (*client.DeviceConfigList, error)
 }
 
+func (lco *LogConfigsOptions) GetUsersLookupById(api *client.Api, factory string) {
```
doanac, line 81:
```suggestion
func (opts *LogConfigsOptions) GetUsersLookupById(api *client.Api, factory string) {
```

doanac (commented):
looking good to me now

subcommands/common_config.go
```diff
@@ -93,6 +102,11 @@ func LogConfigs(opts *LogConfigsOptions) {
 		}
 		DieNotNil(err)
 		for _, cfg := range dcl.Configs {
+			if len(cfg.CreatedBy) > 0 {
+				if v, ok := opts.UserLookup[cfg.CreatedBy]; ok {
+					cfg.CreatedBy = fmt.Sprintf("%s / %s", v.PolisId, v.Name)
+				}
```
vkhoroz, line 108: Maybe, show `deadbeef / <deleted user>` or `deadbeef / <unknown user>` when there is no information in a map.
It looks like we are losing some information _as is_.

subcommands/devices/show.go
```diff
@@ -126,6 +126,11 @@ func doShow(cmd *cobra.Command, args []string) {
 	}
 	if device.ActiveConfig != nil {
 		fmt.Println("Active Config:")
+		if len(device.ActiveConfig.CreatedBy) > 0 {
+			user, err := api.UserAccessDetails(factory, device.ActiveConfig.CreatedBy)
+			subcommands.DieNotNil(err)
```
vkhoroz, line 131: It is OK to have a config which was created by a user which was removed.
So, I'd show smth like `<deleted user>` or `<unknown user>` and a user ID in this case, rather than erroring out.

That said, I don't know all the details.
Maybe, we keep users info forever, even when they are technically deleted.

vkhoroz (approved):
LGTM

Just need to decide what to do with deleted/unknown users.

subcommands/common_config.go
doanac: I *think* (but couldn't confirm with a quick search) that we do a "?" when the username isn't found.

vkhoroz (approved):
LGTM

doanac (approved):

MrCry0 (approved):

PR #247, opened 2023-07-10
Add in new "ci" subcommand

doanac:
This is primarily for adding two new subcommands:
 * ci workers
 * ci status

These give us a view like gavelci.net/status/ and ci.foundries.io/status/ have. They are hidden commands because only administrative users can run them.

While doing this work, I went ahead and moved the somewhat confusing "secrets" subcommand under this new CI umbrella to help make it more clear.


subcommands/ci/cmd.go
```diff
@@ -8,14 +8,15 @@ import (
...
+}
+
 func NewCommand() *cobra.Command {
-	cmd := &cobra.Command{
-		Use:   "ci",
-		Short: "Commands that interact with the FoundriesFactory CI server",
-		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			api = subcommands.Login(cmd)
-		},
-	}
```
vkhoroz: I think it'd look nicer if it looks like this immediately in the `skeleton` commit.

client/foundries_ci.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (a *Api) CiWorkersList() ([]Worker, error) {
+	url := a.serverUrl + "/workers/"
+	type response struct {
+		Data struct {
+			Next    string   `json:"next"`
+			Workers []Worker `json:"workers"`
+		} `json:"data"`
+	}
```
vkhoroz, line 29: I buy that you want this to be private, that's fine.
But, any reason it is not a module level type?

vkhoroz (approved):
LGTM

The only _serious_ comment is about deprecation.

client/foundries_ci.go
doanac: a big regret i have with jobserv was this "data" envelope. It serves no value or purpose. 

but - milo now implemented this support in admin.fio. I'm not sure this feature is actually needed any more in fioctl. I'm thinking we just close this for now and keep it in our back-pocket for another day.

PR #250, opened 2023-07-19
Adjust subcommands to execute correctly on Windows

detsch:
Key changes are related to implementing keys and certificates generation in Go instead of using external script. I've tried to mimic the exact same output when using the original openssl commands. The x509 library, however, always considers the `X509v3 Key Usage` fields as critical, and `X509v3 Extended Key Usage` as not critical, leading to the following differences on the generated certificates:

**`tls-crt`, openssl:**
```
            X509v3 Extended Key Usage: critical
                TLS Web Server Authentication
```
**`tls-crt`, Go:**
```
            X509v3 Extended Key Usage: 
                TLS Web Server Authentication
```

and

**`factory_ca.pem`, openssl:**
```
            X509v3 Basic Constraints: 
                CA:TRUE
            X509v3 Key Usage: 
                Certificate Sign
            X509v3 Extended Key Usage: critical
                TLS Web Client Authentication, TLS Web Server Authentication
```
**`factory_ca.pem`, Go:**
```
            X509v3 Basic Constraints: critical
                CA:TRUE
            X509v3 Key Usage: critical
                Certificate Sign
            X509v3 Extended Key Usage: 
                TLS Web Server Authentication, TLS Web Client Authentication
```

It is possible to adjust the `critical` flag for each one of those fields, but it requires copying a bunch of code from the x509 library within fioctl, which I believe is not worth it.

I'm also not sure if there is a better way to get the CommonName for `local-ca.pem` (e.g., `fio-61263c864de34394ed8b66e4`). I'm currently looking for a `CN = ` line in the `create_device_ca` script

### DIFF
--- a/subcommands/keys/ca_create.go
+++ b/subcommands/keys/ca_create.go
@@ -3,13 +3,13 @@ package keys
 import (
 	"fmt"
 	"os"
-	"os/exec"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
 	"github.com/foundriesio/fioctl/subcommands"
+	"github.com/foundriesio/fioctl/x509"
 )
 
 var (
@@ -70,12 +70,10 @@ func writeFile(filename, contents string, mode os.FileMode) {
 	}
 }
 
-func run(name string, arg ...string) string {
-	cmd := exec.Command(name, arg...)
-	cmd.Stderr = os.Stderr
-	out, err := cmd.Output()
+func getDeviceCaCommonName(factory string) string {
+	user, err := api.UserAccessDetails(factory, "self")
 	subcommands.DieNotNil(err)
-	return string(out)
+	return "fio-" + user.PolisId
 }
 
 func doCreateCA(cmd *cobra.Command, args []string) {
@@ -97,42 +95,34 @@ func doCreateCA(cmd *cobra.Command, args []string) {
 	resp, err := api.FactoryCreateCA(factory)
 	subcommands.DieNotNil(err)
 
-	writeFile("ca-csr", resp.CaCsr, 0400)
-	writeFile("tls-csr", resp.TlsCsr, 0400)
-	writeFile("create_ca", *resp.CreateCaScript, 0700)
-	writeFile("create_device_ca", *resp.CreateDeviceCaScript, 0700)
-	writeFile("sign_ca_csr", *resp.SignCaScript, 0700)
-	writeFile("sign_tls_csr", *resp.SignTlsScript, 0700)
+	writeFile(x509.OnlineCaCsrFile, resp.CaCsr, 0400)
+	writeFile(x509.TlsCsrFile, resp.TlsCsr, 0400)
+	writeFile(x509.CreateCaScript, *resp.CreateCaScript, 0700)
+	writeFile(x509.CreateDeviceCaScript, *resp.CreateDeviceCaScript, 0700)
+	writeFile(x509.SignCaScript, *resp.SignCaScript, 0700)
+	writeFile(x509.SignTlsScript, *resp.SignTlsScript, 0700)
 
 	fmt.Println("Creating offline root CA for Factory")
-	run("./create_ca")
+	resp.RootCrt = x509.CreateFactoryCa(factory)
 
 	fmt.Println("Signing Foundries TLS CSR")
-	resp.TlsCrt = run("./sign_tls_csr", "tls-csr")
-	writeFile("tls-crt", resp.TlsCrt, 0400)
+	resp.TlsCrt = x509.SignTlsCsr(resp.TlsCsr)
+	writeFile(x509.TlsCertFile, resp.TlsCrt, 0400)
 
 	if createOnlineCA {
 		fmt.Println("Signing Foundries CSR for online use")
-		resp.CaCrt = run("./sign_ca_csr", "ca-csr")
-		writeFile("online-crt", resp.CaCrt, 0400)
+		resp.CaCrt = x509.SignCaCsr(resp.CaCsr)
+		writeFile(x509.OnlineCaCertFile, resp.CaCrt, 0400)
 	}
+
 	if createLocalCA {
 		fmt.Println("Creating local device CA")
 		if len(resp.CaCrt) > 0 {
 			resp.CaCrt += "\n"
 		}
-		run("./create_device_ca", "local-ca.key", "local-ca.pem")
-		buf, err := os.ReadFile("local-ca.pem")
-		if err != nil {
-			fmt.Println("ERROR:", err)
-			os.Exit(1)
-		}
-		resp.CaCrt += string(buf)
+		commonName := getDeviceCaCommonName(factory)
+		resp.CaCrt += x509.CreateDeviceCa(commonName, factory)
 	}
-
-	buf, err := os.ReadFile("factory_ca.pem")
-	subcommands.DieNotNil(err)
-	resp.RootCrt = string(buf)
 
 	fmt.Println("Uploading signed certs to Foundries")
 	subcommands.DieNotNil(api.FactoryPatchCA(factory, resp))
--- a/x509/bash.go
+++ b/x509/bash.go
@@ -1,0 +1,46 @@
+//go:build !windows
+
+package x509
+
+import (
+	"os"
+	"os/exec"
+
+	"github.com/foundriesio/fioctl/subcommands"
+)
+
+func run(name string, arg ...string) string {
+	cmd := exec.Command(name, arg...)
+	cmd.Stderr = os.Stderr
+	out, err := cmd.Output()
+	subcommands.DieNotNil(err)
+	return string(out)
+}
+
+func CreateFactoryCa(ou string) string {
+	run("./" + CreateCaScript)
+	return string(readFile(FactoryCaCertFile))
+}
+
+func CreateDeviceCa(cn string, ou string) string {
+	run("./"+CreateDeviceCaScript, DeviceCaKeyFile, DeviceCaCertFile)
+	return string(readFile(DeviceCaCertFile))
+}
+
+func SignTlsCsr(csrPem string) string {
+	return run("./"+SignTlsScript, TlsCsrFile)
+}
+
+func SignCaCsr(csrPem string) string {
+	return run("./"+SignCaScript, OnlineCaCsrFile)
+}
+
+func SignEl2GoCsr(csrPem string) string {
+	tmpFile, err := os.CreateTemp("", "el2g-*.csr")
+	subcommands.DieNotNil(err)
+	defer os.Remove(tmpFile.Name())
+	defer tmpFile.Close()
+	_, err = tmpFile.Write([]byte(csrPem))
+	subcommands.DieNotNil(err)
+	return run("./"+SignCaScript, tmpFile.Name())
+}
--- a/x509/golang.go
+++ b/x509/golang.go
@@ -1,0 +1,196 @@
+//go:build windows
+
+package x509
+
+import (
+	"bytes"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/asn1"
+	"encoding/pem"
+	"math/big"
+	"os"
+	"time"
+
+	"github.com/foundriesio/fioctl/subcommands"
+)
+
+func writeFile(filename, contents string, mode os.FileMode) {
+	err := os.WriteFile(filename, []byte(contents), mode)
+	subcommands.DieNotNil(err)
+}
+
+func genRandomSerialNumber() *big.Int {
+	// Generate a 160 bits serial number (20 octets)
+	max := big.NewInt(0).Exp(big.NewInt(2), big.NewInt(160), nil)
+	serial, err := rand.Int(rand.Reader, max)
+	subcommands.DieNotNil(err)
+	return serial
+}
+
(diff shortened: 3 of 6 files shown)


StealthyCoder:
If the command in full Golang library is feature parity with the bash script relying on openssl, why not just only have the x509 Golang implementation? That makes the code much simpler. :slightly_smiling_face:

subcommands/keys/ca_create.go
```diff
@@ -78,7 +83,24 @@ func run(name string, arg ...string) string {
...
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		if strings.HasPrefix(scanner.Text(), "CN = ") {
+			return strings.TrimSpace(strings.Split(scanner.Text(), "=")[1])
+		}
+	}
+	err = scanner.Err()
+	subcommands.DieNotNil(err)
+	return ""
+}
```
vkhoroz, line 100: Just decide on your own... I think below should help.

IIUC the CN is always in a form `fio-<current-user-id>`.
We can use it to simplify this code like below by calling the server:
```
user, err := api.UserAccessDetails(factory, "self")
return "fio-" + user.PolisId
```

subcommands/keys/ca_create.go
```diff
@@ -105,29 +128,53 @@ func doCreateCA(cmd *cobra.Command, args []string) {
...
+		factoryKey, factoryCa = subcommands.CreateFactoryCa(factory)
+	}
 
 	fmt.Println("Signing Foundries TLS CSR")
-	resp.TlsCrt = run("./sign_tls_csr", "tls-csr")
+	if useBashScripts {
+		resp.TlsCrt = run("./sign_tls_csr", "tls-csr")
+	} else {
+		resp.TlsCrt = subcommands.SignTlsCsr([]byte(resp.TlsCsr), factoryCa, factoryKey)
+	}
```
vkhoroz, line 145: It'd be nice if this could be implemented via the build constraints.

What I mean is to have this layout directory:
```
x509/
    bash.go
    windows.go
```

Above, both files will declare a corresponding build constraint and implement a common interface:
```
// bash.go:
// go:build !windows

func CreateFactoryCa() {
    run("./create_ca")
}

// other functions below
```

```
// windows.go
// go:build windows

// This uses a global context to hold the CA operations intermediate results
// Even better approach would be if Windows implementation writes files to disk, just like Bash implementation.
// The user needs a backup of those files, as they contain important bits, just IMHO.

type context struct {
    rootCaKey *ecdsa.PrivateKey
    rootCaCert *x509.Certificate
}

var ctx context

func CreateFactoryCa() {
   ctx.rootCaKey = ...
   ctx.rootCaCert = ...
}

// other functions below
```

Then, the el2g and keys could simply call the common function interfaces, and be sure that a correct flavor is used.
For example, here you could write:
```
    // add above the import for github.com/foundriesio/fioctl/x509
    x509.CreateFactoryCa()
```

This way, we get a different binary for windows vs non-windows executables, which only contain the necessary bits.

doanac:
> Having the ability to customize the certificates creation process might be important to some users. Using the external script allows that.

Some people are uncomfortable with us executing scripts (remote code execution). Can't make everyone happy on this. However, over time we could probably move to this totally as we see it tested more and more.

detsch:
> Overall looks good, just build pragmas are more to my taste than a necessity to put useBashScripts everywhere.
> 
Assuming we want only one implementation available for each built binary, I totally agree, that would be a better way to handle the code.
The current approach, however, makes it easier to switch between the alternatives at runtime. For example, we could have useBashScripts=false by default and have some option (or automatic mechanism) to force the use of the bash scripts if the user wants to use customized scripts for certificates creation.
So, it depends on what our plan is. If we may want to eventually use the Go implementation in Linux as well, I would keep the current approach. Otherwise I can adjust the code to use pragmas.

vkhoroz:
> If we may want to eventually use the Go implementation in Linux as well, I would keep the current approach. Otherwise I can adjust the code to use pragmas.

Me thinks that _we would like to eventually move all OSes to the Golang-based implmenetation_ and leave bash scripts for true nerds.
For that, my plan would be this:
- have a golang pragma only enabled for Windows, and have it undergo a live test in the wild.
- when it is stable enough, enable pragma for all OSes locally, and run some smoke tests.
- finally, just remove the bash-based implementation and get rid of all pragmas.

For this plan having 2 distinct files implementing one interface makes less code changes in the future.
Command-line option - is not an option; I'd never want to go for that.

Lets say this: if 5 years from now there is somebody who wants to do bash scripts - let them invent their own wheel; ours will be all Golang.

StealthyCoder:
I concur with @vkhoroz on the build pragmas. It is the Golang way of doing things, makes for great and easier maintainable code and paves the way for a better future for us. 

We do not have to throw bash (or BATCH or any other scripting) out the window, you could easily add with that interface a check for some runtime custom override action and execute that. Similar to how those credential helpers work for `git` or `docker`. :+1:

StealthyCoder:
> bash scripting does not work for Windows.

Not entirely true nowadays with that whole [Windows Subsystem for Linux (WSL)](https://learn.microsoft.com/en-us/windows/wsl/install).

detsch:
> For that, my plan would be this:
> 
> * have a golang pragma only enabled for Windows, and have it undergo a live test in the wild.
> * when it is stable enough, enable pragma for all OSes locally, and run some smoke tests.
> * finally, just remove the bash-based implementation and get rid of all pragmas.

Sounds good. I'll update the code using pragmas per @vkhoroz suggestion.

detsch:
An additional comment: I'm fine with merging this pull request into the `mcu-api` branch instead of `main` if that is more appropriate for now. Either way, rebasing the `mcu-api` branch would be good, so we can have all the required support for MCUs and Windows in a single branch with a recent code-base.

subcommands/keys/ca_create.go
```diff
@@ -97,43 +95,35 @@ func doCreateCA(cmd *cobra.Command, args []string) {
...
 	subcommands.DieNotNil(err)
 
-	writeFile("ca-csr", resp.CaCsr, 0400)
-	writeFile("tls-csr", resp.TlsCsr, 0400)
+	writeFile(x509.OnlineCaCsrFile, resp.CaCsr, 0400)
+	writeFile(x509.TlsCsrFile, resp.TlsCsr, 0400)
 	writeFile("create_ca", *resp.CreateCaScript, 0700)
 	writeFile("create_device_ca", *resp.CreateDeviceCaScript, 0700)
 	writeFile("sign_ca_csr", *resp.SignCaScript, 0700)
 	writeFile("sign_tls_csr", *resp.SignTlsScript, 0700)
```
vkhoroz, line 103: If you defined a few constants here; I guess it's worth defining a constant for all these file names, right?

subcommands/keys/ca_create.go
```diff
@@ -97,43 +95,35 @@ func doCreateCA(cmd *cobra.Command, args []string) {
...
 
 	fmt.Println("Creating offline root CA for Factory")
-	run("./create_ca")
+	resp.RootCrt = x509.CreateFactoryCa(factory)
```
vkhoroz, line 106: This sets the same bits as the previous implementation at lines 133-135, right?
Just asking to double check.

x509/bash.go
```diff
@@ -0,0 +1,50 @@
...
+	// If csrFileName contains a "*", csrPemBytes is written to a temp file to be used inside the script
+	var fileName string
+	if strings.Contains(csrFileName, "*") {
+		tmpFile, err := os.CreateTemp("", csrFileName)
+		subcommands.DieNotNil(err)
+		_, err = tmpFile.Write([]byte(csrPemBytes))
+		subcommands.DieNotNil(err)
+		defer tmpFile.Close()
+		fileName = tmpFile.Name()
+		defer os.Remove(fileName)
```
vkhoroz, line 45: From https://go.dev/tour/flowcontrol/13:
```
Deferred function calls are pushed onto a stack. When a function returns, its deferred calls are executed in last-in-first-out order. 
```

So, here, the `os.Remove` will be executed before the `os.Close`; that may lead to _unexpected consequences_.

We actually need to close the file _before_ its name is passed into a script.
(Thus, I see a bug both here and in the original implementation).

So, the below order would be a correct one:
```
... := CreateTemp(...)
defer os.Remove(...)
... = tmpFile.Write(...)
tmpFile.Close()
// After this line you may be sure the file contents were flushed to OS, so we can pass a file name into a script.
```

x509/bash.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func SignTlsCsr(csrPemBytes []byte) string {
+	return run("./sign_tls_csr", TlsCsrFile)
+}
+
+func SignCaCsr(csrPemBytes []byte, csrFileName string) string {
+	// If csrFileName contains a "*", csrPemBytes is written to a temp file to be used inside the script
+	var fileName string
+	if strings.Contains(csrFileName, "*") {
```
vkhoroz, line 38: I've spent a few minutes to understand _WHY_ this is special cased like this.

Although this technically works, I see a few problems with it:
1. The `x509/bash.go:SignCaCsr` is _polluted_ with the temp file handling, only specific to the El2Go.
2. The `x509/golang.go:SignCaCsr` is _polluted_ with the `csrFileName` it doesn't use at all.

So, I'd choose which of the above problems needs to be solved, and eliminate it.

To me, it seems that the second problem has a higher priority, as we're looking forward to switch to the `golang` implementation in the future.
So, it looks like we need to do smth like this:
- Define the `func SignCaCsr(csrPemBytes []byte) string` to only accept bytes.
- This works as is for the `ca_create` use case.
- For the `el2go` use case we need to incorporate the temp file logic somehow.
  I see two options how to solve this:
  * A: implement the temp file handling in el2go via `if runtime.OS == "windows"`.
  * B (my preference): define a separate function for el2go like `SignEl2GoCsr` which would:
    + hide the temp file specifics in `bash.go` file;
    + just call the SignCaCsr in golang.go.

x509/golang.go
```diff
@@ -0,0 +1,192 @@
...
+		IsCA:        true,
+		KeyUsage:    x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment | x509.KeyUsageKeyAgreement,
+		ExtKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		DNSNames:    csr.DNSNames,
+	}
+	crtPem := genCertificate(&crtTemplate, factoryCa, csr.PublicKey, factoryKey)
+	return crtPem
+}
+
+func SignCaCsr(csrPemBytes []byte, csrFileName string) string {
```
vkhoroz, line 174: We need to decide if we _do everything via files_ versus _do everything via bytes_, and then _if we use bytes or strings everywhere_.
Each option has its pros and cons, so the decision would be tough.

I mean literally to do either of these two:
A. Use bytes:
```
func SignTlsCsr(csrPem, factoryKeyPem, factoryCaPem []byte) (crtPem []byte)
func SignCaCsr(csrPem, factoryKeyPem, factoryCaPem []byte) (crtPem []byte)
```
B. Use strings:
```
func SignTlsCsr(csrPem, factoryKeyPem, factoryCaPem string) (crtPem string)
func SignCaCsr(csrPem, factoryKeyPem, factoryCaPem string) (crtPem string)
```
C. Use files:
```
// factoryKeyFile and factoryCaFile can be constants, like now
func SignTlsCsr(csrPemFile string) (crtPemFile string)
func SignCaCsr(csrPemFile string) (crtPemFile string)
```

Just IMHO a situation were we mix those things leads to interface discrepancies.

That said, **_(!) I'm not strong on this (!!!)_**.
Except for the `SignCaCsr` thing, all other functions seem easy to understand.
So, we can move on "as is" and improve iteratively later.

@detsch just give it a thought, and if you don't see a fast way forward to address this comment - then it is better to ignore it. Think about the `x509/golang.go` interface; `x509/bash.go` is secondary.

vkhoroz (approved):
Nice work @detsch, thanks.

This looks good to me, and I vote to merge it into main.

Let @doanac give it a second look before the merge.

subcommands/keys/ca_create.go
detsch: Correct. In the case of the bash implementation, there is a `return string(readFile(FactoryCaCertFile))` inside CreateFactoryCa.

subcommands/keys/ca_create.go
detsch: I did not create constants for the scripts names. But since the same scripts names are used here and inside the bash implementation, it seems to be worth it. I'll do it.

x509/bash.go
detsch: Fixed, thanks.

x509/bash.go
detsch: Having a separate `SignEl2GoCsr` makes sense. Implementing it now.

x509/golang.go
detsch: That inconsistency was bothering me too. Using our x509.* methods without relying on file names would be the best approach. However, we would have to use more temp files inside the `bash.go` methods, and the change would not apply to the `factory_ca.*` files, whose names are hardcoded inside the bash scripts. I'm tempted to leave the implementation as is to avoid messing too much with the bash-based execution flow.
I've changed all pem parameters from `[]byte` to `strings` though, no more bytes there.

vkhoroz: I expected an answer like that... which sounds reasonable to me.
I'm glad you were able to extract a "quick and valuable" bit from my comment. 💪

vkhoroz: I've approved your changes already.
So please, ask @doanac to take a second look, and we're fine to merge.

detsch:
Hi @doanac, please take a look at the changes. Should be ready for merging.

doanac:
volodya reviewed it - its good.

PR #253, opened 2023-07-24
targets: show: Print LmP version if available

mike-sul:


mike-sul:
Once this https://github.com/foundriesio/ci-scripts/pull/295 is merged, then the given change will print a lmp version for targets built after the merge.

mike-sul:
```
## Target: raspberrypi4-64-lmp-2030
	Created:       2023-07-24T09:27:23Z
	Tags:          master
	OSTree Hash:   3b5a89a72b7e952f309010615a901f1cdecff27a7759826c682cd4fd131622bc
	Origin Target: 2028
	LmP Version:   91

```

mike-sul:
The second commit https://github.com/foundriesio/fioctl/pull/253/commits/0022f22f67bb3232172d91b1ce0752bd303c8d5b adds a lmp version to the `device show` output.
I think it's really useful to include a lmp version into the `device show` command output, I am just not sure about implementation. I think adding another field to the device record doesn't make much sense, it's easier to let the api client make another call to fetch details of the target a device currently runs.

```
fioctl devices show 7ddfba8b-09cc-4c59-8880-0bb24725f47d
UUID:		7ddfba8b-09cc-4c59-8880-0bb24725f47d
Owner:		5e9dc69d7407c5010eed1c5a
Factory:	msul-dev01
Production:	false
Up to date:	true
Target:		intel-corei7-64-lmp-2030 / sha256(39af9fd81595b48584c2946035f2a63e71e20d18f6e0403e53f857a2e68c1469)
Ostree Hash:	39af9fd81595b48584c2946035f2a63e71e20d18f6e0403e53f857a2e68c1469
LmP Version:	91
Created At:	2023-06-23T12:37:08+00:00
Created By:	5e9dc69d7407c5010eed1c5a
Last Seen:	2023-07-24T10:53:00+00:00
Tag:		master
Docker Apps:	app-04
Healthy Apps:	app-04
Unhealthy Apps:	
...
```

subcommands/devices/show.go
```diff
@@ -49,6 +49,11 @@ func doShow(cmd *cobra.Command, args []string) {
 	fmt.Printf("Up to date:\t%v\n", device.UpToDate)
 	fmt.Printf("Target:\t\t%s / sha256(%s)\n", device.TargetName, device.OstreeHash)
 	fmt.Printf("Ostree Hash:\t%s\n", device.OstreeHash)
+	if target, errTarget := api.TargetGet(factory, device.TargetName); errTarget == nil {
+		if custom, errCustom := api.TargetCustom(*target); errCustom == nil {
+			fmt.Printf("LmP Version:\t%s\n", custom.LmpVer)
+		}
+	}
```
vkhoroz, line 56: We do fetch the targets for the device already in the API call to get device info.
So, I'd prefer to extend the API to return this additional attribute, rather than calling for targets again.

**TL;DR**
There are more problems to this API call than you might think.
For example, if this is a production device which runs e.g. target `foo-42`, and that target was already pruned from the CI targets due to its age - an API can still return its lmp-ver from prod targets.  You'd need to fully duplicate parts of the current API logic in Fioctl to achieve the same.

vkhoroz (commented):
I like this idea.

subcommands/devices/show.go
mike-sul: >We do fetch the targets for the device already in the API call to get device info.

I actually was looking for it here https://git.foundries.io/development/cloudplatforms/infrastructure/ota-lite/-/blob/master/ota_lite/api/device.py#L916 and couldn't find. Please, point me to the line where it happens.

mike-sul: >For example, if this is a production device which runs e.g. target foo-42, and that target was already pruned from the CI targets due to its age - an API can still return its lmp-ver from prod targets.

I guess, I can add another call to the prod targets if 404 is returned to the initial/CI call.

client/foundries.go
```diff
@@ -291,6 +291,7 @@ type TufCustom struct {
 	OrigUri        string                `json:"origUri,omitempty"`
 	CreatedAt      string                `json:"createdAt,omitempty"`
 	UpdatedAt      string                `json:"updatedAt,omitempty"`
+	LmpVer         string                `json:"lmp-ver,omitempty"`
```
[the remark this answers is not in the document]
mike-sul: I suppose the validation is always useful.

subcommands/devices/show.go
vkhoroz: [Here](https://git.foundries.io/development/cloudplatforms/infrastructure/ota-lite/-/blob/master/ota_lite/api/device.py#L236) and [here](https://git.foundries.io/development/cloudplatforms/infrastructure/ota-lite/-/blob/master/ota_lite/api/device.py#L259).

You've got the idea.

mike-sul: It seems easier just leave it as it is now :).

Trying to adjust `FactoryVersions` for this use-case seems like too much. I am not sure I understand how I can get the `custom.lmp-ver` field if I have just a target name as an input (the custom fields are not stored).

subcommands/devices/show.go
```diff
@@ -49,6 +49,9 @@ func doShow(cmd *cobra.Command, args []string) {
 	fmt.Printf("Up to date:\t%v\n", device.UpToDate)
 	fmt.Printf("Target:\t\t%s / sha256(%s)\n", device.TargetName, device.OstreeHash)
 	fmt.Printf("Ostree Hash:\t%s\n", device.OstreeHash)
+	if len(device.LmpVer) > 0 {
+		fmt.Printf("LmP Version:\t%s\n", device.LmpVer)
```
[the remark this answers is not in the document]
mike-sul: I am not sure about it, at least in scope of this PR/effort.

I think it might be useful to filter out (and/or sort out) devices by a specific lmp version.

Let's roll out this change at first, get some feedback, and then decide on the lmp version application to the devices list command.

PR #254, opened 2023-07-26
Feature: ability to sign a wave by more than 1 key

vkhoroz:

### DIFF
--- a/subcommands/waves/sign.go
+++ b/subcommands/waves/sign.go
@@ -1,0 +1,43 @@
+package waves
+
+import (
+	"github.com/docker/go/canonical/json"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/foundriesio/fioctl/subcommands"
+	tuf "github.com/theupdateframework/notary/tuf/data"
+)
+
+func init() {
+	signCmd := &cobra.Command{
+		Use:   "sign <wave>",
+		Short: "Sign an existing wave targets with additional key",
+		Long: `Sign an existing wave targets with additional key.
+
+This command is only needed when your TUF root requires more than 1 signature for production targets.
+In this case, you cannot roll out or complete a wave before it has enough signatures.`,
+		Run:  doSignWave,
+		Args: cobra.ExactArgs(1),
+	}
+	cmd.AddCommand(signCmd)
+	signCmd.Flags().StringP("keys", "k", "", "Path to <offline-creds.tgz> used to sign wave targets.")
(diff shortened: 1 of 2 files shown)


vkhoroz:
This works with the staging API

subcommands/waves/sign.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/spf13/viper"
+
+	"github.com/foundriesio/fioctl/subcommands"
+	tuf "github.com/theupdateframework/notary/tuf/data"
+)
+
+func init() {
+	signCmd := &cobra.Command{
+		Use:   "sign <wave>",
+		Short: "Sign an existing wave targets with additional key",
```
mike-sul, line 15: Is it `key` or could be `keys`?

subcommands/waves/sign.go
```diff
@@ -0,0 +1,43 @@
...
+
+	wave, err := api.FactoryGetWave(factory, name, true)
+	subcommands.DieNotNil(err)
+
+	var targets tuf.Signed
+	subcommands.DieNotNil(json.Unmarshal(*wave.Targets, &targets))
+	meta, err := json.MarshalCanonical(targets.Signed)
+	subcommands.DieNotNil(err)
+
+	signatures := signTargets(meta, factory, offlineKeys)
```
mike-sul, line 41: So, if there are a few offline keys than this command signs a wave with all of them?

Is it reasonable to allow users choose which of the available keys sign a wave with?

vkhoroz: Our `fioctl keys tuf` commands does not allow to place several _actual_ targets keys into one tgz file.
So, there is no need for any feature supporting key.tgz files with multiple _actual_ targets keys.

subcommands/waves/sign.go
vkhoroz: It is a KEY (no _s_).
See the other comment below explaining why.

subcommands/waves/sign.go
mike-sul: So, why is it `signatures` no `signature`?

vkhoroz: Because `signTargets` returns a slice for historic reasons, and this code undergoes a lot of copy-pasting without overthinking.
I'm not kidding, just being honest ;-).

Does it sound reasonable?

mike-sul: Sure, that's fine just wanna understand the context.

mike-sul (approved):

doanac (approved):

PR #256, opened 2023-08-07
Security: set buildmode to PIE

vkhoroz:
We tested the PIE (position independent executable) and RELRO (read-only relocation) linker flags for quite some time in our server-side Golang executables; they behaved quite well. So, now it is time to apply them to the Fioctl as well to harden its run-time security.

Platform-wise this change has the following effects:
- On Windows and Arm Darwin: no effect, everything is a PIE on these platforms. The `make build` creates _exactly the same binary_.
- On Amd64 Darwin: make a PIE executable. The binary size is slightly increase, giving a slightly better run-time protection.
- On Linux: make a PIE executable using RELRO. (see https://github.com/golang/go/blob/master/src/cmd/link/internal/ld/lib.go#L1551). The binary size is increase by 10%, giving a way better run-time protection.

A size increase from ~21MB to ~23MB for Linux is acceptable for the increased security.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 COMMIT:=$(shell git describe HEAD)$(shell git diff --quiet || echo '+dirty')
 
-# Use linker flags to provide commit info
-LDFLAGS=-ldflags "-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)"
+# Use linker flags to provide commit info and create a PIE
+LDFLAGS=-buildmode=pie -ldflags "-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)"
 
 linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-lint)
 


vkhoroz:
I tested that the new Amd64 Linux binary works well.
Need volunteers to also test Amd64 Darwin and Arm Linux binaries.

doanac (approved):
i'll defer to the security people on this unless they just want to say LGTM.

ricardosalveti (approved):
LGTM, it is a good change.

mike-sul:
I wonder if this condition https://github.com/golang/go/blob/87fe5faffca1bce0db513a3c1bb640ccf6a04bc7/src/cmd/link/internal/ld/target.go#L74 enables RELRO on Linux for non-arm64 architectures.

mike-sul (approved):

vkhoroz:
> I wonder if this condition https://github.com/golang/go/blob/87fe5faffca1bce0db513a3c1bb640ccf6a04bc7/src/cmd/link/internal/ld/target.go#L74 enables RELRO on Linux for non-arm64 architectures.

I believe _everything on Linux is Elf_, at least that's how I read that code: https://github.com/golang/go/blob/master/src/cmd/link/internal/amd64/obj.go#L112

A quick view on the app size tells me that both Amd64 and Arm architectures for Linux had the RELRO enabled when I built all binaries, as it usually results into a noticeable binary size increase. An overhead of the PIE itself is rather small.

vkhoroz:
@mike-sul This:
```
RELRO           STACK CANARY      NX            PIE             RPATH      RUNPATH	Symbols		FORTIFY	Fortified	Fortifiable	FILE
Partial RELRO   No canary found   NX enabled    PIE enabled     No RPATH   No RUNPATH   21280 Symbols	  No	0		0		bin/fioctl-linux-amd64
RELRO           STACK CANARY      NX            PIE             RPATH      RUNPATH	Symbols		FORTIFY	Fortified	Fortifiable	FILE
Partial RELRO   No canary found   NX enabled    PIE enabled     No RPATH   No RUNPATH   21191 Symbols	  No	0		0		bin/fioctl-linux-arm64
```

mike-sul:
> I believe _everything on Linux is Elf_

Sure, that [condition](https://github.com/golang/go/blob/87fe5faffca1bce0db513a3c1bb640ccf6a04bc7/src/cmd/link/internal/ld/target.go#L76) is more complex though. This is fine, I am just too curious :).

vkhoroz:
BTW. In order to make it `Full RELRO` I'd need to pass the flags like this: `extldflags=-Wl,-z,relro,-z,now'`.
But, that would mean all sorts of platform-dependent `if..else` branching logic inside `Makefile`.
I think we're not ready for that yet; at this stage `Partial RELRO` is _good enough_ for the effort.

Issue #45, opened 2020-08-13
fioctl devices config updates: validate --apps and --tags entries


mike-scott:
User can easily typo the entries for `--apps` or `--tags` when using `fioctl`.

In a perfect world, it would validate that the apps exist on the resulting target tag.

doanac:
@mike-scott - i think we have this in place now. re-open if you'd like some additional improvements

PR #257, opened 2023-08-09
WIP: Add self-updater support to fioctl

doanac:
This adds a feature backed by TUF to allow fioctl to detect when there's an update and then optionally update itself.

Its been tested on:
 * linux (amd64)
 * macos (amd64)
 * Windows server (amd64)


subcommands/version/tuf_repo.go
```diff
@@ -0,0 +1,10 @@
+package version
+
+import _ "embed"
+
+// This is generated with:
+//  git show tuf-metadata:repository/root.json | json_pp > ./subcommands/version/root.json
+//go:embed root.json
+var RootContent []byte
+
+const RootUrl = "https://raw.githubusercontent.com/foundriesio/fioctl/tuf-metadata/repository/"
```
vkhoroz, line 10: IIUC this is the TUF repo URL, not just the root URL.
So, I'd name it like `TufRepoUrl`.

subcommands/version/tuf_repo.go
```diff
@@ -0,0 +1,10 @@
+package version
+
+import _ "embed"
+
+// This is generated with:
+//  git show tuf-metadata:repository/root.json | json_pp > ./subcommands/version/root.json
+//go:embed root.json
+var RootContent []byte
```
vkhoroz, line 8: Maybe, `InitialRootContent` or `PreloadedTufRoot`, or any combination of these 2 choices?
That'd make it clear _how_ this variable is used.

Also, it'd be interesting to try making this a `const`, unless you want to override this value in tests.

subcommands/version/tuf_repo.go
vkhoroz, line 10: IIUC this is the TUF repo URL, not just the root URL.
So, I'd name it like TufRepoUrl.

subcommands/version/updates.go
```diff
@@ -0,0 +1,145 @@
...
+type fioctlCustom struct {
+	Platform string `json:"platform"`
+	Uri      string `json:"uri"`
+	Version  string `json:"version"`
+}
+
+func (f *FioctlUpdateFinder) FindLatest() (*FioctlUpdate, error) {
+	latestVer, err := version.NewVersion(Commit)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse binary's version: %s", err)
```
vkhoroz, line 57: IIUC we'll get this error for dirty/dev builds, right?
FWIW that's fine, just asking.

subcommands/version/updates.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+type FioctlUpdate struct {
+	Version *version.Version
+	Uri     string
+	Sha512  data.HexBytes
+	len     int64
+}
+
+func NewFioctlUpdateFinder(localJsonPath string) (*FioctlUpdateFinder, error) {
```
vkhoroz, line 36: IMHO there is no need to prefix this with `Fioctl`. Everything in this repo is `Fioctl`.

subcommands/version/updates.go
```diff
@@ -0,0 +1,145 @@
...
+	return os.Rename(dst, s.path)
+}
+
+func (s jsonFilesStore) DeleteMeta(name string) error {
+	return s.SetMeta(name, nil)
+}
+
+func (j jsonFilesStore) Close() error {
+	return nil
+}
```
vkhoroz, line 210: How are you going to use these 2 methods?
Or, are they needed to implement some `go-tuf` interface?

subcommands/version/updates.go
```diff
@@ -0,0 +1,145 @@
...
+			return nil, err
+		}
+	}
+	if _, err = f.tuf.Update(); err != nil {
+		return nil, err
+	}
+	targets, err := f.tuf.Targets()
+	if err != nil {
+		return nil, err
+	}
```
vkhoroz, line 72: Does it work well if there are no targets yet, or they are empty?

Not sure this is of any criticality. I was just thinking about the _first release with TUF_.

subcommands/version/cmd.go
```diff
@@ -10,23 +10,46 @@ import (
...
 	update, err := updateFinder.FindLatest()
 	DieNotNil(err)
+
+	if len(updateTo) != 0 {
+		if updateTo != update.Version.String() {
+			DieNotNil(fmt.Errorf("invalid version %s != %s", updateTo, update.Version))
+		}
+		if err := update.Do(); err != nil {
+			if os.IsPermission(err) {
+				fmt.Println("There was a permission error while updating fioctl. Please run this again with 'sudo'")
```
vkhoroz, line 39: I'd still call `DieNotNil` in this case.

subcommands/version/updates.go
```diff
@@ -0,0 +1,145 @@
...
+					Sha512:  target.Hashes["sha512"],
+					Version: v,
+					Uri:     custom.Uri,
+					len:     target.Length,
+				}
+				latestVer = v
+			}
+		}
+	}
+	return update, nil
```
mike-sul, line 99: looks like `update` can be `nil` `if custom.Platform != f.platform` or not true at #80?

subcommands/version/updates.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+	dst := s.path + ".tmp"
+	if err = os.WriteFile(dst, bytes, 0o644); err != nil {
+		return err
+	}
+	return os.Rename(dst, s.path)
+}
+
+func (s jsonFilesStore) DeleteMeta(name string) error {
```
mike-sul, line 204: It seems like `DeleteMeta` is not used in this commit, I guess it will be in one of the next commits?

subcommands/version/updates.go
```diff
@@ -91,6 +99,72 @@ func (f *FioctlUpdateFinder) FindLatest() (*FioctlUpdate, error) {
...
+	written int64
+	sha     hash.Hash
+	buff    bytes.Buffer
+}
+
+func (p *progressBar) Write(buff []byte) (n int, err error) {
+	if _, err = p.sha.Write(buff); err != nil {
+		return 0, err
+	}
+	n, err = p.buff.Write(buff)
```
vkhoroz, line 113: Can we write immediately into a temporary file instead?
Buffering 20MB into memory does not seem sane enough.

subcommands/version/updates.go
```diff
@@ -91,6 +99,72 @@ func (f *FioctlUpdateFinder) FindLatest() (*FioctlUpdate, error) {
...
+		return fmt.Errorf("unable to download %s. HTTP_%d: %s", u.Uri, res.StatusCode, pb.buff.String())
+	}
+	fmt.Println("Validating the checksum is", u.Sha512)
+	sha := pb.sha.Sum(nil)
+	if !hmac.Equal(sha, u.Sha512) {
+		return fmt.Errorf("download has incorrect sha: %x != %s", sha, u.Sha512)
+	}
+
+	fmt.Println("Saving new version to", exe)
+	tmp := exe + ".tmp"
```
vkhoroz: Not sure it is sane to write into e.g. `/usr/bin/fioctl.tmp`.
Maybe, we need to createa a file like `/tmp/fioctl.<nonse>/tmp` using some `tempfile` library instead?

subcommands/version/cmd.go
```diff
@@ -34,6 +35,7 @@ func doVersion(cmd *cobra.Command, args []string) {
 		}
 		if err := update.Do(); err != nil {
 			if os.IsPermission(err) {
+				logrus.Debugf("Error was: %s", err)
```
vkhoroz, line 38: I _believe_ this belongs to the previous commit.

subcommands/version/update_self_windows.go
```diff
@@ -0,0 +1,56 @@
...
+
+	delSelfArgs := []string{"cmd.exe", "/C", "start", "/b", "timeout", "1", "&&", "del", oldFile}
+	cmd := exec.Command(delSelfArgs[0], delSelfArgs[1:]...)
+	devnull, _ := os.OpenFile(os.DevNull, os.O_WRONLY, 0755)
+	cmd.Stdout = devnull
+	cmd.Stderr = devnull
+	cmd.Stdin = os.Stdin
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		CreationFlags: syscall.CREATE_NEW_PROCESS_GROUP,
+	}
```
vkhoroz, line 49: You deserve a medal for this 🎖️

subcommands/version/updates.go
```diff
@@ -158,11 +158,7 @@ func (u FioctlUpdate) Do() error {
 	}
 
 	fmt.Println("Saving new version to", exe)
-	tmp := exe + ".tmp"
-	if err = os.WriteFile(tmp, pb.buff.Bytes(), st.Mode()); err != nil {
-		return err
-	}
-	return os.Rename(tmp, exe)
+	return updateSelf(exe, pb.buff.Bytes(), st.Mode())
```
vkhoroz, line 161: I _believe_ the tempfile writing needs to be a common part.

Now, after some thought I think that _maybe_ writing to `/tmp` is not that good after all, as it may be on the other partition, making the rename non-atomic.
Then, adding _nonse_ does not make sense, as it may lead to garbage in `/usr/bin`.
So your original idea with just adding the `.tmp` prefix winds up to be the best choice for us.

I leave it up to you to decide. I guess you've gave it a good thought.

vkhoroz (commented):
LGTM

I have only minor comments and questions.

subcommands/version/updates.go
doanac: It actually shouldn't happen in a dirty/dev build. This would be if you did something like:
```
COMMIT=ThisIsNotAVersionString make fioctl-linux-amd64
```

subcommands/version/updates.go
doanac: yes, but most people see "Update" in relation to our product and think we are talking about device updates

subcommands/version/updates.go
doanac: you want to write to the actual directory up-front to make sure you can write to that directory before wasting everyone's time downloading. Also - can't do a cheap rename (symlink) unless you stay in that directory - you risk crossing partitions.

subcommands/version/updates.go
doanac: its not great but I landed on this and the other tools I see in this realm do it also.

subcommands/version/tuf_repo.go
doanac: 6a17a12450a830ddde791ba680263fb8ceb16c9f

subcommands/version/tuf_repo.go
doanac: you can't do consts with go:embed but - 6a17a12450a830ddde791ba680263fb8ceb16c9f

subcommands/version/updates.go
doanac: 2aa3eb75d5a884cd1c731490c0f5f9781b728863

subcommands/version/updates.go
doanac: 2aa3eb75d5a884cd1c731490c0f5f9781b728863

subcommands/version/updates.go
doanac: yes. its okay. with the caveat below to mike's comment.

subcommands/version/updates.go
doanac: update can be nil. this is okay. cmd.go can handle it now that i added  e6264857fa3ea20984bfbd6064aefd6721fd84bf for the update path.

subcommands/version/cmd.go
doanac: i removed this an embeded it in the error message below

subcommands/version/cmd.go
doanac: c82d19703a95e9cfb5ddf875d240a5ebf360b65d

subcommands/version/updates.go
doanac: we can do a temp file. i tried to avoid because it makes the code a bit more cumbersome and i also was a bit apprehensive on storing something, even temporary, to disk that might be hostile/unverified content.

Its easy enough though - especially now that I've worked mentally through whats required of the windows updater logic.

subcommands/version/cmd.go
vkhoroz: Maybe, this would look better:
```
There was a permission error while updating fioctl: %w.\nPlease, run this again with 'sudo'.
```

I'm tempting to add `sudo !!` here, but that's probably too geeky.
Also, I guess this code path cannot occur on Windows where there is no `sudo`, right?
If not, we can paraphrase this to smth like `Please, run this command again as an admin`.

subcommands/version/updates.go
vkhoroz: There are pros and cons each way.

But, there are code paths where we can write garbage to disk anyway.
So, it's better to have _just one evil_ than a duo.

You've raised a good point about writing _garbage that can be potentially executed_.
So, I'd do this:
```
tmpName := fileName + ".tmp"
tmp = os.Open(tmpName, "w", "600")
// write and verify here
if err != nil {
  os.Remove(tmpName)
}
os.Chmod(tmpName, st.Mode())
os.Rename(tmpName, fileName)
```

That way we protect the user from both _over-consuming computer memory_ and _running unverified code_.

subcommands/version/updates.go
vkhoroz: Yes, you're right. This is well-thought.

subcommands/version/cmd.go
```diff
@@ -18,4 +21,35 @@ func NewCommand() *cobra.Command {
...
+	}
+}
+
+func newUpdateFinder(cmd *cobra.Command) *FioctlUpdateFinder {
+	DieNotNil(viper.BindPFlags(cmd.Flags()))
+	path, err := homedir.Expand("~/.config/fioctl-tuf.json")
+	DieNotNil(err)
+	client, err := NewFioctlUpdateFinder(path)
+	DieNotNil(err)
+	return client
```
mike-sul, line 68: I suppose we expect that client in not nil if `err==nil`, so we don't check it here nor in the `newUpdateFinder` client.

subcommands/version/updates.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	bytes, err := json.Marshal(files)
+	if err != nil {
+		return err
+	}
+	dst := s.path + ".tmp"
+	if err = os.WriteFile(dst, bytes, 0o644); err != nil {
+		return err
+	}
+	return os.Rename(dst, s.path)
```
mike-sul, line 201: sync before the renaming call :) - just kidding, don't think it's really important in this context.

subcommands/version/cmd.go
```diff
@@ -10,23 +10,46 @@ import (
...
 }
 
 func doVersion(cmd *cobra.Command, args []string) {
 	fmt.Println(Commit)
 	updateFinder := newUpdateFinder(cmd)
 	update, err := updateFinder.FindLatest()
 	DieNotNil(err)
+
+	if len(updateTo) != 0 {
+		if updateTo != update.Version.String() {
```
mike-sul, line 36: Don't we need to check if `update` is `nil` at #32 as we do at #45.

subcommands/version/updates.go
mike-sul: I think that keeping 20M in RAM is not an issue at all since modern PCs/laptops have at least 4GB.

mike-sul: I just would do it in the way that is the most simple yet reliable.

subcommands/version/updates.go
```diff
@@ -91,6 +99,72 @@ func (f *FioctlUpdateFinder) FindLatest() (*FioctlUpdate, error) {
...
+	pb := &progressBar{
+		total: u.len,
+		sha:   sha512.New(),
+		buff:  bytes.Buffer{},
+	}
+	if _, err := io.Copy(pb, io.LimitReader(res.Body, u.len)); err != nil {
+		return fmt.Errorf("unable to read response. HTTP_%d: %w", res.StatusCode, err)
+	}
+	fmt.Println()
+	if res.StatusCode != 200 {
```
mike-sul, line 151: I think the command will return the `target size mismatch` error if `res.StatusCode != 200 `.

mike-sul: Just nitpicking. I think the program execution might hang at #147 if a bad actor on a server side declares `res.ContentLength == u.len` and sends less than `res.ContentLength` data and never closes connection.

subcommands/version/updates.go
```diff
@@ -91,6 +99,72 @@ func (f *FioctlUpdateFinder) FindLatest() (*FioctlUpdate, error) {
...
+		return fmt.Errorf("unable to download %s. HTTP_%d: %s", u.Uri, res.StatusCode, pb.buff.String())
+	}
+	fmt.Println("Validating the checksum is", u.Sha512)
+	sha := pb.sha.Sum(nil)
```
mike-sul, line 155: I think that hasher generates a hash/checksum as data goes through it, it doesn't need to store the overall file to generate checksum.
If so, then Volodya's comment makes sense, since we can stream down to the the tmp file.

Maybe it's possible to construct `socket-buffer --> limit reader --> hasher --> tmp.file` pipe that `io.Copy` would pipe data through.

subcommands/version/updates.go
```diff
@@ -91,6 +99,72 @@ func (f *FioctlUpdateFinder) FindLatest() (*FioctlUpdate, error) {
...
+	if !hmac.Equal(sha, u.Sha512) {
+		return fmt.Errorf("download has incorrect sha: %x != %s", sha, u.Sha512)
+	}
+
+	fmt.Println("Saving new version to", exe)
+	tmp := exe + ".tmp"
+	if err = os.WriteFile(tmp, pb.buff.Bytes(), st.Mode()); err != nil {
+		return err
+	}
+	return os.Rename(tmp, exe)
```
mike-sul: `sync` before rename :).

mike-sul: I wonder if OS won't get crazy (or memory access violations or crashes) if the executable that is being memory mapped into the process's virtual memory is replaced on a file system.

subcommands/version/update_self_unix.go
```diff
@@ -0,0 +1,17 @@
...
+	"io/fs"
+	"os"
+)
+
+func updateSelf(exe string, buff []byte, mode fs.FileMode) error {
+	tmp := exe + ".tmp"
+	if err := os.WriteFile(tmp, buff, mode); err != nil {
+		return err
+	}
+	return os.Rename(tmp, exe)
```
mike-sul, line 16: I am still surprised that it works like that in linux. Maybe it works in 99% and some times may lead to crashes.

mike-sul: afaik, not whole executable content is loaded into memory during process startup, it's memory mapped, and data are loaded and written on demand.
At least that is what I remember/learned around 15 years ago. Maybe it's only related only to shared libraries...

subcommands/version/updates.go
mike-sul: I wonder if this "self-updating" will work if `fioctl` is a link?
E.g. `/usr/bin/fioctl --> /myplace/fioctl` will `os.Executable` return `/myplace/fioctl`?

subcommands/version/cmd.go
mike-sul: I think you fixed it in the later commit.

mike-sul:
LGTM, just minor comments and the only concern regarding self-replacing.

subcommands/version/updates.go
```diff
@@ -91,6 +99,72 @@ func (f *FioctlUpdateFinder) FindLatest() (*FioctlUpdate, error) {
...
+	// Print 20 dashes total
+	dashes := 100 * p.written / p.total / 5
+	spaces := 20 - dashes
+	progress := strings.Repeat("*", int(dashes)) + strings.Repeat(" ", int(spaces))
+	fmt.Printf("[%s] %d%%\r", progress, dashes*5)
+	return
+}
+
+func (u FioctlUpdate) Do() error {
+	exe, err := os.Executable()
```
mike-sul, line 124: "Executable returns the path name for the executable that started the current process. There is no guarantee that the path is still pointing to the correct executable. If a symlink was used to start the process, depending on the operating system, the result might be the symlink or the path it pointed to. If a stable result is needed, path/filepath.EvalSymlinks might help."

subcommands/version/updates.go
mike-sul: Ah, I've forgotten about the `inode` stuff in linux and you do renaming - that's why it works on linux.

subcommands/version/update_self_unix.go
mike-sul: I've forgotten about inode and renaming stuff, hence it works on linux systems.

mike-sul (approved):
Just minor comments.

subcommands/version/update_self_unix.go
vkhoroz: Yep, on linux you can run `rm -rf /` and the system will continue to work for quite some time.

vkhoroz (approved):
I have no critical comments leftover.

So, LGTM.

subcommands/version/update_self_unix.go
mike-sul: OK. Let me test it on my dev laptop :)

subcommands/version/cmd.go
doanac: correct. that's just standard golang

subcommands/version/updates.go
doanac: i nearly did :)

subcommands/version/updates.go
doanac: golang has a teewriter that can write to the hasher and file at the same time.

subcommands/version/updates.go
doanac: it works correctly. my home setup has /usr/local/bin/fioctl as a symlink to /home/doanac/code/fioctl.....

subcommands/version/updates.go
doanac: > I wonder if this "self-updating" will work if fioctl is a link?

It does. This is my home set up. I had to set things up to test the "normal" path where its not a link

subcommands/version/updates.go
doanac: the more files/chmod/renames we have the more `if err!=nil` bits you get. I think volodya's requires to just write it straight to disk is fine. ironically - the idea to write it to `/tmp` - that's *RAM* on many linux systems.

I'm going to squash my commits and push something clean. Then I'll do this change so that it won't create crazy conflicts

subcommands/version/updates.go
doanac: nice find - 1b3cec710f1b90145ba4fd84cec82c9d51b694c4

subcommands/version/updates.go
mike-sul: ok.

subcommands/version/cmd.go
doanac: 9b1e12361ca1a48c32e033319dc47e85710962b0

subcommands/version/updates.go
doanac: > That way we protect the user from both over-consuming computer memory and running unverified code.

NOTE: If you at how LInux's VFS layer works - this whole file is going to be in memory. And we limit how much we are willing to read based on TUF.

I think the decision driver here is what looks good stylistically and is easy to maintain. That said - writing to a file doesn't change it much - so i'm going to do that.

doanac: probably easier to read the new Update() logic as a whole, but here's the delta 1f151acebe8e10287b6ba697d9e24137827762dc

I found a slightly cleaner CLI to run on Win32 while testing

mike-sul (approved):

subcommands/version/updates.go
```diff
@@ -0,0 +1,226 @@
...
+
+	pb := &progressBar{
+		total: u.len,
+		sha:   sha512.New(),
+	}
+	tr := io.TeeReader(reader, pb)
+	if _, err := io.Copy(f, tr); err != nil {
+		return fmt.Errorf("unable to save new version: %w", err)
+	}
+	_ = f.Close()
```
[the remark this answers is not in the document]
doanac: cf349e5a13760a29d1a3924c405f5b1b00f81ba5

subcommands/version/updates.go
```diff
@@ -163,6 +163,7 @@ func (u FioctlUpdate) Do() error {
 	_ = f.Close()
 	sha := pb.sha.Sum(nil)
 	if !hmac.Equal(sha, u.Sha512) {
+		_ = os.Remove(tmpExe)
```
mike-sul, line 166: So, the `tmpExe` is removed only if it is fully downloaded and its hash doesn't match expected one?
I thought about removing it if some other error happens during download, e.g. size of data being downloaded exceeded the expected one (the limit reader yields an error) or the `fioctl command` is interrupted by system signal (ctrl-c).
I think it's not critical, anyway the next attempt will rewrite it.

mike-sul (approved):

mike-sul:
> @vkhoroz @mike-sul - I know you've both already approved, but I did do a couple of small changes after that. Please give me one more thumbs up before i squash and merge. thanks

LGTM

subcommands/version/updates.go
doanac: the "limit reader" is put into the "tee reader". if that reader fails, we remove the file. In the event of an "under read", the sha won't match and that will remove the file.

ctrl-c is a problem but I think you are right - lets wait till we redo things for that

mike-sul (approved):

PR #258, opened 2023-08-10
Fix indentation error for gen-rst

kprosise:
A newline was added before each list in command documentation across three commands. This should fix the issue where the output is used in sphinx and an indentation error causes a failure in the doc build.

Docs built without issue when changes were applied.

No related issue to link.


kprosise:
@vkhoroz

vkhoroz (approved):
Thank you @kprosise 

I'll merge and make a new release

PR #261, opened 2023-08-15
github: Add cron job for refreshing TUF timestamp

doanac:
This will run every hours. It will call a script:

 https://github.com/foundriesio/fioctl/blob/tuf-metadata/refresh-metadata.sh

That will check if the timestamp is expiring in the next 48 hours. If so, it will update the timestamp for another 7 days.

### DIFF
--- a/.github/workflows/tuf.yml
+++ b/.github/workflows/tuf.yml
@@ -1,0 +1,27 @@
+name: Refresh timestamp metadata
+on:
+  schedule:
+    # every 8 hours
+    - cron: '0 */8 * * *'
+  workflow_dispatch:
+    branches: [ main ]
+
+jobs:
+  resign:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: tuf-metadata
+      - name: Setup signing key
+        env:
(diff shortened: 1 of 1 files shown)


doanac:
This is kind of weird to test (can't really do it until its in `main`). However, I've run something similar from doanac/fioctl. I believe this gets us to the threshold/logic we agreed upon during the design of this.

mike-sul (approved):
LGTM

StealthyCoder:
If you want you can add other triggers to the workflow so you can actually test if it is working in this PR for example, and/or on your branch :+1: 

for example:

```
on:
  pull_request:
    types:
      - opened
      - synchronize
    paths:
      - 'frontend/**'
    branches:
      - develop
```

which will activate when a PR is created and/or updated against develop and something changed in the frontend directory.

mike-sul:
How the timestamp secret is going to be updated?

doanac:
> How the timestamp secret is going to be updated?

I created one yesterday and set it here: https://github.com/foundriesio/fioctl/settings/secrets/actions

doanac:
> If you want you can add other triggers to the workflow ...

Once in main I can trigger by hand in the github actions UI. They just make you get in it the default branch first for security reasons.

PR #262, opened 2023-08-31
fix(actions): don't run action on forked repos

StealthyCoder:
This small filter will make sure it will only run on the foundriesio/fioctl repo by default.


MrCry0 (approved):
LGTM

doanac (approved):

PR #264, opened 2023-09-06
fix(git): Use absolute path instead

StealthyCoder:
Installing `git` on Windows will most likely place it in the `Program Files` directory.

Following our own [docs](https://docs.foundries.io/latest/getting-started/install-fioctl/index.html#configuring-git) on Windows will run the command as Administrator and place the resulting credential helper in `C:\Program Files\Git\bin\`. In order to make it work as seamless as possible on Windows we use the [absolute path configuration](https://git-scm.com/docs/gitcredentials#_custom_helpers) for the credential helper in the `.gitconfig` file managing the global configuration for git. 

On Windows the absolute path has to be converted into a different format.

1. All `\` have to become `/`
2. All spaces have to be 'escaped' making a ` ` into `\\ `

This works around symlinks, permission errors and other edge cases on Windows.

### DIFF
--- a/subcommands/git/cmd.go
+++ b/subcommands/git/cmd.go
@@ -8,6 +8,7 @@ import (
 	"os/user"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/foundriesio/fioctl/subcommands"
 	"github.com/sirupsen/logrus"
@@ -73,8 +74,10 @@ func doGitCreds(cmd *cobra.Command, args []string) {
 	helperName := "fio"
 	dst := filepath.Join(helperPath, GIT_CREDS_HELPER)
 	if runtime.GOOS == "windows" {
+		// To get around edge cases with git on Windows we use the absolute path
+		// So for example the following path will be used: C:/Program\\ Files/Git/bin/git-credential-fio.exe
 		dst += ".exe"
-		helperName += ".exe"
+		helperName = strings.ReplaceAll(filepath.ToSlash(dst), " ", "\\ ")
 	}
 
 	if len(sudoer) > 0 {


subcommands/git/cmd.go
```diff
@@ -73,8 +74,16 @@ func doGitCreds(cmd *cobra.Command, args []string) {
...
+		/*
+		** On Windows should use the absolute path.
+		** If you install git for windows, there is an option to not support symlinks everywhere.
+		** For some reason it will not run the configured credential helper command correctly.
+		** When using an absolute path the `\` should be replaced for `/` characters.
+		** In addition to that, spaces ` ` should be encoded to `\\ `.
+		** So for example the following path will be used: C:/Program\\ Files/Git/bin/git-credential-fio.exe
+		 */
+		dst = strings.ReplaceAll(strings.ReplaceAll(dst, "\\", "/"), " ", "\\\\ ") + ".exe"
+		helperName = dst
```
vkhoroz, line 86: hmm... this is interesting...

So, as I understand the `git` utility under Windows sometimes (or always) runs in a Linux-like shell; otherwise, the conversion to slashes would not work (Windows uses backslashes).

For the first replace (backslash-to-slash), Golang has a [filepath.ToSlash function](https://pkg.go.dev/path/filepath#ToSlash).

Then, a problem with spaces is IMHO not related to Windows, it seems generic enough to be handled outside this `if` clause.
I suspect this same problem could happen in Linux had the `git` utility contain a space in its path.
Thus, if I were you I would try putting the `helperName` into double quotes at lines 94 and 98 below.

subcommands/git/cmd.go
```diff
@@ -73,8 +74,16 @@ func doGitCreds(cmd *cobra.Command, args []string) {
...
-		dst += ".exe"
-		helperName += ".exe"
+		/*
+		** On Windows should use the absolute path.
+		** If you install git for windows, there is an option to not support symlinks everywhere.
+		** For some reason it will not run the configured credential helper command correctly.
+		** When using an absolute path the `\` should be replaced for `/` characters.
+		** In addition to that, spaces ` ` should be encoded to `\\ `.
+		** So for example the following path will be used: C:/Program\\ Files/Git/bin/git-credential-fio.exe
+		 */
```
vkhoroz, line 84: Once done, try to truncate this comment to at most 2 lines.

subcommands/git/cmd.go
StealthyCoder: Tried that, git on Windows will then treat it as a string aka name, and will prepend it with git-credential again :joy:

StealthyCoder: Did not know the ToSlash function. Will utilize that for the first part.

subcommands/git/cmd.go
doanac: nitpick - but this is an unusual way to do comments in golang (the two asterisk lines). https://tip.golang.org/doc/comment

subcommands/git/cmd.go
doanac: Seems like the fix should be in the code that sets `helperPath` and not here. That way the user sees the correct default value in the usage message. Also - have you looked at `filepath.Abs`? I'd think that would work correctly on windows.

StealthyCoder: I will look into `filepath.Abs` as well. To be clear, this format I need to put in there is specific for how `git` on Windows needs it to be in order to run. On a related note, we are [not the only ones struggling with spaces on Windows and commands and formatting](https://github.com/golang/go/issues/17149).

StealthyCoder: `filepath.Abs` sadly does not give that specific format. :slightly_smiling_face:

subcommands/git/cmd.go
StealthyCoder: Fixed this.

subcommands/git/cmd.go
doanac: k. lgtm then. i'll let volodya approve

vkhoroz: Did you check if the Git on Linux processes spaces inside the path correctly?
I suspect it would have the same problem, in which case we need space escaping to be common.

StealthyCoder: I never tried to do it in Linux yet, with absolute paths. So I will try and see what I can do there as well. I also realized I put the logic in the wrong place. The `helperName` should be changed, not `dst`. So I fixed that now. Tested it on Windows and it works correctly now. 

Let me see what happens on Linux when given a file path with spaces and absolute path.

StealthyCoder: Added some logic that detects the usage of `creds-path`, then always use the full path. Most likely scenario on Linux is somebody does not have something in their `PATH` variable we can write to, like a fresh install of any distro and then they give the `--creds-path` and on that change we use the full path in the git config as well.

vkhoroz: I'm not sure I understand the `creds-path` logic; but I guess you know what you do here.

As long as it works I'm fine.

vkhoroz (approved):

subcommands/git/cmd.go
```diff
@@ -73,10 +74,16 @@ func doGitCreds(cmd *cobra.Command, args []string) {
...
+		// To get around edge cases with git on Windows we use the absolute path
+		// So for example the following path will be used: C:/Program\\ Files/Git/bin/git-credential-fio.exe
 		dst += ".exe"
-		helperName += ".exe"
+		helperName = filepath.ToSlash(dst)
+	} else if credsPath.Changed {
+		helperName = dst
 	}
 
+	helperName = subcommands.EncodeWhiteSpace(helperName)
```
doanac, line 85: I'm now confused by this whole PR. I think the normal case (not windows i guess) is to:
 * find a directory in `$PATH` we can write to
 * link file fioctl -> git-credential-fio
 * run `git config ... ` to set the helper

The whole point of finding something in $PATH is so we don't need absolute paths?

It almost sounds like on windows - we aren't finding something in the user's actual path?

If we do need absolute paths, rather than trying to "encode" (nitpick - I think we are *escaping*) - we probably need to put quotes around what we present to git-config. That way what winds up in git config looks like:
```
helper = "c:\\program files\blah.exe"
```
I suspect the quotations would address things beyond spaces we haven't considered.

That said - seems like the bug is that we can't find PATH correctly on windows

StealthyCoder: I agree, the problem partly stems from us requiring the user to run as `root`/`Administrator` and the `PATHS` being different for them on Windows. I will briefly look into a way if you can _easily_ get the original calling user on Windows. If that is the case we could either add in logic to add things to the PATH for that user, or place the symlink into one of the places in their PATH locations. Then we can drop this special case for Windows again, apart from the naming having an extra `.exe`. :+1:

vkhoroz (approved):
LGTM

I'm fine to merge this as long as @doanac is fine with it.

doanac (approved):

PR #265, opened 2023-09-07
Fix: the initial offline TUF targets key rotation

vkhoroz:
A search for the previous offline key to remove old signature fails for the initial rotation. This is expected as there is no offline key yet; so it should not prevent the key rotation.


doanac (approved):
thanks!

PR #266, opened 2023-09-08
client: Fix user API URL

doanac:
The User API as defined by ota-lite wants a trailing slash. ota-lite sends a redirect back with the trailing slash. However, certainer certain reverse-proxies (GCP in our case) will try and resend the request but omit the auth header which results in our API getting an HTTP 401 response.

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -1628,7 +1628,7 @@ func (a *Api) UsersList(factory string) ([]FactoryUser, error) {
 }
 
 func (a *Api) UserAccessDetails(factory string, user_id string) (*FactoryUserAccessDetails, error) {
-	url := a.serverUrl + "/ota/factories/" + factory + "/users/" + user_id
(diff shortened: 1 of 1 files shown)


doanac:
@vkhoroz - i still think its probably worth me doing a bit of deeper dive into this, but I decided we probably should have our code base do things as "correct" as possible and not rely on redirects to save us.

doanac:
The entire ota-lite API is set to redirect to a trailing slash. I'm not happy that our whole API wants trailing slashes, but that seems to be our reality

vkhoroz:
> The entire ota-lite API is set to redirect to a trailing slash. I'm not happy that our whole API wants trailing slashes, but that seems to be our reality

I think you're right.
I checked that the `/devices/<name>/` endpoint also requires a training slash.
The only paths which have loose ends are the reposerver ones; but they are too special.

Anyway, I already approved your PR. It makes total sense to me.

PR #267, opened 2023-09-18
Enable Golang implementation of PKI globally

vkhoroz:
We have already tested this implementation on Windows in the field.
Also tested it on Linux using a local build.

### DIFF
--- a/x509/golang.go
+++ b/x509/golang.go
@@ -1,4 +1,6 @@
-//go:build windows
+//go:build go1.15 && !bashpki
+
+// The crypto/ecdsa and crypto/elliptic were introduced in Golang 1.15.
 
 package x509
 
--- a/x509/bash.go
+++ b/x509/bash.go
@@ -1,4 +1,7 @@
-//go:build !windows
+//go:build bashpki || !go1.15
+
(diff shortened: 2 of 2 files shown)


vkhoroz:
I actually tested it with MEDS and it works well.

detsch:
Just a reminder that HSM support was not added to the Golang implementation.
Other than that, LGTM.

detsch (approved):

x509/golang.go
```diff
@@ -1,4 +1,6 @@
-//go:build windows
+//go:build go1.15 && !bashpki
```
doanac, line 1: so `go1.15` means greater than or equal to go1.15?

doanac: i.e. our go.mod is pinned on go1.19 already

vkhoroz: Yes and Yes

Maybe this is redundant, and we only need the `bashpki` flag.
In the initial version I thought to have no `bashpki`, and needed some flag to disable the bash implementation.

vkhoroz:
Closing until I add the HSM support to Golang implementation.
Then I will reopen.

PR #269, opened 2023-09-21
WIP: Use XGo to enabled the CGo based builds

vkhoroz:
This should (expectedly) fail the GitHub build.
I will finish the CI part tomorrow.

But, this works well for me locally, so it should unblock @detsch on the HSM work.

It builds for all supported platforms, and much more.
I tested on Linux AMD64 (Ubuntu and Alpine), also may test on Windows.
I need a volunteer to test the binaries on Mac OS and Arm Linux.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 COMMIT?=$(shell git describe HEAD)$(shell git diff --quiet || echo '+dirty')
 
 # Use linker flags to provide commit info
-LDFLAGS=-ldflags "-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)"
+COMMON_LDFLAGS=-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)
+TARGET_LDFLAGS=
 
 linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-lint)
 builder:=$(shell which xgo 2>/dev/null || echo $(HOME)/go/bin/xgo)
@@ -22,7 +23,8 @@ fioctl-%:
 	@test -x $(builder) || (echo "Please install xgo toolchain $(HOME)/go/bin: go install github.com/crazy-max/xgo@v0.30.0")
 	$(eval GOOS:=$(shell echo $* | cut -f1 -d\- ))
 	$(eval GOARCH:=$(shell echo $* | cut -f2- -d\-))
-	$(builder) --targets=$(GOOS)/$(GOARCH) -out bin/fioctl $(LDFLAGS) .
+	$(eval COMBINED_LDFLAGS=$(COMMON_LDFLAGS) $(TARGET_LDFLAGS))
+	$(builder) --targets=$(GOOS)/$(GOARCH) -out bin/fioctl --ldflags "$(COMBINED_LDFLAGS)" .
 	# This creates files as root, use `sudo chown --reference bin bin/fioctl-*` if you wish them under your user.
 
 format:
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 COMMIT?=$(shell git describe HEAD)$(shell git diff --quiet || echo '+dirty')
 
 # Use linker flags to provide commit info
-COMMON_LDFLAGS=-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)
+VERSION_LDFLAGS=-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)
+COMMON_LDFLAGS=-v -linkmode=external $(VERSION_LDFLAGS)
 TARGET_LDFLAGS=
 
 linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-lint)
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ COMMIT?=$(shell git describe HEAD)$(shell git diff --quiet || echo '+dirty')
 
 # Use linker flags to provide commit info
 VERSION_LDFLAGS=-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)
-COMMON_LDFLAGS=-v -linkmode=external $(VERSION_LDFLAGS)
+COMMON_LDFLAGS=-v -s -w -linkmode=external $(VERSION_LDFLAGS)
 
 linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-lint)
 builder:=$(shell which xgo 2>/dev/null || echo $(HOME)/go/bin/xgo)
@@ -24,7 +24,7 @@ fioctl-%:
 	$(eval GOOS:=$(shell echo $* | cut -f1 -d\- ))
 	$(eval GOARCH:=$(shell echo $* | cut -f2- -d\-))
 	$(eval TARGET_GOTAGS:=netgo,osusergo)
-	$(eval TARGET_LDFLAGS:=$(if $(shell test $(GOOS) = linux && echo "ok"),'-extldflags=-static',))
+	$(eval TARGET_LDFLAGS:=$(if $(shell test $(GOOS) = linux && echo "ok"),'-extldflags=-static -O1',))
 	$(eval TARGET_LDFLAGS:=$(if $(shell test $(GOOS) = windows && echo "ok"),'-extldflags=-static',$(TARGET_LDFLAGS)))
 	$(eval COMBINED_LDFLAGS=$(COMMON_LDFLAGS) $(TARGET_LDFLAGS))
 	$(builder) --targets=$(GOOS)/$(GOARCH) -out bin/fioctl --tags=$(TARGET_GOTAGS) --ldflags "$(COMBINED_LDFLAGS)" .
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,12 @@ COMMON_LDFLAGS=-v -s -w -linkmode=external $(VERSION_LDFLAGS)
 linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-lint)
 builder:=$(shell which xgo 2>/dev/null || echo $(HOME)/go/bin/xgo)
 
+# A crazy-max/xgo does not allow to provide docker args, and creates files as root. The below script sets proper ownership.
+docker::=$(shell which docker)
+docker_command:=xgo-build . && chown --reference /build/bin /build/bin/fioctl-*
+# A substitution below removes the first (run) and the last (.) arguments from the docker run command passed by xgo.
+docker_alias=test $$1 = run && $(docker) run --entrypoint=sh "$${@:2:(($$\#-2))}" -c "$(docker_command)" || $(docker) "$$@"
+
 build: fioctl-linux-amd64 fioctl-linux-arm64 fioctl-windows-amd64 fioctl-darwin-amd64 fioctl-darwin-arm64
 	@true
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,7 +1,8 @@
 COMMIT?=$(shell git describe HEAD)$(shell git diff --quiet || echo '+dirty')
 
 # Use linker flags to provide commit info
-COMMON_LDFLAGS=-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)
+VERSION_LDFLAGS=-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)
+COMMON_LDFLAGS=-v -linkmode=external $(VERSION_LDFLAGS)
```
mike-sul: Is this `-V` needed in the release version?

Makefile
```diff
@@ -2,7 +2,7 @@ COMMIT?=$(shell git describe HEAD)$(shell git diff --quiet || echo '+dirty')
 
 # Use linker flags to provide commit info
 VERSION_LDFLAGS=-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)
-COMMON_LDFLAGS=-v -linkmode=external $(VERSION_LDFLAGS)
+COMMON_LDFLAGS=-v -s -w -linkmode=external $(VERSION_LDFLAGS)
```
mike-sul, line 5: What if a developer would like to debug `fioctl`? Maybe those parameters should be set conditionally depending on "build type"?

mike-sul:
```
INFO: Completed!
# This creates files as root, use `sudo chown --reference bin bin/fioctl-*` if you wish them under your user.
```
Is it necessary to create the resultant files as root?

mike-sul (approved):
LGTM, it works for me. I think it's important to check if it is work for @detsch.

Makefile
vkhoroz: Yes, I want to see what our CI prints.
There is just a handful lines of additional output which gives a lot of context in case of a problem.

Makefile
vkhoroz: This is a kind of an output the developer gets in case of an (intentional) panic:
```
panic: Something went wrong

goroutine 1 [running]:
github.com/foundriesio/fioctl/subcommands/keys.doShowCA(0xc000436800?, {0x7fbd9fef6609?, 0x4?, 0x7fbd9fef660d?})
	/source/subcommands/keys/ca_show.go:94 +0x4ec
github.com/spf13/cobra.(*Command).execute(0xc000439200, {0xc0001c9f20, 0x2, 0x2})
	/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:920 +0x863
github.com/spf13/cobra.(*Command).ExecuteC(0x7fbda0674020)
	/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044 +0x3a5
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:968
github.com/foundriesio/fioctl/cmd.Execute()
	/source/cmd/root.go:65 +0x17b
main.main()
	/source/main.go:6 +0xf
```

In 99.9% of cases it is more than enough to get the idea what went wrong.
The symbols being stripped are the ones needed by e.g. `gbd` to debug local variables etc etc.
If the developer _really needs_ them, and know how to work with `gdb`, they would know how to update the Makefile.

For a regular use case we don't need that functionality, while the storage saving is worthy (~20MB on Windows and 10MB on Linux per one binary).

StealthyCoder: The only thing I want to add is if we ever want to use this in a recipe in bitbake we either need to patch the `-s -w` out or just use the [INHIBIT_PACKAGE_STRIP](https://docs.yoctoproject.org/ref-manual/variables.html#term-INHIBIT_PACKAGE_STRIP) and [INHIBIT_SYSROOT_STRIP](https://docs.yoctoproject.org/ref-manual/variables.html#term-INHIBIT_SYSROOT_STRIP) :+1:

StealthyCoder (approved):

Makefile
vkhoroz: I think for the Yocto we'd need a separate recipe somewhere in meta-lmp.

vkhoroz: I mean, it will not use this make file at all, and use it's own built tool chain, like e.g. fioconfig does.

mike-sul:
> the build is being run inside a container as a root.

Invoke the container with `-u` option?

mike-sul:
>But, I see no real problem with this other than aestetic. So I'd just let the user know and move on.

It's just a bit odd that the resultant file owner is a root user while the `make` command doesn't require root permission :).

mike-sul:
>But, I see no real problem with this other than aestetic.

Yes, it's not important indeed.

vkhoroz:
> > the build is being run inside a container as a root.
> 
> Invoke the container with `-u` option?

For that to work I need to add a user inside with the same id inside the container. That was my first thought. Too much work.

vkhoroz:
BTW... tested the Windows binary and _it just works_.

Does anybody want to check the Mac OS / Arm binaries?

doanac:
> Does anybody want to check the Mac OS / Arm binaries?

I can test Mac x86. Have you tried running the x86 binary in the LmP?

Makefile
```diff
@@ -13,16 +14,16 @@ fioctl-static:
 
 fioctl-linux-amd64:
 fioctl-linux-arm64:
-fioctl-linux-armv7:
+fioctl-linux-arm-7:
```
doanac, line 18: I've never seen it noted that way before. Another thing we could do - stop building it. We don't produce a release artifact for it or ever try and test.

Makefile
```diff
@@ -13,16 +14,16 @@ fioctl-static:
...
-	CGO_ENABLED=0 \
-	GOOS=$(shell echo $* | cut -f1 -d\- ) \
-	GOARCH=$(shell echo $* | cut -f2 -d\-) \
-		go build $(LDFLAGS) -o bin/$@ main.go
-	@if [ "$@" = "fioctl-windows-amd64" ]; then mv bin/$@ bin/$@.exe; fi
+	@test -x $(builder) || (echo "Please install xgo toolchain $(HOME)/go/bin: go install github.com/crazy-max/xgo@v0.30.0")
+	$(eval GOOS:=$(shell echo $* | cut -f1 -d\- ))
+	$(eval GOARCH:=$(shell echo $* | cut -f2- -d\-))
+	$(builder) --targets=$(GOOS)/$(GOARCH) -out bin/fioctl $(LDFLAGS) .
+	# This creates files as root, use `sudo chown --reference bin bin/fioctl-*` if you wish them under your user.
```
doanac, line 35: I don't suppose xgo has a `--user` type option? This will probably only be a problem for someone like myself who has their code on a NFS volume that blocks root access. It's not a blocker, but will change my workflow

Makefile
```diff
@@ -1,7 +1,8 @@
 COMMIT?=$(shell git describe HEAD)$(shell git diff --quiet || echo '+dirty')
 
 # Use linker flags to provide commit info
-LDFLAGS=-ldflags "-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)"
+COMMON_LDFLAGS=-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)
+TARGET_LDFLAGS=
```
doanac: I've had the need for this in the past. thanks

Makefile
```diff
@@ -24,8 +23,11 @@ fioctl-%:
 	@test -x $(builder) || (echo "Please install xgo toolchain $(HOME)/go/bin: go install github.com/crazy-max/xgo@v0.30.0")
 	$(eval GOOS:=$(shell echo $* | cut -f1 -d\- ))
 	$(eval GOARCH:=$(shell echo $* | cut -f2- -d\-))
+	$(eval TARGET_GOTAGS:=netgo,osusergo)
+	$(eval TARGET_LDFLAGS:=$(if $(shell test $(GOOS) = linux && echo "ok"),'-extldflags=-static',))
+	$(eval TARGET_LDFLAGS:=$(if $(shell test $(GOOS) = windows && echo "ok"),'-extldflags=-static',$(TARGET_LDFLAGS)))
```
doanac, line 30: just to make sure I'm groking this: These eval statements put these variable into the environment so that xgo also gets them, correct?

Makefile
doanac: Don't worry about developer mode. If you want, `go run ..`. And tools like `delve` basically do that also.

doanac (approved):
LGTM. I sometimes wonder if the PIE/RELRO for linux-amd64 is adding too much mental overhead when trying to read/maintain this. 99% of people aren't going to benefit from this.

vkhoroz:
So you can play with new binaries.

I temporarily set up an action in my temporary fork:
https://github.com/vkhoroz/fioctl/releases/tag/v0.0.3

Makefile
vkhoroz: No, this is a weird Makefile syntax to set the variable inside the target.
GNU Make ignores all such variables, unless wrapped into the eval.

Anyway, I'm not quite fond of this script and will try to split into separate targets.
I'm really really scared that this Makefile becomes not only hard to grok, but may also fail on other platforms than GNU make.

Makefile
vkhoroz: This actually breaks the CI... so I had to make those files owned by a "real" user.

A problem with crazy-max/xgo is that it does not allow customizing docker args (like the original fork).
I've tried several options, and the one which works the most reliably is to alias the `docker` to pass in a modified `--entrypoint` argument.  I'll push that local commit on Monday.
A plain `--user` thingy did not work properly, it fails to operate inside the container as it tries to create some root-level folders.

Makefile
vkhoroz: We don't build this. So I agree to remove this. A user can anyway try to build any `fioctl-foo-bar` thing from the list of xgo-supported platforms.

The `arm-$X` is how the ARM platforms listed in xgo. I have no preferences for this.

Makefile
mike-sul: That's painful...

detsch:
Paying a bit more attention to the compilation output, that might be the issue:

```
/usr/bin/ld: /tmp/go-link-4230498301/000004.o: in function `New':
/go/pkg/mod/github.com/miekg/pkcs11@v1.1.1/pkcs11.go:77: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

vkhoroz:
> Paying a bit more attention to the compilation output, that might be the issue:
> 
> ```
> /usr/bin/ld: /tmp/go-link-4230498301/000004.o: in function `New':
> /go/pkg/mod/github.com/miekg/pkcs11@v1.1.1/pkcs11.go:77: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
> ```

That might be the thing.
But, I doubt what would be the solution for us in this case.
I kind of understand _why pkcs11_ needs to open the shared object, but then comes the question:
- What are the correct `shared libraries from the glibc version used for linking`?
- There is no way to make sure that we have the same glibc libraries on the running system as we had on the building system.
- So, we should do either of:
  * Get rid of _static linking_ and make a _separate build for every libc we support_.
  * Invent a solution to not require the `libc` dependency for `dlopen`.
  * Invent some other kind of a wheel.

I need to walk the dog on it, and we also need to let @doanac know immediately that HSM + Static binaries risk dragging us into the rabbit hole.

doanac:
> I need to walk the dog on it, and we also need to let @doanac know immediately that HSM + Static binaries risk dragging us into the rabbit hole.

This might be a big rabbit hole. We moved to static binaries ages ago because a binary that worked for say Ubuntu 16.04 wouldn't work for 20.04. I agree this needs some dog walking. And maybe if we restate our priorities, we'll find this isn't quite urgent and there's a different way to get here (probably wishful thinking).

vkhoroz:
Abandoning this for 2 reasons:
- This PoC implements the same thing in a better way: https://github.com/foundriesio/fioctl/pull/277
- This PoC implements HSM support without a necessity to use `dlopen`, and we agreed to go this way: https://github.com/foundriesio/fioctl/pull/278

PR #270, opened 2023-09-22
[WIP] feat: add HSM support to Golang PKI code

detsch:
This commit adds to the Golang PKI code implementation the same HSM access features that were available in the original bash implementation.

----

I did some tests, including a complete `fioctl keys ca create` sequence using the new code + registration of a device with a key signed by the generated key. All seems fine. If the overall approach is OK, I'll do a few more detailed tests, to make sure it did not break something.

The code requires `CGO_ENABLED=1` but I did not commit any changes to the Makefile since @vkhoroz is working on the proper fix (PR #269)

I decided to keep the PR as simple as possible, avoiding to add characteristics that are not present on the original bash implementation. There are at least two things that I believe can be improved. Both seem to be easy to implement:
- Save local device CA private key to HSM as well;
- Allow `fioctl el2g config-device-gateway` to access the factory key from the HSM. That would require the `--hsm-*` options to be replicated in that sub-command

### DIFF
--- a/subcommands/keys/ca_create.go
+++ b/subcommands/keys/ca_create.go
@@ -15,9 +15,7 @@ import (
 var (
 	createOnlineCA bool
 	createLocalCA  bool
-	hsmModule      string
-	hsmPin         string
-	hsmTokenLabel  string
+	hsm            x509.HsmInfo
 )
 
 func init() {
@@ -58,9 +56,9 @@ This is optional.`,
 	caCmd.AddCommand(cmd)
 	cmd.Flags().BoolVarP(&createOnlineCA, "online-ca", "", true, "Create an online CA owned by Foundries that works with lmp-device-register")
 	cmd.Flags().BoolVarP(&createLocalCA, "local-ca", "", true, "Create a local CA that you can use for signing your own device certificates")
-	cmd.Flags().StringVarP(&hsmModule, "hsm-module", "", "", "Create key on an PKCS#11 compatible HSM using this module")
-	cmd.Flags().StringVarP(&hsmPin, "hsm-pin", "", "", "The PKCS#11 PIN to set up on the HSM, if using one")
-	cmd.Flags().StringVarP(&hsmTokenLabel, "hsm-token-label", "", "device-gateway-root", "The label of the HSM token created for this")
+	cmd.Flags().StringVarP(&hsm.Module, "hsm-module", "", "", "Create key on an PKCS#11 compatible HSM using this module")
+	cmd.Flags().StringVarP(&hsm.Pin, "hsm-pin", "", "", "The PKCS#11 PIN to set up on the HSM, if using one")
+	cmd.Flags().StringVarP(&hsm.TokenLabel, "hsm-token-label", "", "device-gateway-root", "The label of the HSM token created for this")
 }
 
 func writeFile(filename, contents string, mode os.FileMode) {
@@ -77,19 +75,20 @@ func getDeviceCaCommonName(factory string) string {
 }
 
 func doCreateCA(cmd *cobra.Command, args []string) {
+	var storage x509.KeyStorage
 	factory := viper.GetString("factory")
 	certsDir := args[0]
 	logrus.Debugf("Create CA for %s under %s", factory, certsDir)
 	subcommands.DieNotNil(os.Chdir(certsDir))
 
-	if len(hsmModule) > 0 {
-		if len(hsmPin) == 0 {
+	if len(hsm.Module) > 0 {
+		if len(hsm.Pin) == 0 {
 			fmt.Println("ERROR: --hsm-pin is required with --hsm-module")
 			os.Exit(1)
 		}
-		os.Setenv("HSM_MODULE", hsmModule)
-		os.Setenv("HSM_PIN", hsmPin)
-		os.Setenv("HSM_TOKEN_LABEL", hsmTokenLabel)
+		storage = &x509.KeyStorageHsm{Hsm: hsm}
+	} else {
+		storage = &x509.KeyStorageFiles{}
 	}
 
 	resp, err := api.FactoryCreateCA(factory)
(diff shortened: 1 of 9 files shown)


mike-sul:
> * Save local device CA private key to HSM as well;

+1

mike-sul:
LGTM, will look closer to it once the `WIP` label is removed.

Maybe the implementation would look slightly better if some OOP is applied, so no need to pass the `hsm` parameter to and do `if hsm {} else {}` in each function. I mean, adding interface, then base implementation, and the hsm implementation that extends the base one.
Just suggestion, I don't have a strong opinion on it.

x509/golang.go
```diff
@@ -112,8 +118,115 @@ func marshalSubject(cn string, ou string) pkix.Name {
...
+		TokenLabel:  hsm.TokenLabel,
+		Pin:         hsm.Pin,
+		MaxSessions: 0,
+	}
+
+	ctx, err := crypto11.Configure(&cfg)
+	subcommands.DieNotNil(err)
+
+	privKey, err := ctx.FindKeyPair(id, nil)
+	subcommands.DieNotNil(err)
```
vkhoroz, line 205: Just asking... So, pkcs11 interface is too low-level to conveniently work with it, right?
That's why we need ThalesIgnite: to call two functions instead of writing 100500 lines of obscure code ourselves, right?

vkhoroz (approved):
LTGM

I only have minor suggestions and questions.

As for the other _nice to have additions_, I think they can be added separately later in case depending on our availability.

x509/golang.go
```diff
@@ -112,8 +118,115 @@ func marshalSubject(cn string, ou string) pkix.Name {
 	return pkix.Name{ExtraNames: pkixAttrTypeValue}
 }
 
-func CreateFactoryCa(ou string) string {
-	priv := genAndSaveKey(FactoryCaKeyFile)
+func genAndSaveKeyToHsmNoRet(hsm HsmInfo, id []byte, label []byte) {
```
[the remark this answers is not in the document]
detsch: Sure, will do.

subcommands/keys/ca_create.go
```diff
@@ -82,14 +80,14 @@ func doCreateCA(cmd *cobra.Command, args []string) {
...
+		if len(hsm.Pin) == 0 {
 			fmt.Println("ERROR: --hsm-pin is required with --hsm-module")
 			os.Exit(1)
 		}
-		os.Setenv("HSM_MODULE", hsmModule)
-		os.Setenv("HSM_PIN", hsmPin)
-		os.Setenv("HSM_TOKEN_LABEL", hsmTokenLabel)
+		os.Setenv("HSM_MODULE", hsm.Module)
+		os.Setenv("HSM_PIN", hsm.Pin)
+		os.Setenv("HSM_TOKEN_LABEL", hsm.TokenLabel)
```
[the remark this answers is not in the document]
detsch: Moved it. Will be part of next push.

x509/golang.go
```diff
@@ -112,8 +118,115 @@ func marshalSubject(cn string, ou string) pkix.Name {
...
+	subcommands.DieNotNil(err)
+	if privKey == nil {
+		subcommands.DieNotNil(errors.New("Key with requested id not found"))
+	}
+
+	return privKey
+}
+
+func genAndSaveKeyToHsm(hsm HsmInfo, id []byte, label []byte) (crypto11.Signer, crypto.PublicKey) {
+	genAndSaveKeyToHsmNoRet(hsm, id, label)
```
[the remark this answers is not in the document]
vkhoroz: Yeah... this is reasonable.

Just add that important bit as a comment, and consider using an anonymous closure to make the code nicer:

```
def genAndSaveKeyToHsm(...) (...) {
    // Need all deferred statements to commit before the getKeyHsm call below.
    func () {
        // Your genAndSaveKeyToHsmNoRet body goes here.
    }()
    priv := getKeyHsm(hsm, id)
    return ...
}
```

x509/golang.go
vkhoroz: Makes sense.

x509/golang.go
```diff
@@ -112,8 +118,115 @@ func marshalSubject(cn string, ou string) pkix.Name {
...
+	var (
+		pub  any
+		priv any
+	)
+
+	if len(hsm.Module) > 0 {
+		priv, pub = genAndSaveKeyToHsm(hsm, []byte(FactoryCaKeyPkcs11Id), []byte(FactoryCaKeyPkcs11Label))
+	} else {
+		priv, pub = genAndSaveKeyToFile(FactoryCaKeyFile)
+	}
```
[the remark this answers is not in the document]
detsch: I've implemented the interface last, so I've got the updated version without the interface as well, if required. I was not sure at first, but I now agree that it is better to have the interface.

subcommands/keys/ca_create.go
```diff
@@ -77,19 +75,20 @@ func getDeviceCaCommonName(factory string) string {
...
+		if len(hsm.Pin) == 0 {
 			fmt.Println("ERROR: --hsm-pin is required with --hsm-module")
 			os.Exit(1)
 		}
-		os.Setenv("HSM_MODULE", hsmModule)
-		os.Setenv("HSM_PIN", hsmPin)
-		os.Setenv("HSM_TOKEN_LABEL", hsmTokenLabel)
+		storage = &x509.KeyStorageHsm{Hsm: hsm}
+	} else {
+		storage = &x509.KeyStorageFiles{}
```
doanac, line 91: I think we need to separate factory-root storage from online-ca storage. It doesn't make sense to have both keys on the same device. You want to keep the root keypair offline and in safe and the online keypair is going to wind up in some manufacturing location.

mike-sul: I think the online device CA private key is stored on our backend. I don't think we need to store its public part/cert in HSM.

I guess the offline/local CA keypair is meant here.

It sounds like we need two params, one to specify "destination" for factory CA and another one for offline/local device CA.

x509/bash.go
```diff
@@ -17,25 +17,26 @@ func run(name string, arg ...string) string {
 	return string(out)
 }
 
-func CreateFactoryCa(ou string) string {
+func CreateFactoryCa(storage KeyStorage, ou string) string {
```
mike-sul, line 20: Are we planning to switch from the bash to go approach for the non-HSM use-case too?

subcommands/keys/ca_create.go
doanac: sorry, i meant offline-ca

x509/bash.go
vkhoroz: @mike-sul yes.

An intent of this PR was to enable the Golang approach for all use cases.
The HSM support was the only missing part to do that.

detsch:
Due to to the impossibility of `dlopen`ing a HSM access .so in a portable manner when using static binaries PR (https://github.com/foundriesio/fioctl/pull/269#issuecomment-1733769404), moved the HSM-specific to a new file, only building if when explicitly enabled. Static builds return an error if the user tries to invoke the Golang HSM implementation.

x509/bash.go
mike-sul: Why the script is called at line #22 then?

vkhoroz: There will be a separate PR where the [go tags here](https://github.com/foundriesio/fioctl/blob/main/x509/bash.go#L1) disable Bash implementation altogether.
I already created such PR, but it is temporarily closed: https://github.com/foundriesio/fioctl/pull/267

vkhoroz:
Tabling this as we agreed to implement the HSM support in a way which does not require cross-compilation using CGo: https://github.com/foundriesio/fioctl/pull/278

If we do need the CGo based compilation, we might revisit this.

Still, a large portion of refactoring done here, was reused in https://github.com/foundriesio/fioctl/pull/278.

PR #272, opened 2023-09-23
(chore): Add linter improvements

camilamacedo86:
## Description
This PR improves our linting process by:
- Update Makefile to improve lint target and add the lint-fix which can fix small issues like format and to sort the imports
- Add linter check using github action, removing the need to check logs. Example:

Screenshot 2023-09-23 at 15 45 45

##Motivation
- Make it easier to fix lint issues and do the review
- Move checks to GitHub actions
- Ensuring consistent linting behaviour
- **Remove the need to set up lint manually. See that when we run make lint if the bin does not exist the target will download it**, see;

```
$ make check                         
make lint
/Users/camilamacedo/go/src/github.com/foundriesio/fioctl/bin/golangci-lint run
```

- Allow leverage in the common checks to avoid errors; see proposed follow-up:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,20 @@
+name: Lint
+on:
+  push:
+  pull_request:
+
+jobs:
+  lint:
+    name: golangci-lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '~1.19'
+      - name: Clone the code
+        uses: actions/checkout@v4
+      - name: Run linter
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.51.2
--- a/Makefile
+++ b/Makefile
@@ -24,12 +24,25 @@ fioctl-%:
 		go build $(LDFLAGS) -o bin/$@ main.go
 	@if [ "$@" = "fioctl-windows-amd64" ]; then mv bin/$@ bin/$@.exe; fi
 
-format:
-	@gofmt -l  -w ./
-check:
-	@test -z $(shell gofmt -l ./ | tee /dev/stderr) || echo "[WARN] Fix formatting issues with 'make format'"
-	@test -x $(linter) || (echo "Please install linter from https://github.com/golangci/golangci-lint/releases/tag/v1.51.2 to $(HOME)/go/bin")
+install-linter:
+	echo "[WARN] Installing golangci binary version v1.51.2 at $(HOME)/go/bin/golangci-lint"
+	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(linter)) v1.51.2
+
+has-linter:
+	@test -x $(linter) || (echo '[ERROR] Please install go linter using "make install-linter"' && exit 1)
+
+linter-check: has-linter
 	$(linter) run
+
+linter-fix: has-linter
+	echo "[WARN] Attempting to automatically fix some linter issues."
+	$(linter) run --fix
+
+format-check:
+	@test -z $(shell gofmt -l ./ | tee /dev/stderr) || echo "[WARN] Fix formatting issues with 'make format-check'"
+
+check: format-check linter-check
+	@true
 
 # Use the image for Dockerfile.build to build and install the tool.
 container-init:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,3 @@
+run:
+  deadline: 5m
+  allow-parallel-runners: true
--- a/.jobserv.yml
+++ b/.jobserv.yml
@@ -1,13 +1,5 @@
 timeout: 30
 triggers:
-  - name: pull-request
-    type: github_pr
-    runs:
-      - name: unit-test
-        container: golangci/golangci-lint:v1.51.2
-        host-tag: amd64
-        script: unit-test
-
   - name: create-release
     type: git_poller
     params:


StealthyCoder:
@doanac I will leave it to you if you want to move these checks to Github Actions rather than our GavelCI :+1:

Makefile
```diff
@@ -24,12 +22,26 @@ fioctl-%:
 		go build $(LDFLAGS) -o bin/$@ main.go
 	@if [ "$@" = "fioctl-windows-amd64" ]; then mv bin/$@ bin/$@.exe; fi
 
+GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
+golangci-lint:
+	@[ -f $(GOLANGCI_LINT) ] || { \
+	set -e ;\
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(GOLANGCI_LINT)) v1.54.2 ;\
+	}
```
vkhoroz, line 30: I'm fine to have this target if it is improved a bit:

1. [required] Install a specific linter version, which is _exactly the same_ as the one used in CI.
2. [recommended] Remove the dependency between lint and `golangci-lint`.
    I call `make check` quite often, and making it run a code which _does nothing_ (except for the first time) almost useless.
    An alternative would be for this target to auto-detect a necessity to upgrade the linter.
    But then comes a question how to keep our many repos in sync.
    For a newbie it would be enough to tell smth like `Please, run "make golangci-lint" to install the linter` when they call the `make check` and there is no linter installed.
    I'd even call it `make dev-deps` in kind if we need to install something else in the future.

Makefile
```diff
@@ -24,12 +22,26 @@ fioctl-%:
...
+
+.PHONY: lint
+lint: golangci-lint  ## Run linter
+	$(GOLANGCI_LINT) run
+
+.PHONY: lint-fix
+lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
+	$(GOLANGCI_LINT) run --fix
```
vkhoroz, line 38: This is kind of nice.

subcommands/keys/ca_show.go
```diff
@@ -169,6 +169,9 @@ func prettyPrint(cert string) {
 		case *ecdsa.PublicKey:
 			fmt.Println("\t\tNIST CURVE:", pub.Curve.Params().Name)
 			fmt.Print("\t\t\t")
+
+			//FIXME: Fix deprecation and remove the nolint
+			//nolint: staticcheck: SA1019: elliptic.Marshal has been deprecated since Go 1.21: for ECDH, use the crypto/ecdh package. This function returns an encoding equivalent to that of PublicKey.Bytes in crypto/ecdh. (staticcheck)
```
vkhoroz, line 174: This one should not have happened; probably related to an uninteded linter upgrade.

The thing is: Golang moves too fast for us.
So we usually stick to stable releases, which are sometimes 1-2 major versions behind.
For example, the highest Golang version supported by LmP is `1.19`, soon will be `1.20`.
Still, we have some users trying to compile Fioctl inside LmP, and we don't want to upset them by using a linter which prints warnings for an unsupported Golang version.

vkhoroz (commented):
It think this is the correct direction, and @doanac would probably agree.

But, please, make this in the least intrusive way.

vkhoroz:
@camilamacedo86 this is also conflicting with my PR https://github.com/foundriesio/fioctl/pull/269.
So, I'd like your changes to be as minimal and the review go as smoothly as possible.

subcommands/keys/ca_show.go
camilamacedo86: >  and we don't want to upset them by using a linter which prints warnings for an unsupported Golang version.

**None of the changes proposed in this PR affect end users; the linter solely raises warnings for us/maintainers**. Could you please clarify your comment? I apologize, but it seems there might have been a misunderstanding, and I want to ensure we're on the same page.

> Still, we have some users trying to compile Fioctl inside LmP

**Linters are not checked during the build/compile process of the tool,** so these issues wouldn't arise in such scenarios. Could you please clarify this point further?

Makefile
camilamacedo86: > [required] Install a specific linter version, which is exactly the same as the one used in CI.
Are you speaking about the job? https://ci.foundries.io/projects/fioctl/builds/574/unit-test/
If so, can you explain why we cannot update the linter version used?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '~1.21'
+      - name: Clone the code
+        uses: actions/checkout@v4
+      - name: Run linter
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.54
```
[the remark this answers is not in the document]
camilamacedo86: All done.

Makefile
camilamacedo86: 1 - Done
2 - I am not sure if I could follow up on your suggestion.

If we call make lint or make check (which calling the lint now) it will download the lint into the bin directory so that we do not need to download it manually. If the linter exists already then we just ignore 
So, please let me know what makes you not happy then I will try to accommodate it for you.

subcommands/keys/ca_show.go
camilamacedo86: By downgrading the linter version the warning is not raised. 
So, I removed the comment.

camilamacedo86:
Hi @vkhoroz,

I addressed all suggestions as much as possible.
Regards:
> @camilamacedo86 this is also conflicting with my PR https://github.com/foundriesio/fioctl/pull/269.
So, I'd like your changes to be as minimal and the review go as smoothly as possible.

Could you please let me know where is the conflict?
The changes in the Makefile are not on the same lines so it seems that will be fine. 
Please, let me know if I missed something.

Makefile
```diff
@@ -24,12 +22,26 @@ fioctl-%:
...
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(GOLANGCI_LINT)) v1.51.2 ;\
+	}
+
+.PHONY: lint
+lint: golangci-lint  ## Run linter
+	$(GOLANGCI_LINT) run
+
+.PHONY: lint-fix
+lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
+	$(GOLANGCI_LINT) run --fix
```
vkhoroz, line 38:
```suggestion
# This is the original line, which should stay at the original place; apparently there are important nuances in it.
linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-lint)

has-linter:
	# Another important nuance here is the `test -x` instead of `test -f`.
	# Your check will succeed when a file is not executable; but the command will then fail.
	@test -x $(linter) || (echo 'Please install go linter using "make install-linter"' && exit 1)

install-linter:
	# Here an important nuance is to have an ability to easily upgrade the linter.
	# Your suggested command is a one-time shooter; this makes it eternally usable.
	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(linter)) v1.51.2

# There is little risk to ever have a file named `lint, lint-fix, lint-check`, so I see no necessity to use the `.PHONY` here.
lint-check: has-linter
	$(linter) run

# This may be a lint-fix if you like it more.
I simply prefer shorter names for regularly used commands (`build, check, format, lint`).
lint: has-linter
	$(linter) run --fix
```

A part of this suggestion is "how we did things in the past", and another part is "how I planned to do things" (a part of my unpublished changes for the new build process).
A rationale is present in inline comments where relevant.  These comments should not be a part of the final commit.

Let's talk in Slack if you have questions about some detail.

Makefile
```diff
@@ -24,12 +22,26 @@ fioctl-%:
...
+
 format:
 	@gofmt -l  -w ./
 check:
 	@test -z $(shell gofmt -l ./ | tee /dev/stderr) || echo "[WARN] Fix formatting issues with 'make format'"
-	@test -x $(linter) || (echo "Please install linter from https://github.com/golangci/golangci-lint/releases/tag/v1.51.2 to $(HOME)/go/bin")
-	$(linter) run
+	make lint
```
vkhoroz, line 44: Now that you improved the linter usage so much, the next natural step would be to also improve the formatter usage the same way:

```suggestion
format-check:
	@test -z $(shell gofmt -l ./ | tee /dev/stderr) || echo "[WARN] Fix formatting issues with 'make format'"

check: format-check linter-check
	@true
```

subcommands/keys/ca_show.go
```diff
@@ -169,6 +169,7 @@ func prettyPrint(cert string) {
 		case *ecdsa.PublicKey:
 			fmt.Println("\t\tNIST CURVE:", pub.Curve.Params().Name)
 			fmt.Print("\t\t\t")
+
```
vkhoroz, line 172: This _no-change_ blank line should not be a part of the commit.

vkhoroz (changes requested):

subcommands/keys/ca_show.go
vkhoroz: Anyway, your new changes are good, so this discussion is no longer needed.

Makefile
vkhoroz: Your new changes were minimal enough to comment on them in a form of a suggestion.
Please, see the related comment below and let's talk in Slack if you disagree.

Makefile
camilamacedo86: Hi @vkhoroz, 
 
> # This is the original line, which should stay at the original place; apparently there are important nuances in it.
linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-lint)

I see some issues with this one:
- 1 - Not necessary your GOPATH is $(HOME)/go (Because of that the suggestion use gopath)
- 2 - We are adding and using the lint from GOPATH/bin so that NOT necessary will be the version required. In this way, due to the fact that it is a bin used in the project itself (specific version). IMO it should be inside of the project/bin and not change or misconfigure things that are valid for ANY project in the dev env. 

> > # A user may be suprised that `make check` installs software on their laptop.
> # Here we explicitly notify the user what software we need and how to easily install it.
> # The user then gives a conscent to it by explicitly running an install command (or can do it their way).
> has-linter:
> 	# Another important nuance here is the `test -x` instead of `test -f`.
> 	# Your check will succeed when a file is not executable; but the command will then fail.
> 	@test -x $(linter) || (echo 'Please install go linter using "make install-linter"' && exit 1)

That is suck because you need to install the right version for each project everytime that you need to check the stuff. I do not think that is a good DX. 

Because of this, the suggestion is:
- Install the required version into the BIN of the project NOT GOPATH and missed up with your staff
- Install when you run if the specific version not be there already
- We can add an echo "We are installing goling bin into the root project bin to do this check"

> # This may be a lint-fix if you like it more.
> # I simply prefer shorter names for regularly used commands (`build, check, format, lint`).
> lint: has-linter
> 	$(linter) run --fix

We should not run the lint run --fix when we do the check. Why? 
- It will do fix and the dev might not be aware of and do not understand why the check lint passes locally (where the issues are fixed because of this command) and is not passing in the CI. It might take a while until the person figures out to run `git status` and check out the changes. 

Also, I would not recommend trying to fix the issue automatically without the dev knolodge. The dev should use it as a helper when the linter fails to see if is or not possible to solve automatically and be conscious about its usage.

camilamacedo86: Hi @vkhoroz,

I have attempted to implement ALL of your suggestions, with just two points that I hope we can agree weren't addressed exactly as you suggested. Below are the inline comments where I share the rationale behind this.

> linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-lint)

- The version of golangci-lint is specific to each project that we are working on, so I do not agree with, nor like the idea of adding it outside of the project directory where it will be valid for all projects. I hope we can agree to keep it inside the project dir `bin/`.
  
- Additionally, using `$(HOME)/go/bin` does not cover all scenarios. I might not set up my GOPATH in `$(HOME)/go/bin`. For such purposes, I would suggest using the $GOPATH environment variable instead.

> I don't go along with the idea of calling $(linter) run --fix every time that we check the linter

Executing `lint run --fix` during a check seems inadvisable. The developer might not be aware of the fixes and may wonder why the lint check passes locally but fails in CI. It might take some time before the developer realizes the need to run `git status` and inspect the changes.

I am of the opinion that fixing issues should be a conscious decision by the developer, not an automated process during a check. The linter should serve as a helper when issues arise, thereby giving the developer the autonomy to decide whether or not to apply fixes automatically.

Please let me know what do you think now.

Makefile
```diff
@@ -3,7 +3,7 @@ COMMIT?=$(shell git describe HEAD)$(shell git diff --quiet || echo '+dirty')
 # Use linker flags to provide commit info
 LDFLAGS=-ldflags "-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)"
 
-linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-lint)
+linter:=$(shell pwd)/bin/golangci-lint
```
camilamacedo86, line 6: Each project requires a specific version of the tool, and as a developer, I prefer not to install or check it in GOPATH/bin, which would make it universally applicable across all projects.

P.S.: $(HOME)/go/bin is not a universal solution for all environments; the correct path would be $GOPATH/bin.

@vkhoroz ^

vkhoroz: That's just how it works for many years, even before I joined Foundries.io.
It may be not ideal, it just works.
Just IMHO it is logical to add GOPATH into PATH in `.bashrc`, in which case it fits all sizes.

vkhoroz (approved):
LGTM

PR #274, opened 2023-09-23
(chore) : add dependabot to keep project updated

camilamacedo86:
## Description

Introduce Dependabot to ensure consistent and effortless updating of the project's dependencies.

GitHub's Dependabot will perform weekly checks and automatically create pull requests to update the Dockerfile, GitHub Actions, and go.mod dependencies. With a reliable Continuous Integration (CI) in place to validate the tool's functionality, this approach significantly streamlines the review process. This allows us to effortlessly keep our project updated, minimizing technical debt and promptly addressing any potential Common Vulnerabilities and Exposures (CVEs).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"


vkhoroz:
Although we don't have a `a reliable Continuous Integration (CI) in place` for this project, adding dependabot might give us some added value.

@doanac I think we can try this, unless you had bad experience with dependabot for Fioctl.

camilamacedo86:
Hi @vkhoroz,

> Although we don't have a a reliable Continuous Integration (CI) in place for this project, adding dependabot might give us some added value.

That would be the goal of https://github.com/foundriesio/fioctl/pull/275

If we can ensure that the build is successful and the binary is functional, then I believe we can have confidence in updating the Go dependencies. The only scenario in which something might stop working would be if a new bug or regression is introduced in the updated dependency version. However, this seems highly unlikely to me, and consequently, I consider this to be a low-risk situation.

.github/dependabot.yml
```diff
@@ -0,0 +1,24 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
```
vkhoroz, line 4: Please, drop this comment.
We use dependabot in many projects already, so it is not something new to us.

.github/dependabot.yml
```diff
@@ -0,0 +1,24 @@
...
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the
+    # default location of `.github/workflows`
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+
+  # Maintain dependencies for go
```
vkhoroz, line 18: Please, strip all these comments. The code is self-descriptive enough.

vkhoroz (changes requested):
LGTM

Please, strip those comments, squash all your commits **locally**, force-push a single commit signed by your GPG (your can use `git commit -sS --amend` at the head for that), and force-push the updated commit history into your branch.

I will do the final sanity check for our SDLC rules, approve and merge this (using the rebase strategy).

No worries, you did a great job.
I wrote this whole paragraph jusy to let you better know in details our usual workflow.

camilamacedo86:
@vkhoroz all suggestions applied !

vkhoroz (approved):
LGTM

PR #275, opened 2023-09-23
Add checks to validate bin builds

camilamacedo86:
### Description

- **Add GitHub action responsible for building and checking binaries for all platforms (less darwin/arm64) using the make targets.**
       - Darwin/arm64 is excluded from the build because it is not supported by GitHub's current infrastructure. 
       - For ARM architectures, this action utilizes QEMU to ensure that the generated ARM binaries are functional.

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -1,0 +1,75 @@
+name: Binaries Build
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build-binaries:
+    name: Build Binaries
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '~1.19'
+
+      - name: Build All Binaries
+        run: make
+
+      - name: Archive Binaries
+        uses: actions/upload-artifact@v2
+        with:
+          name: binaries
+          path: bin/
+
+  test:
+    needs: build-binaries
+    name: Check ${{ matrix.binary }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            binary: fioctl-linux-amd64
+          - os: ubuntu-latest
+            binary: fioctl-linux-arm64
+            qemu: true
+            qemu-platform: aarch64
+          - os: windows-latest
+            binary: fioctl-windows-amd64.exe
+          - os: macos-latest
+            binary: fioctl-darwin-amd64
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Set up QEMU
+        if: matrix.qemu
+        uses: docker/setup-qemu-action@v3
+        with:
+          platforms: ${{ matrix.qemu-platform }}
+
+      - name: Download Binaries
+        uses: actions/download-artifact@v2
+        with:
+          name: binaries
+          path: bin/
+
+      - name: Set Execute Permission
+        if: runner.os != 'Windows'
+        run: chmod +x ./bin/${{ matrix.binary }}
+
+      - name: Install QEMU user static
+        if: matrix.qemu
+        run: sudo apt-get update && sudo apt-get install -y qemu-user-static
+        shell: bash
+
+      - name: Verify binary
+        run: |
+          if ("${{ matrix.qemu }}" -eq "true") {
+            qemu-${{ matrix.qemu-platform }}-static ./bin/${{ matrix.binary }} logout
+          } else {
+            ./bin/${{ matrix.binary }} logout
+          }
+        shell: pwsh
--- a/.jobserv.yml
+++ b/.jobserv.yml
@@ -24,25 +24,3 @@ scripts:
   unit-test: |
     #!/bin/sh -ex
     make check
-
-  create-binaries: |
-    #!/bin/sh -e
-    make
-
-    GH_REPO="https://api.github.com/repos/foundriesio/fioctl"
-    GH_TAGS="$GH_REPO/releases/tags/$tag"
-    AUTH="Authorization: token $(cat /secrets/githubtok)"
-
-    # Validate token.
-    curl -o /dev/null -sH "$AUTH" $GH_REPO || { echo "Error: Invalid repo, token or network issue!";  exit 1; }
-
-    response=$(curl -sH "$AUTH" $GH_TAGS)
-    eval $(echo "$response" | grep -m 1 "id.:" | grep -w id | tr : = | tr -cd '[[:alnum:]]=')
-    [ "$id" ] || { echo "Error: Failed to get release id for tag: $tag"; echo "$response" | awk 'length($0)<100' >&2; exit 1; }
-
-    cd bin
-    for filename in `ls`; do
-      echo "Uploading $filename ..."
-      GH_ASSET="https://uploads.github.com/repos/foundriesio/fioctl/releases/$id/assets?name=$filename)"
-      curl --data-binary @"$filename" -H "$AUTH" -H "Content-Type: application/octet-stream" $GH_ASSET
-    done


.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Build Container
+        run: |
+          make container-init
+
+      - name: Run Container to Build Binaries
+        run: |
+          docker run --rm -v $(pwd):/fioctl fioctl-build sh -c "git config --global --add safe.directory /fioctl && make build"
```
camilamacedo86, line 20: @doanac why do we need this Dockerfile
How are we building the binaries to release? 
Are we using it?

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,126 @@
...
+            qemu-platform: arm64
+          # Let's build on the ubuntu to be faster
+          - os: ubuntu-latest
+            binary: fioctl-windows-amd64.exe
+          - os: macos-latest
+            binary: fioctl-darwin-amd64
+          - os: ubuntu-latest
+            binary: fioctl-linux-armv7
+            qemu: true
+            qemu-platform: linux/arm/v7
```
camilamacedo86, line 27: @vkhoroz @doanac 
To address this one https://github.com/foundriesio/fioctl/pull/269/files#r1334637439 I added the teste here as well

Makefile
```diff
@@ -24,6 +24,12 @@ fioctl-%:
 		go build $(LDFLAGS) -o bin/$@ main.go
 	@if [ "$@" = "fioctl-windows-amd64" ]; then mv bin/$@ bin/$@.exe; fi
 
+fioctl-linux-armv7:
```
StealthyCoder, line 27: I would wait a bit for Volodya his work on the xgo stuff. 😊

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,126 @@
+name: Binaries Build
```
StealthyCoder, line 1: We could take a look at [buildjet,](https://buildjet.com/for-github-actions) to make it easier for us to create ARM64 stuff and test them.

Makefile
camilamacedo86: This PR is all about adding some tests. Volodya needs to build and test manually (very time consuming see;`I need a volunteer to test the binaries on Mac OS and Arm Linux.` ) what here has been done here automatically (all less mac/arm64);

Therefore, I’m thinking this PR could really speed things up and bring confidence regards his changes! So, I’m kinda wondering, why wait? How could this PR negatively impact this effort?

camilamacedo86: My first thought was we use the check to ensure the xgo stuff but it is totally fine.

.github/workflows/binaries.yml
camilamacedo86: I've looked at this option, and I have some reservations due to its cost implications. Since our repository is public, GitHub Actions are free, and our goal is to ensure the build and functionality of the binaries.

Considering the use of [buildjet](https://buildjet.com/for-github-actions) to replace QEMU, we may need to duplicate checks—one using GitHub runners for AMD and another with [buildjet](https://buildjet.com/for-github-actions) for ARM- that would not make things more simple I think. 

The main advantage of [buildjet](https://buildjet.com/for-github-actions) is that it runs on native ARM systems, which could be faster. However, all tests added here don't take more than 5 minutes.

So, the question is whether the cost outweighs the benefits in this case.

vkhoroz:
@camilamacedo86 @StealthyCoder please, hold on this.

I already have a hard time with #269, and adding more friction to it is highly unwanted at this time.
We have a `.jobserv` runner which already validates the build.
So, I prefer to keep it open for the XGo based build if we want to migrate to using GitHub actions or keep using JobServ.

Once, all other PRs impacting our build process are merged and show good results, we might return back to this.

camilamacedo86:
Hi @vkhoroz,

Thank you for sharing your insights; I value your perspective. Here are my thoughts on the points raised:

> Our other projects (like e.g. fioconfig) are tied to Golang 1.19 at the moment. 
> Having different projects at different versions makes sense sometimes, but there should be a strong motivation for doing that.

Considering Golang’s notable stability, is there a reason not to keep all projects updated? The introduction of tools like [Dependabot](https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/) has considerably eased the process of managing updates and addressing CVEs, assuming there is basic coverage in place.

> On practice, Golang upgrades are not always smooth, sometimes they lead to subtle bugs, which are not always visible on surface.

Could you possibly share an example of this? I haven’t yet come across an instance where Golang’s stability led to a breaking change that wasn’t caught during the build. Any illustration of a scenario where a change in Golang might cause malfunction while still building successfully would be enlightening.

> For the minimum, I already see [this code friction](https://github.com/foundriesio/fioctl/pull/272/files#r1335793568) related to the linter upgrade, which could be avoided.

I see things a bit differently on this point. The referenced PR proposes adding a check to highlight deprecations, which we are currently overlooking. Addressing deprecations promptly is pivotal for maintaining a project effectively. The eventual need to replace deprecated methods is unavoidable, and early awareness allows for smoother transitions, preventing the myriad of issues that can stem from delayed updates.

> This is not a Chore, the Golang upgrade requires full regression testing.

**I’m curious to understand in which scenarios a Golang upgrade would introduce issues not identified during the build. Could you provide an example or elaborate a bit more on this?**

Thank you for considering my queries; I look forward to furthering our discussion on this matter.

camilamacedo86:
Hi @vkhoroz,

> I already have a hard time with https://github.com/foundriesio/fioctl/pull/269, and adding more friction to it is highly unwanted at this time.

This PR introduces tests to verify that the tool can be built and is functional across all supported architectures. Could you please elaborate on how having test coverage could adversely impact the feature you are adding? Additionally, aren't these the exact tests you are conducting manually and seeking assistance with, to ensure the binary functions correctly post-modifications? Wouldn't automating these tests facilitate a more efficient development process? 

> We have a .jobserv runner which already validates the build.

Could you please indicate where this is in the logs? I can only see golint being called. 

> So, I prefer to keep it open for the XGo based build

This PR triggers the make targets for the build, executing it as defined by them. Could you please help me understand how an XGo-based build is conducted? How does that process differ? Aren't those builds still initiated when the targets are called?

Also, where are the implementation of those tests?
How keep them maintained?

.github/workflows/build.yml
doanac: the container method was a drive by contribution from someone that didn't have golang on their machine.

doanac: its not how we build release artificates

Makefile
doanac: we've agreed in a different PR to drop armv7. We've never included it in a release artifact.

.github/workflows/binaries.yml
doanac: We have a CI system we've built and are already paying for that excels at building for arm. Speaking from experience with the release ci job - github is fast enough.

Makefile
camilamacedo86: We can just remove this build/check

.github/workflows/binaries.yml
camilamacedo86: So, I think we can convey in keep GitHub

camilamacedo86:
Hi @doanac, @vkhoroz, @StealthyCoder,

Based on the reviews:

- I have removed the test for arm7.
- I have also excluded the tests related to the Dockerfile. It seems to me that we could remove it, but we will address that in a separate PR.

Consequently, this PR now solely focuses on adding builds/checks for the supported arcs, excluding Darwin/arm64, as it is not supported by GitHub. Since this PR is exclusively about adding tests, and assuming no one has objections to the added tests, I believe it is ready to be merged. It should not impact any existing implementation.

Please let me know if there are any concerns or further adjustments needed.
Thank you for your tim.

.github/workflows/build.yml
camilamacedo86: if no longer useful we could remove it right?

.github/workflows/binaries.yml
camilamacedo86: removed

.github/workflows/binaries.yml
camilamacedo86: closing as sorted out.

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,118 @@
...
+        with:
+          platforms: ${{ matrix.qemu-platform }}
+
+      - name: Build Binary
+        run: |
+          if [ "${{ matrix.binary }}" == "fioctl-windows-amd64.exe" ]; then
+            make fioctl-windows-amd64
+          else
+            make ${{ matrix.binary }}
+          fi
```
StealthyCoder, line 41: If we run in `Bash`, can we do something like the following? :

```
binary="${{ matrix.binary }}"
make ${binary%.exe}
```

On the off chance we ever get other windows binaries or rename them. This just chops off the `.exe` part, but it is a `bashism`

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,118 @@
...
+
+      - name: Install QEMU user static
+        if: matrix.qemu
+        run: sudo apt-get update && sudo apt-get install -y qemu-user-static
+        shell: bash
+
+      - name: Should run successfully
+        run: |
+          if ($env:RUNNER_OS -eq "Windows") {
+            .\bin\${{ matrix.binary }} logout
```
StealthyCoder, line 91: I do not think powershell cares about the slashes. So you can drop this particular line. I tested it on PowerShell in Windows 10.

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,118 @@
...
+            $exitCode = $LASTEXITCODE
+          } else {
+            $output = ./bin/${{ matrix.binary }} targets list 2>&1
+            $exitCode = $LASTEXITCODE
+          }
+          if ($exitCode -eq 1 -and $output -match 'ERROR: Please run: "fioctl login" first') {
+            Write-Host "Expected error occurred"
+            $host.SetShouldExit(0)
+          } else {
+            Write-Error "Unexpected output or exit code: $output, Exit Code: $exitCode"
```
StealthyCoder, line 116: will the `$exitCode` be propagated out to the reported failure? As in, let us say the `$LASTEXITCODE` is 1337. Then `$exitCode` is also 1337, will the GitHub Action report failed with exit code 1337? If so, then we do not have to put it in this message.

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,118 @@
...
+            $output = .\bin\${{ matrix.binary }} targets list 2>&1
+            $exitCode = $LASTEXITCODE
+          } elseif ("${{ matrix.qemu }}" -eq "true") {
+            $output = qemu-${{ matrix.qemu-platform }}-static ./bin/${{ matrix.binary }} targets list 2>&1
+            $exitCode = $LASTEXITCODE
+          } else {
+            $output = ./bin/${{ matrix.binary }} targets list 2>&1
+            $exitCode = $LASTEXITCODE
+          }
+          if ($exitCode -eq 1 -and $output -match 'ERROR: Please run: "fioctl login" first') {
```
StealthyCoder, line 112: Could we have a `contains`, rather than a match? Maybe we introduce color output later (therefore it will contain ANSI Escape codes) on or change our messages slightly to:
```
ERROR:
Please run "fioctl login" first.
```

Then this will fail.

StealthyCoder (commented):
Only have small nitpick questions and suggestions.

.github/workflows/binaries.yml
doanac: Does running a binary really give us anything? i.e. has there ever been a case where golang produces binaries that can't run? It might - I'm not familiar with this stuff. But I'd weigh the likelihood of these binaries not running against the likelihood of the logout message changing and causing our unit test to fail

.github/workflows/binaries.yml
camilamacedo86: I am looking at : https://github.com/foundriesio/fioctl/blob/main/.github/workflows/release.yml#L25-L27
Are we using this one to do the releases right?

So, I think we should only call make and not build in the matrix. 
It should work since we are working with cross-compilation.

camilamacedo86 (commented):
Hi @vkhoroz,


> We have a .jobserv runner which already validates the build.

I found the test:  https://github.com/foundriesio/fioctl/blob/main/.jobserv.yml#L28-L48
When I checked the logs I did not see that we were doing the builds. 
thank you for pointing out.

.github/workflows/binaries.yml
camilamacedo86: I changed that

.github/workflows/binaries.yml
camilamacedo86: We have a test that is calling the binary already so if the build fails I understand that should fail.
In this way, I think we can drop this one. KISS

.github/workflows/binaries.yml
camilamacedo86: removed thx

PR #277, opened 2023-09-26
PoC: complete XGo based build solution with HSM support

vkhoroz:
This PoC utilizes a Bash script which allowed me to focus on the implementation rather than the fancy Makefile syntax.
It is a full solution in one commit, so that you can easily compare it to the https://github.com/foundriesio/fioctl/pull/269.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,28 +1,54 @@
 COMMIT?=$(shell git describe HEAD)$(shell git diff --quiet || echo '+dirty')
 
 # Use linker flags to provide commit info
-LDFLAGS=-ldflags "-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)"
+VERSION_LDFLAGS=-X=github.com/foundriesio/fioctl/subcommands/version.Commit=$(COMMIT)
+BUILD_STATIC=1
+PREFIX=bin/fioctl
 
 linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-lint)
+builder:=$(shell which xgo 2>/dev/null || echo $(HOME)/go/bin/xgo)
+xgopkg::=src.techknowlogick.com/xgo@v1.7.0+1.19.5
+xgoimg::=ghcr.io/crazy-max/xgo:1.21.0
 
+# Platforms listed in build and build-hsm targets are fully supported and tested.
+# See the README for a full list of platforms tentatively supported by our XGo cross-compiler toolchain.
+
+# Statically linked binary without HSM support on platforms that support static binaries.
+# On platforms not supporting static binaries (e.g. Darwin) this builds dynamic library without HSM support.
 build: fioctl-linux-amd64 fioctl-linux-arm64 fioctl-windows-amd64 fioctl-darwin-amd64 fioctl-darwin-arm64
 	@true
 
-fioctl-static:
-	CGO_ENABLED=0 go build -a -ldflags '-w -extldflags "-static"' -o ./bin/fioctl-static ./main.go
+# Dynamically linked binary with HSM support.
+# There is no way to build static binary with HSM support due to the PKCS11 design choice to use dlopen.
+# For Darwin and Windows we link against the lowest possible library version for the greatest compatibility.
+# For Linux we link against GNU Libc 6; so it must be installed on a target system.
+build-hsm: fioctl-hsm-darwin-amd64 fioctl-hsm-darwin-arm64 fioctl-hsm-windows-amd64 fioctl-hsm-libc6-linux-amd64
+	@true
 
-fioctl-linux-amd64:
-fioctl-linux-arm64:
-fioctl-linux-armv7:
-fioctl-windows-amd64:
-fioctl-darwin-amd64:
-fioctl-darwin-arm64:
-fioctl-%:
-	CGO_ENABLED=0 \
-	GOOS=$(shell echo $* | cut -f1 -d\- ) \
-	GOARCH=$(shell echo $* | cut -f2 -d\-) \
-		go build $(LDFLAGS) -o bin/$@ main.go
-	@if [ "$@" = "fioctl-windows-amd64" ]; then mv bin/$@ bin/$@.exe; fi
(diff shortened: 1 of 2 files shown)


vkhoroz:
A way to go forward is this:
- Andre will verify that it works well with https://github.com/foundriesio/fioctl/pull/270 when enabling on Windows/Linux/maybe Darwin (whichever platforms available).
- Once/if it works well - Andre disables the HSM support and merges that PR.
- Then I work on this PR to break down into individual nice commits, and add the missing pieces for the CI.
- When all pieces are captured into a nice puzzle - we do a final review of the clean PR.
- Before the merge we need to split the regression testing effort based on our platforms and time availability.

Makefile
```diff
@@ -1,28 +1,54 @@
...
+# For Linux we link against GNU Libc 6; so it must be installed on a target system.
+build-hsm: fioctl-hsm-darwin-amd64 fioctl-hsm-darwin-arm64 fioctl-hsm-windows-amd64 fioctl-hsm-libc6-linux-amd64
+	@true
+
+# Below 3 targets allow you to build the local binary using your host toolchain
+fioctl-local-dynlink:
+	go build -ldflags '-s -w' -o ./bin/fioctl-local-dynlink ./main.go
+
+fioctl-hsm-local-dynlink:
+	go build -tags=hsm -ldflags '-s -w' -o ./bin/fioctl-hsm-local-dynlink ./main.go
```
StealthyCoder, line 33: Does this one also need `dynamic` added to the tags in order for it to trigger the hsm dynamic on windows? I base this on #270 .

build.sh
```diff
@@ -0,0 +1,58 @@
...
+    TARGET_PREFIX=${TARGET_PREFIX}-libc6
+fi
+
+if test "x${TARGET_STATIC}" = "x1"; then
+    # Darwin officially disallows building static binaries.
+    # Even internal linker always dynlinks libSystem.B.dylib, libresolv.9.dylib, CoreFoundation, and Security.
+    test "x${GOOS}" = "xwindows" && TARGET_LDFLAGS="'-extldflags=-static -O1'"
+    test "x${GOOS}" = "xlinux" && TARGET_LDFLAGS="'-extldflags=-static -O1'"
+    if test "x${GOOS}-${GOARCH}" = "xlinux-amd64"; then
+        # For Amd64 Linux add binary hardening using PIE and Full RELRO.
```
StealthyCoder, line 40: But this one does not use `-z relro -z now`, is that correct?

Makefile
vkhoroz: This is intended.

My initial idea was to prevent accidental build of a static binary with HSM support (which will certainly either fail or produce a corrupt binary).

But, this subject is more convoluted than I thought initially.
For example, there is no such thing as static binary on Darwin and few other platforms.
Which brings me to the idea that the whole "static/dynamic" thinking is kind of narrow-minded in our non-binary world.

So, it is easier to have just `hsm` tag to control if HSM support is being built.

build.sh
vkhoroz: This all is so convoluted (again).
The `-static-pie` flag does the magic to apply both PIE and Full RELRO on top of Static.
On practice, whatever `-z` flags I tried to pass in were ignored in this case; and as usual there is little to no documentation on the subject.

StealthyCoder (approved):

mike-sul:
Maybe we should just extend the existing Makefile with new targets(s) that invoke the script, and the script just builds the non-static HSM targets.
I mean leave Makefile for the existing/current target building as it is, just extend it with the HSP specifics.

mike-sul:
When I run `make fioctl-linux-amd64` I get
```
CGO_ENABLED=0 \
GOOS=linux \
GOARCH=amd64 \
	go build -buildmode=pie -ldflags "-X=github.com/foundriesio/fioctl/subcommands/version.Commit=v0.34-2-g8a8948c" -o bin/fioctl-linux-amd64 main.go

```

With the new build system when I run `make fioctl-linux-amd64` I get

```
Bulding for linux-amd64
INFO: Starting xgo/dev
flag provided but not defined: -image
Usage of /work/go/bin/xgo:
  -branch string
    	Version control branch to build
  -buildmode string
    	Indicates which kind of object file to build (default "default")
  -buildvcs string
    	Whether to stamp binaries with version control information
  -deps string
    	CGO dependencies (configure/make based archives)
  -depsargs string
    	CGO dependency configure arguments
  -dest string
    	Destination folder to put binaries in (empty = current)
  -docker-image string
    	Use custom docker image instead of official distribution
  -docker-repo string
    	Use custom docker repo instead of official distribution
  -go string
    	Go release to use for cross compilation (default "latest")
  -goproxy string
    	Set a Global Proxy for Go Modules
  -ldflags string
    	Arguments to pass on each go tool link invocation
  -out string
    	Prefix to use for output naming (empty = package name)
  -pkg string
    	Sub-package to build if not root import
  -race
    	Enable data race detection (supported only on amd64)
  -remote string
    	Version control remote repository to build
  -tags string
    	List of build tags to consider satisfied during the build
  -targets string
    	Comma separated targets to build for (default "*/*")
  -trimpath
    	Remove all file system paths from the resulting executable
  -v	Print the names of packages as they are compiled
  -x	Print the command as executing the builds
make: *** [Makefile:39: fioctl-linux-amd64] Error 2

```

vkhoroz:
@mike-sul probably you tried to build using https://github.com/foundriesio/fioctl/pull/269 earlier.
In this PR I used a different flavor of `xgo` which allows to better customize the docker arguments.

Please, run `make install-builder` and retry.

mike-sul:
> @mike-sul probably you tried to build using #269 earlier.
> In this PR I used a different flavor of `xgo` which allows to better customize the docker arguments.

```
On branch vkhoroz-xgo-toolchain-with-bash
Your branch is up to date with 'origin/vkhoroz-xgo-toolchain-with-bash'.

```

vkhoroz:
Tabling this as we agreed to implement the HSM support in a way which does not require cross-compilation using CGo: https://github.com/foundriesio/fioctl/pull/278

PR #278, opened 2023-09-27
WIP: pkcs11-tool based HSM support

vkhoroz:
An idea is to continue building Fioctl statically, and provide HSM support by wrapping pkcs11-tool.
The tool is available for Linux, Windows, and Darwin; a user can also build from source.

The PR contains 3 implementations:
- Our current/legacy Bash based PKI, which is now turned off. The HSM support for it was fixed, and script running made more secure.
- The Golang based implementation was enhanced with pkcs11-tool to support HSM devices, and turned on for all platforms.
- An alternative native Golang implementation via CGo package is also added, but turned off by default as it only works for a dynamically linked (not static) binary.

I find it reasonable to keep all 3 implementations for reference.

### DIFF
--- a/.github/workflows/test-pki.yml
+++ b/.github/workflows/test-pki.yml
@@ -1,0 +1,19 @@
+name: Test PKI
+on:
+  pull_request:
+
+jobs:
+  test-pki:
+    name: test-pki
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@v4.1.0
+        with:
+          go-version: '~1.19'
+      - name: Clone the code
+        uses: actions/checkout@v4
+      - name: Setup test deps
+        run: sudo make install-test-pki-deps
+      - name: Run tests
+        run: make test-pki
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ linter-check: has-linter
 	$(linter) run ${EXTRA_LINTER_FLAGS}
 	$(linter) run --build-tags bashpki ${EXTRA_LINTER_FLAGS}
 	$(linter) run --build-tags cgopki ${EXTRA_LINTER_FLAGS}
+	$(linter) run --build-tags testhsm ${EXTRA_LINTER_FLAGS}
 
 linter: has-linter
 	$(linter) run --fix ${EXTRA_LINTER_FLAGS}
@@ -51,6 +52,15 @@ format:
 
 check: format-check linter-check
 	@true
+
+install-test-pki-deps:
+	apt install openssl softhsm2 opensc libengine-pkcs11-openssl
+
+# This needs the following packages on Ubuntu: openssl softhsm2 opensc libengine-pkcs11-openssl
+test-pki:
+	go test ./x509/... -v -tags testhsm
+	go test ./x509/... -v -tags testhsm,bashpki
+	go test ./x509/... -v -tags testhsm,cgopki
 
 # Use the image for Dockerfile.build to build and install the tool.
 container-init:
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,10 @@ build: fioctl-linux-amd64 fioctl-linux-arm64 fioctl-windows-amd64 fioctl-darwin-
 fioctl-static:
 	CGO_ENABLED=0 go build -a -ldflags '-w -extldflags "-static"' -o ./bin/fioctl-static ./main.go
 
+# Allows building a dyn-linked fioctl on platforms without pkcs11-tool (not built by default)
+fioctl-cgo-pkcs11:
+	CGO_ENABLED=1 go build -tags cgopki $(LDFLAGS) -o bin/$@ ./main.go
+
 fioctl-linux-amd64:
 fioctl-linux-arm64:
 fioctl-linux-armv7:
@@ -34,6 +38,7 @@ has-linter:
 linter-check: has-linter
 	$(linter) run ${EXTRA_LINTER_FLAGS}
 	$(linter) run --build-tags bashpki ${EXTRA_LINTER_FLAGS}
+	$(linter) run --build-tags cgopki ${EXTRA_LINTER_FLAGS}
 
 linter: has-linter
 	$(linter) run --fix ${EXTRA_LINTER_FLAGS}
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,7 @@ has-linter:
 
 linter-check: has-linter
 	$(linter) run ${EXTRA_LINTER_FLAGS}
+	$(linter) run --build-tags bashpki ${EXTRA_LINTER_FLAGS}
 
 linter: has-linter
 	$(linter) run --fix ${EXTRA_LINTER_FLAGS}
--- a/client/foundries_pki.go
+++ b/client/foundries_pki.go
@@ -19,11 +19,6 @@ type CaCerts struct {
 	TlsCsr  string `json:"tls-csr"`
 
 	ChangeMeta ChangeMeta `json:"change-meta"`
-
-	CreateCaScript       *string `json:"create_ca"`
-	CreateDeviceCaScript *string `json:"create_device_ca"`
-	SignCaScript         *string `json:"sign_ca_csr"`
-	SignTlsScript        *string `json:"sign_tls_csr"`
 }
 
 func (a *Api) FactoryGetCA(factory string) (CaCerts, error) {
--- a/subcommands/keys/ca_create.go
+++ b/subcommands/keys/ca_create.go
@@ -95,13 +95,6 @@ func doCreateCA(cmd *cobra.Command, args []string) {
 	resp, err := api.FactoryCreateCA(factory)
 	subcommands.DieNotNil(err)
 
-	writeFile(x509.OnlineCaCsrFile, resp.CaCsr, 0400)
-	writeFile(x509.TlsCsrFile, resp.TlsCsr, 0400)
-	writeFile(x509.CreateCaScript, *resp.CreateCaScript, 0700)
-	writeFile(x509.CreateDeviceCaScript, *resp.CreateDeviceCaScript, 0700)
-	writeFile(x509.SignCaScript, *resp.SignCaScript, 0700)
-	writeFile(x509.SignTlsScript, *resp.SignTlsScript, 0700)
-
 	fmt.Println("Creating offline root CA for Factory")
 	resp.RootCrt = x509.CreateFactoryCa(factory)
 
@@ -121,7 +114,9 @@ func doCreateCA(cmd *cobra.Command, args []string) {
 			resp.CaCrt += "\n"
 		}
 		commonName := getDeviceCaCommonName(factory)
-		resp.CaCrt += x509.CreateDeviceCa(commonName, factory)
+		deviceCaCrt := x509.CreateDeviceCa(commonName, factory)
+		writeFile(x509.DeviceCaCertFile, deviceCaCrt, 0400)
+		resp.CaCrt += deviceCaCrt
 	}
 
 	fmt.Println("Uploading signed certs to Foundries")
--- a/subcommands/keys/ca_create.go
+++ b/subcommands/keys/ca_create.go
@@ -1,6 +1,7 @@
 package keys
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
@@ -84,12 +85,13 @@ func doCreateCA(cmd *cobra.Command, args []string) {
 
 	if len(hsmModule) > 0 {
 		if len(hsmPin) == 0 {
-			fmt.Println("ERROR: --hsm-pin is required with --hsm-module")
-			os.Exit(1)
+			subcommands.DieNotNil(errors.New("--hsm-pin is required with --hsm-module"))
 		}
-		os.Setenv("HSM_MODULE", hsmModule)
-		os.Setenv("HSM_PIN", hsmPin)
-		os.Setenv("HSM_TOKEN_LABEL", hsmTokenLabel)
+		x509.InitHsm(x509.HsmInfo{
+			Module:     hsmModule,
+			Pin:        hsmPin,
+			TokenLabel: hsmTokenLabel,
+		})
 	}
 
 	resp, err := api.FactoryCreateCA(factory)
--- a/subcommands/keys/ca_create.go
+++ b/subcommands/keys/ca_create.go
@@ -61,7 +61,7 @@ This is optional.`,
 	cmd.Flags().BoolVarP(&createLocalCA, "local-ca", "", true, "Create a local CA that you can use for signing your own device certificates")
 	cmd.Flags().StringVarP(&hsmModule, "hsm-module", "", "", "Create key on an PKCS#11 compatible HSM using this module")
 	cmd.Flags().StringVarP(&hsmPin, "hsm-pin", "", "", "The PKCS#11 PIN to set up on the HSM, if using one")
-	cmd.Flags().StringVarP(&hsmTokenLabel, "hsm-token-label", "", "device-gateway-root", "The label of the HSM token created for this")
+	cmd.Flags().StringVarP(&hsmTokenLabel, "hsm-token-label", "", "", "The label of the HSM token created for this")
 }
 
 func getDeviceCaCommonName(factory string) string {
@@ -79,6 +79,9 @@ func doCreateCA(cmd *cobra.Command, args []string) {
 	if len(hsmModule) > 0 {
 		if len(hsmPin) == 0 {
 			subcommands.DieNotNil(errors.New("--hsm-pin is required with --hsm-module"))
+		}
+		if len(hsmTokenLabel) == 0 {
+			subcommands.DieNotNil(errors.New("--hsm-token-label is required with --hsm-module"))
 		}
 		x509.InitHsm(x509.HsmInfo{
 			Module:     hsmModule,
--- a/subcommands/keys/ca_create.go
+++ b/subcommands/keys/ca_create.go
@@ -64,13 +64,6 @@ This is optional.`,
 	cmd.Flags().StringVarP(&hsmTokenLabel, "hsm-token-label", "", "device-gateway-root", "The label of the HSM token created for this")
 }
 
-func writeFile(filename, contents string, mode os.FileMode) {
-	if err := os.WriteFile(filename, []byte(contents), mode); err != nil {
-		fmt.Printf("ERROR: Creating %s: %s", filename, err)
-		os.Exit(1)
-	}
-}
-
 func getDeviceCaCommonName(factory string) string {
 	user, err := api.UserAccessDetails(factory, "self")
 	subcommands.DieNotNil(err)
@@ -102,12 +95,10 @@ func doCreateCA(cmd *cobra.Command, args []string) {
 
 	fmt.Println("Signing Foundries TLS CSR")
 	resp.TlsCrt = x509.SignTlsCsr(resp.TlsCsr)
-	writeFile(x509.TlsCertFile, resp.TlsCrt, 0400)
 
 	if createOnlineCA {
 		fmt.Println("Signing Foundries CSR for online use")
 		resp.CaCrt = x509.SignCaCsr(resp.CaCsr)
-		writeFile(x509.OnlineCaCertFile, resp.CaCrt, 0400)
 	}
 
 	if createLocalCA {
@@ -116,9 +107,7 @@ func doCreateCA(cmd *cobra.Command, args []string) {
 			resp.CaCrt += "\n"
 		}
 		commonName := getDeviceCaCommonName(factory)
-		deviceCaCrt := x509.CreateDeviceCa(commonName, factory)
-		writeFile(x509.DeviceCaCertFile, deviceCaCrt, 0400)
-		resp.CaCrt += deviceCaCrt
+		resp.CaCrt += x509.CreateDeviceCa(commonName, factory)
 	}
 
 	fmt.Println("Uploading signed certs to Foundries")
--- a/x509/bash.go
+++ b/x509/bash.go
@@ -12,38 +12,190 @@ import (
 	"github.com/foundriesio/fioctl/subcommands"
 )
 
-func run(name string, arg ...string) string {
-	cmd := exec.Command(name, arg...)
+func run(script string, arg ...string) string {
+	arg = append([]string{"-s"}, arg...)
+	cmd := exec.Command("/bin/sh", arg...)
 	cmd.Stderr = os.Stderr
+	in, err := cmd.StdinPipe()
+	subcommands.DieNotNil(err, "Failed to start the shell")
+
+	go func() {
+		defer in.Close()
+		_, err := in.Write([]byte(script))
+		subcommands.DieNotNil(err, "Failed to pass the script to the shell")
+	}()
+
 	out, err := cmd.Output()
-	subcommands.DieNotNil(err)
+	subcommands.DieNotNil(err, "Failed to execute the shell script")
 	return string(out)
 }
 
 func CreateFactoryCa(ou string) string {
-	run("./" + CreateCaScript)
+	const script = `#!/bin/sh -e
+## This script creates the offline private key, factory_ca.key, and x509 certficate, factory_ca.pem,
+## owned by the customer that provides a chain of trust for all other certficates used by this factory.
+
+if [ $# -ne 1 ] ; then
+	echo "ERROR: $0 <ou>"
+	exit 1
+fi
+
+ou=$1
+
+cat >ca.cnf <<EOF
+[req]
+prompt = no
+distinguished_name = dn
+x509_extensions = ext
+
+[dn]
+CN = Factory-CA
+OU = ${ou}
+
+[ext]
+basicConstraints=CA:TRUE
+keyUsage = keyCertSign
+extendedKeyUsage = critical, clientAuth, serverAuth
+EOF
+
+if [ -n "$HSM_MODULE" ] ; then
+	lbl=${HSM_TOKEN_LABEL-device-gateway-root}
+	pkcs11-tool --module $HSM_MODULE \
+		--keypairgen --key-type EC:prime256v1 \
+		--token-label $lbl \
+		--id 01 \
+		--label root-ca \
+		--pin $HSM_PIN
+
+	key="pkcs11:token=${lbl};object=root-ca;type=private;pin-value=$HSM_PIN"
+	extra="-engine pkcs11 -keyform engine"
+else
+	openssl ecparam -genkey -name prime256v1 | openssl ec -out factory_ca.key
+	key=factory_ca.key
+fi
+openssl req $extra -new -x509 -days 7300 -config ca.cnf -key "$key" -out factory_ca.pem -sha256
+rm ca.cnf
+`
+	run(script, ou)
 	return readFile(FactoryCaCertFile)
 }
 
 func CreateDeviceCa(cn string, ou string) string {
-	run("./"+CreateDeviceCaScript, DeviceCaKeyFile, DeviceCaCertFile)
-	return readFile(DeviceCaCertFile)
+	const script = `#!/bin/sh -e
+## This is an optional script a customer can use to create a certificate
+## capable of signing a certicate signing request from an LMP device.
+## The certicate created here will be trusted by the Foundries device gateway.
+## This is useful for creating CA owned by the customer for use in a manufacturing facility.
+
+if [ $# -ne 3 ] ; then
+	echo "ERROR: $0 <key-create> <cn> <ou>"
+	exit 1
+fi
+key=$1
+cn=$2
+ou=$3
+
+cat >ca.cnf <<EOF
+[req]
+prompt = no
+distinguished_name = dn
+x509_extensions = ext
+
+[dn]
+CN = ${cn}
+OU = ${ou}
+
+[ext]
+keyUsage=critical, keyCertSign
+basicConstraints=critical, CA:TRUE, pathlen:0
+EOF
+
+openssl ecparam -genkey -name prime256v1 | openssl ec -out $key
+openssl req -new -config ca.cnf -key $key
+rm ca.cnf`
+	csrPem := run(script, DeviceCaKeyFile, cn, ou)
+	return signCaCsr("device-ca-*", csrPem)
 }
 
 func SignTlsCsr(csrPem string) string {
-	return run("./"+SignTlsScript, TlsCsrFile)
+	const script = `#!/bin/sh -e
+## This script signs the "tls-csr" returned when creating Factory certificates.
+## This certificate are signed so that the devices trust the TLS connection with the Foundries device gateway.
+
+if [ $# -ne 2 ] ; then
+	echo "ERROR: $0 <tls csr> <tls crt>"
+	exit 1
+fi
+csr=$1
+crt=$2
+
+dns=$(openssl req -text -noout -verify -in $csr | grep DNS:)
+echo "signing with dns name: $dns" 1>&2
+
+cat >server.ext <<EOF
+keyUsage=critical, digitalSignature, keyEncipherment, keyAgreement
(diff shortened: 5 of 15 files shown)


vkhoroz:
I took a part of the refactoring made by @detsch from https://github.com/foundriesio/fioctl/pull/270 and replaced the HSM implementation with calls to pkcs11-tool.

The result is a working solution as confirmed by a unit test, which is able to generate a valid certificate.
It is not at all complete and needs to be enhanced a lot.
But, it saves our lives from the CGo cross-compilation nightmare and a necessity to provide separate binaries for HSM.

Steps to test on Ubuntu:
```
# Initialize environment
sudo apt install softhsm2 opensc
mkdir -p ${HOME}/.softhsm/tokens
echo "directories.tokendir = ${HOME}/.softhsm/tokens" > ~/.softhsm/softhsm2.conf
export SOFTHSM2_CONF=$HOME/.softhsm/softhsm2.conf
softhsm2-util --init-token --slot 0 --label mine --so-pin=1234 --pin=1234

# Run tests
go test ./x509 -v

# Verify results
pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --token-label mine --id 1 --label root-ca --pin 1234 --list-objects
openssl x509 -in x509/factory_ca.pem -text

# clean env (before re-running the test, otherwise it fails)
pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --token-label mine --id 1 --label root-ca --pin 1234 --delete-object --type pubkey
pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --token-label mine --id 1 --label root-ca --pin 1234 --delete-object --type privkey
rm -f x509/factory_ca.pem
```

vkhoroz:
Test produced a valid certificate based on softhsm keys for me locally:
```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            fd:84:6b:59:3f:96:64:98:62:2b:e9:bc:01:a7:4c:d4:92:78:08:76
        Signature Algorithm: ecdsa-with-SHA256
        Issuer: CN = Factory-CA, OU = test
        Validity
            Not Before: Sep 27 19:04:53 2023 GMT
            Not After : Sep 27 19:04:53 2043 GMT
        Subject: CN = Factory-CA, OU = test
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                pub:
                    04:63:7d:93:1c:a9:99:50:68:db:da:79:3c:f2:1f:
                    ca:72:6c:20:da:64:61:48:94:33:f5:3c:a0:ca:81:
                    f0:d7:11:6a:87:50:05:1d:21:c5:6c:2e:30:b4:8c:
                    b7:ba:b8:74:fe:ab:6a:9f:2f:5b:57:e4:f8:50:25:
                    23:49:bc:0f:af
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        X509v3 extensions:
            X509v3 Key Usage: critical
                Certificate Sign
            X509v3 Extended Key Usage: 
                TLS Web Server Authentication, TLS Web Client Authentication
            X509v3 Basic Constraints: critical
                CA:TRUE
            X509v3 Subject Key Identifier: 
                90:0D:83:D7:F3:98:EC:D2:A9:5F:8B:35:83:1E:A7:F3:5D:C6:83:E9
    Signature Algorithm: ecdsa-with-SHA256
    Signature Value:
        30:44:02:20:25:77:68:89:d2:d1:6e:ad:42:9d:54:b3:1f:26:
        db:b5:fa:af:20:41:f9:94:77:70:a3:b1:93:b4:7f:65:bf:db:
        02:20:37:41:d6:ff:de:03:51:41:cd:af:16:75:80:e5:c6:67:
        7d:a7:3a:19:f7:c4:d6:34:c0:aa:20:7e:90:f5:f9:f9
-----BEGIN CERTIFICATE-----
MIIBqzCCAVKgAwIBAgIVAP2Ea1k/lmSYYivpvAGnTNSSeAh2MAoGCCqGSM49BAMC
MCQxEzARBgNVBAMMCkZhY3RvcnktQ0ExDTALBgNVBAsMBHRlc3QwHhcNMjMwOTI3
MTkwNDUzWhcNNDMwOTI3MTkwNDUzWjAkMRMwEQYDVQQDDApGYWN0b3J5LUNBMQ0w
CwYDVQQLDAR0ZXN0MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEY32THKmZUGjb
2nk88h/Kcmwg2mRhSJQz9TygyoHw1xFqh1AFHSHFbC4wtIy3urh0/qtqny9bV+T4
UCUjSbwPr6NhMF8wDgYDVR0PAQH/BAQDAgIEMB0GA1UdJQQWMBQGCCsGAQUFBwMB
BggrBgEFBQcDAjAPBgNVHRMBAf8EBTADAQH/MB0GA1UdDgQWBBSQDYPX85js0qlf
izWDHqfzXcaD6TAKBggqhkjOPQQDAgNHADBEAiAld2iJ0tFurUKdVLMfJtu1+q8g
QfmUd3CjsZO0f2W/2wIgN0HW/94DUUHNrxZ1gOXGZ32nOhn3xNY0wKogfpD1+fk=
-----END CERTIFICATE-----

Public Key Object; EC  EC_POINT 256 bits
  EC_POINT:   044104637d931ca9995068dbda793cf21fca726c20da6461489433f53ca0ca81f0d7116a8750051d21c56c2e30b48cb7bab874feab6a9f2f5b57e4f850252349bc0faf
  EC_PARAMS:  06082a8648ce3d030107
  label:      root-ca
  ID:         01
  Usage:      encrypt, verify, wrap, derive
  Access:     local
Private Key Object; EC
  label:      root-ca
  ID:         01
  Usage:      decrypt, sign, unwrap, derive
  Access:     sensitive, always sensitive, never extractable, local
```

vkhoroz:
@mike-sul @doanac @detsch
If you like this idea, I will elaborate on it to make it fault-tolerant.

mike-sul:
> The tool is available for Linux, Windows, and Darwin; a user can also build from source.

If the tool is available on all required platforms then it seems like a good way forward. I thought about building our own tool around pkcs11/openssl, since suhc the tool already exists then why not.

vkhoroz:
> I thought about building our own tool around pkcs11/openssl, since suhc the tool already exists then why not.

What I find even better in this solution is that we don't need to wrap `openssl`.
We still use the Golang native X509 implementation, but offload a tiny fraction of ECDSA key generation and signing to `pkcs11-tool` to convey the crypto functions on the HSM.

Not as good as directly talking to the vendor specific PKCS11 library, but saves us a lot of cross-compilation headache.
At the end, from phylosophical perspective: talking to shared object versus talking to executable are ideologically almost equal.
Either way, we remove at least 2 dependencies:
- (Most important) Bash, so that we no longer can be acused of remote script execution.
- (Less important) OpenSSL, so that we are not subject to related CVEs.

detsch:
> @mike-sul @doanac @detsch If you like this idea, I will elaborate on it to make it fault-tolerant.

It seems to me this is the way to go. Code organization looks good too. I can help with some testing if you need.

vkhoroz:
@doanac @mike-sul @detsch @StealthyCoder @camilamacedo86 

Full testing will be done tomorrow, but this is kind of ready to go.

x509/common.go
```diff
@@ -20,3 +21,26 @@ func readFile(filename string) string {
...
+
+type hsmStorage struct {
+	HsmInfo
+	Label string
+}
+
+var factoryCaKeyStorage KeyStorage = &fileStorage{FactoryCaKeyFile}
+
+func InitHsm(hsm HsmInfo) {
+	if hsm.Module == "" {
```
doanac, line 43: this confuses me. we only call InitHSM if `len(hsmModule) > 0`. So how would this get set?

doanac (commented):
Is there any use keeping the bashpki? I don't think we'll ever test or ship it. So its basically just us carrying around unsupported code, right?

mike-sul:
Is this the correct command to store key/cert in HSM?

```
fioctl keys ca create $PWD/pki  --hsm-module /usr/lib/softhsm/libsofthsm2.so --hsm-pin 1234 --hsm-token-label mine
Creating offline root CA for Factory
Signing Foundries TLS CSR
Signing Foundries CSR for online use
Creating local device CA
Uploading signed certs to Foundries

```

vkhoroz:
Added unit tests and all fixes I found while testing.
I will continue end-2-end testing on Monday.

vkhoroz:
@mike-sul there was a bug introduced while I was shuffling code around.
Because of that HSM was always disabled.
It is fixed in the latest push.

vkhoroz:
@doanac while testing this stuff I found out quite a few discrepancies in the Golang based PKI implementation.
Surprisingly enough, the Bash based version was more accurate, even for the HSM use case; although it also had subtle issues.

That said, after I added unit tests which apply _exactly the same test flow_ to all 3 PKI implementations, I came to the conclusion that it is worth to keep all of them.
A reason is this: when we introduce some new code which breaks something, we could use test results to identify _which version to blame_ based on the quorum principle.
The PKI is complex, especially with the HSM support; so IMHO having only one implementation is like going blind in the wild.

Please, see the example test results:
https://github.com/foundriesio/fioctl/actions/runs/6355711636/job/17264165068?pr=278

Makefile
```diff
@@ -11,6 +11,9 @@ build: fioctl-linux-amd64 fioctl-linux-arm64 fioctl-windows-amd64 fioctl-darwin-
 fioctl-static:
 	CGO_ENABLED=0 go build -a -ldflags '-w -extldflags "-static"' -o ./bin/fioctl-static ./main.go
 
+fioctl-cgo-pkcs11:
+	CGO_ENABLED=1 go build -tags cgopki $(LDFLAGS) -o bin/$@ ./main.go
+
```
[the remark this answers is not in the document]
vkhoroz: See the test results here: https://github.com/foundriesio/fioctl/actions/runs/6355711636/job/17264165068?pr=278

And also my latest global comment which describes the rationale to keep all 3 implementations.
Just IMHO, the CGo based version is the easiest to maintain and understand.
So it could be used as a proof-of-correctness when we develop new features and tests are being added.

x509/common.go
vkhoroz: This was a bug, introduced while I shuffled the code between different options to implement this.

Fixed: https://github.com/foundriesio/fioctl/pull/278/commits/eb88423dfca9700fdff6102dc2b0fbbccb61b389

.github/workflows/test-pki.yml
```diff
@@ -0,0 +1,22 @@
...
+  test-pki:
+    name: test-pki
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@v4.1.0
+        with:
+          go-version: '~1.19'
+      - name: Setup test deps
+        uses: ConorMacBride/install-package@v1.1.0
```
camilamacedo86, line 16: ## Concerns about the Proposed GitHub Action

**Question:** Why are we considering using this particular action?

In our standard workflow configuration, we employ the following GitHub Actions:

\```yaml
steps:
  - name: Checkout code
    uses: actions/checkout@v2

  - name: Set up Go
    uses: actions/setup-go@v2
    with:
      go-version: '~1.19'
\```

I acknowledge the possibility of other GitHub Actions replicating the functionality of `actions/setup-go`, but I have certain reservations about adopting [ConorMacBride/install-package](https://github.com/ConorMacBride/install-package):

1. **Trust and Ownership:** The action is maintained by an individual without affiliation to a recognized organization or entity. This raises concerns about trustworthiness and accountability.
 
2. **Community Adoption:** A reliable indicator of a tool's dependability is its widespread use and endorsement by the developer community. This action does not seem to have garnered such widespread adoption.

3. **Sustainability Concerns:** When tools are maintained by individuals, there's an inherent risk of discontinuation. If the maintainer chooses to no longer update or support the action, we could be left with an unsupported tool in our workflow.

Given these points, I'd suggest sticking with the widely accepted `actions/setup-go` action unless there are overriding advantages in the alternative that we shouldn't overlook. Please, let's not use things that are not really adopted in the community.

vkhoroz: I'm not sure how `setup-go` can install apt packages.
Can you propose a solution?

I could surely run the plain script which calls `apt install $X`.
But, a ready to go action is better at maintaining the cache and resolving all specifics.
As for the _discontinuation_, I've seen two dozens of actions like this; not a concern for me.

That said, if the team prefers a script calling `apt install $X`, I'm fine with that.

camilamacedo86: > I'm not sure how setup-go can install apt packages.

So, if the goal is only install the packages into the Linux env I would prefer to see that done as it usually is (as the following example). I think my main reasons are:
- That makes clear what is required and what is doing so that I can either do manually if I need to do so and to test something (make easier we maintain this one)
- Do not rely on a GitHub action done by an individual that nobody uses or knows how to use.


Example:
```
name: My Workflow

on: [push]

jobs:
  build:
    runs-on: ubuntu-latest

    steps:
    - name: Checkout code
      uses: actions/checkout@v2

    - name: Install dependencies
      run: |
        sudo apt-get update
        sudo apt-get install -y <your-package-name>
```

mike-sul:
I suppose the question whether to allow storing both factory and local/offline CA keys in HSM, or just one of them in out of the scope of the given PR.

mike-sul:
I suppose getting of the factory CA key from HSM to sign a new device offline CA csr is out of the scope of this PR, and the same for device CSRs signing by an offline CA key stored in HSM?

mike-sul:
The happy path works on my host, the factory root CA priv key is stored in the HSM.

```
pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --token-label mine --id 1 --label root-ca --pin 1234 --list-objects
Public Key Object; EC  EC_POINT 256 bits
  EC_POINT:   04410494dea3234d6f29920e63fc9d6815b2053dd9552574017ae84bcc1e38c4bcd032379c749d9f1af99c6fe583db653c05a799dc65ff17b561a9fe9b40b479c9613a
  EC_PARAMS:  06082a8648ce3d030107
  label:      root-ca
  ID:         01
  Usage:      encrypt, verify, wrap, derive
  Access:     local
Private Key Object; EC
  label:      root-ca
  ID:         01
  Usage:      decrypt, sign, unwrap, derive
  Access:     sensitive, always sensitive, never extractable, local

```

mike-sul:
When I ran the `fioctl keys ca create --hsm-module...` for the first time I have my HSM miss-configured. As result, the certs/keys were created and the secret was stored on the backend, but the factory root CA priv key wasn't stored anywhere due to the write to HSM error.
If a customer run into similar issue then they will have to contact our support to reset their factory PKI in order to re-rerun the `fioctl keys ca create --hsm-module...` command.

I am not sure that this potential issue should be discussed in the scope of the given PR at all, just wanna mention for the future consideration.

Makefile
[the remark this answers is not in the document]
vkhoroz: We decided to keep it on the standup.

A rationale is this:
- Bash implementation is the most tested one, so it stays as a backup for a few months, before we scrap it.
- Golang + pkcs11-tool is our new default implementation, so it stays.
- Golang + ThalesIgnite is the easiest to prototype, so it stays to cross-verify the pkcs11-tool one.

.github/workflows/test-pki.yml
vkhoroz: https://github.com/foundriesio/fioctl/pull/278/commits/98a7d6bb2ea9a2c7b271f2f553e2fe7816c2a6cc

vkhoroz:
@mike-sul 

> I suppose $XYZ is out of scope.

Yes, the only things in scope are:
- Make sure that Bash-based implementation is fixed for the HSM use case.
- Achieve a feature parity between Golang-based and Bash-based implementations.
- Enable the Golang based implementation wrapping pkcs11-tool by default (by the build tag).
- I also voluntarily added unit tests to cross-verify that both implementations work for the happy path.

All other improvements need discussion, and risk growing this PR indefinitely.

> secret was stored on the backend, but the factory root CA priv key wasn't stored anywhere due to the write to HSM error

This sounds like a bug, the command should have failed.
I will test for this scenario tomorrow.

.github/workflows/test-pki.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Setup Go
+        uses: actions/setup-go@v4.1.0
+        with:
+          go-version: '~1.19'
+      - name: Clone the code
+        uses: actions/checkout@v4
+      - name: Setup test deps
+        run: sudo make install-test-pki-deps
+      - name: Run tests
+        run: make test-pki
```
camilamacedo86, line 20: That is cool 🥇

x509/common_test.go
```diff
@@ -0,0 +1,194 @@
...
+	key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	require.Nil(t, err)
+	csr := &x509.CertificateRequest{
+		Subject: pkix.Name{CommonName: testDnsGateway, OrganizationalUnit: []string{testFactory}},
+	}
+	csrDer, err := x509.CreateCertificateRequest(rand.Reader, csr, key)
+	require.Nil(t, err)
+	csrPem := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE REQUEST", Bytes: csrDer})
+	return key, string(csrPem)
+}
```
camilamacedo86, line 194: |oh| 🥇

mike-sul:
> > 
> 
> This sounds like a bug, the command should have failed. I will test for this scenario tomorrow.

```
fioctl keys ca create --hsm-module /usr/lib/softhsm/libsofthsm2.so --hsm-pin 1234 --hsm-token-label mine  $PWD -c $HOME/.config/fioctl-fioed.yaml -f msul-fioed-01 
Creating offline root CA for Factory
Main C_Initialize(NULL) rv:CKR_GENERAL_ERROR
error: PKCS11 function C_Initialize failed: rv = CKR_GENERAL_ERROR (0x5)
Aborting.
ERROR: exit status 1

```

```
ls -l
total 28
-rw-r--r-- 1 mike mike  212 Oct  3 10:31 ca.cnf
-r-------- 1 mike mike  534 Oct  3 10:31 ca-csr
-rwx------ 1 mike mike 1005 Oct  3 10:31 create_ca
-rwx------ 1 mike mike  837 Oct  3 10:31 create_device_ca
-rwx------ 1 mike mike  783 Oct  3 10:31 sign_ca_csr
-rwx------ 1 mike mike  735 Oct  3 10:31 sign_tls_csr
-r-------- 1 mike mike  615 Oct  3 10:31 tls-csr

```

```
kubectl -n ota-lite get  secret gateway-f978ad23-23f8-4835-804f-25e7fb43a89f -o json
{
    "apiVersion": "v1",
    "data": {
        "ca_key": "LS0tLS1CRUdJTiBFQyBQUklWQVRFIEtFWS0tLS0tClByb2MtVHlwZTogNCxFTkNSWVBURUQKREVLLUluZm86IEFFUy0yNTYtQ0JDLEExNjEwOEFBQUI2RjE4N0IyRTQxODg1M0U4RkMyNjA0CgoyS3N3dnU5MUNMMitBNi9NeU04RTN1ajlPVGdVdUovUERLdjM5eUU4dEFzZjljS01yRkpKbnY0SXlyM29QUWthCmo4OWNQZktOQk9DVFE1TEZyNDFRWnV6L2VpWmZpSFBpRjIyQ3pGMWhHOVdPdm9LUGJIdkQ5Nk9WZm5tbXg2ZzkKQlZjdkNGTnAzcVI1LzF0K1loa2ZLYjYwL2RKOXcvR1hxOEpCYkN1b3drUT0KLS0tLS1FTkQgRUMgUFJJVkFURSBLRVktLS0tLQo=",
        "change_meta": "eyJjcmVhdGVkX2F0IjogMTY5NjMyMTkwNy44NzQ1ODQ3LCAiY3JlYXRlZF9ieSI6ICI2NGFjMjM4MjJkM2JkMmU3MjUwNDU2OTMiLCAidXBkYXRlZF9hdCI6IDAsICJ1cGRhdGVkX2J5IjogIiJ9",
        "tls_key": "LS0tLS1CRUdJTiBFQyBQUklWQVRFIEtFWS0tLS0tClByb2MtVHlwZTogNCxFTkNSWVBURUQKREVLLUluZm86IEFFUy0yNTYtQ0JDLDREQkZFOTlCOTM5ODY3RTVFNkE4N0FBNkE2OUYyNjRECgppQUdnL01VZzFDM3N5dmdnemFyRk4rM2VlN3RGNlRpSFJMdzNNeHV2RlJweWtkMnd0Q2w1WEdxblJCcmliaUhtCnlaRXhWY2l2TXdLc0xPbERCVG1SVGdZcGVuYXlqa0kzKzdmemxCY1JvRElaRU1aVTNFMkJmSVIxeHpUcHNweXcKZmZrL3NjamxMSHdHdjkwN2w2Ykk3N1duQ1RXcmRFck1QRW53MVJYdnJZTT0KLS0tLS1FTkQgRUMgUFJJVkFURSBLRVktLS0tLQo="
    },
    "kind": "Secret",
    "metadata": {
        "creationTimestamp": "2023-10-03T08:31:47Z",
        "labels": {
            "infra.foundries.io/secret-type": "device-gateway-certs"
        },
        "name": "gateway-f978ad23-23f8-4835-804f-25e7fb43a89f",
        "namespace": "ota-lite",
        "resourceVersion": "87521349",
        "uid": "c5736476-2852-452a-b6de-fb132f6bd691"
    },
    "type": "Opaque"
}

```

client/foundries_pki.go
```diff
@@ -19,11 +19,6 @@ type CaCerts struct {
 	TlsCsr  string `json:"tls-csr"`
 
 	ChangeMeta ChangeMeta `json:"change-meta"`
-
-	CreateCaScript       *string `json:"create_ca"`
```
mike-sul, line 23: Is this require corresponding change in the PKI creation handler at ota-lite so it doesn't return those scripts?

subcommands/keys/ca_create.go
```diff
@@ -84,12 +85,13 @@ func doCreateCA(cmd *cobra.Command, args []string) {
 
 	if len(hsmModule) > 0 {
 		if len(hsmPin) == 0 {
-			fmt.Println("ERROR: --hsm-pin is required with --hsm-module")
-			os.Exit(1)
+			subcommands.DieNotNil(errors.New("--hsm-pin is required with --hsm-module"))
```
mike-sul, line 81: No need to check `hsmTokenLabel` value similarly?

x509/storage_pkcs11_tool.go
```diff
@@ -4,14 +4,101 @@ package x509
...
+	}
+	subcommands.DieNotNil(err)
+	key, err := x509.ParsePKIXPublicKey(out)
+	subcommands.DieNotNil(err)
+	return key
+}
+
+func (s *hsmSigner) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error) {
+	// By default pkcs11-tool returns raw signature, X509 needs it wrapped into the ASN.1 sequence
+	args := append(s.keyArgs(), "--sign", "--mechanism=ECDSA", "--signature-format=sequence")
```
mike-sul, line 60: Do we apply ECDSA unconditionally, or should it depend on the private key type?

x509/storage_pkcs11_tool.go
```diff
@@ -4,14 +4,101 @@ package x509
...
+	}()
+
+	return cmd.Output()
+}
+
+func genAndSaveKeyToHsm(hsm HsmInfo, id, label string) crypto.Signer {
+	// The pkcs11-tool allows creating many objects with the same ID and label, potentially corrupting the storage.
+	// For now, allow to create only one object.
+	// In the future we will use the ID field to allow key rotation.
+	key := &hsmSigner{hsm, id, label}
```
mike-sul, line 79: This construction `key := ..., if key.Public` looks a bit odd at first glance.

x509/storage_pkcs11_tool.go
```diff
@@ -4,14 +4,101 @@ package x509
...
+	if key.Public() != nil {
+		subcommands.DieNotNil(fmt.Errorf("Key %s already exists on the HSM device", label))
```
mike-sul, line 81: Do we wanna "die" in the scope of this function? Maybe just return `nil, err` and "die" in within context of the `genAndSaveKey`/`loadKey` caller?

x509/golang.go
```diff
@@ -52,7 +53,11 @@ func genCertificate(
 }
 
 func parsePemCertificateRequest(csrPem string) *x509.CertificateRequest {
-	pemBlock, _ := pem.Decode([]byte(csrPem))
+	pemBlock, rest := pem.Decode([]byte(csrPem))
+	if pemBlock == nil || len(rest) > 0 {
+		subcommands.DieNotNil(errors.New("Malformed PEM data"))
+		return nil
```
mike-sul, line 53: Do we need to return after "dying"?

Why are we "dying" inside of some low-level function, maybe it makes sense to return nil/err and die within the caller context?

x509/common.go
```diff
@@ -22,6 +22,11 @@ func readFile(filename string) string {
 	return string(data)
 }
 
+func writeFile(filename, contents string) {
+	err := os.WriteFile(filename, []byte(contents), 0400)
```
mike-sul, line 28: I suppose writing file atomically is too much in this context (I mean write to tmp file, fsync, and rename).

Makefile
```diff
@@ -52,6 +52,12 @@ format:
 check: format-check linter-check
 	@true
 
+# This needs the following packages on Ubuntu: openssl softhsm2 opensc libengine-pkcs11-openssl
+test-pki:
```
mike-sul, line 59: Maybe separate targets for each HSM type?
Also, I suppose, we can run the `-tags testhsm` type in CI?

.github/workflows/test-pki.yml
mike-sul: Ah, tests are triggered for all HSM types.

mike-sul (commented):
LGTM.

This is a quite big change. I've reviewed it without detailed analyses of the PKI keys/certs generation details for each implementation types.

I suppose we assume that the unit tests guarantee that the keys/certs are correct.

In any case, I believe it makes sense to test happy path e2e for the default case. I've just checked it partially. Testing of adding a new offline device CA and signing device CSR with the offline CA is gonna take some time.

client/foundries_pki.go
vkhoroz: Not yet.

The `ota-lite` needs to remain unmodified to support old clients.
Once the new version is released and proliferate to enough users (allowing enough time for that), the API code will be cleaned up.

subcommands/keys/ca_create.go
vkhoroz: We have a default value for the label... which I don't really like, but still.

That said, I actually like the idea to require this argument, so that we kind of remind the user once again that the HSM token should be configured properly.
This change would not be breaking, as the HSM support was anyway kind of broken in past versions.

Good catch! I will add a separate commit for that.

x509/storage_pkcs11_tool.go
vkhoroz: I have no strong opinion on this.

Exiting from here allows to keep the interfaces more concise; otherwise we'd need to drag the error up... just to die at a higher level.

x509/golang.go
vkhoroz: The `return` is because of the linter.

The dying is to not change the interface and all associated function calls.
We do `DieNotNil` in many low-level functions across Fioctl; I see no problem with that as long as it helps keep the code more concise.

Makefile
vkhoroz: We do run it in CI.

If someone wants to run it separately - they can do `go test` themselves.
I prefer to keep the Makefile concise enough.

x509/storage_pkcs11_tool.go
mike-sul: Yep, there are many ways to implement it. I am fine with the given one, just a bit odd at first glance.

x509/golang.go
mike-sul: >We do DieNotNil in many low-level functions across Fioctl; I see no problem with that as long as it helps keep the code more concise.

Fair enough, I suppose that's fine as long as there is no need to do some specific cleanup upper in the stack that is not automatically invoked when the "die" is called.

.github/workflows/test-pki.yml
```diff
@@ -0,0 +1,20 @@
+name: Test PKI
+on:
+  push:
```
camilamacedo86, line 3: I saw that you remove from push the linter
Do we want to test it out with push?

x509/golang.go
vkhoroz: hmm... interesting and actually valid point.

But, we do no cleanups ATM... so I will leave it for another day.

.github/workflows/test-pki.yml
vkhoroz: yes, I'm in a process of rebasing and cleaning up

x509/bash.go
```diff
@@ -9,38 +12,221 @@ import (
...
+`
+	run(script, ou)
 	return string(readFile(FactoryCaCertFile))
 }
 
 func CreateDeviceCa(cn string, ou string) string {
-	run("./"+CreateDeviceCaScript, DeviceCaKeyFile, DeviceCaCertFile)
-	return string(readFile(DeviceCaCertFile))
+	const script = `#!/bin/sh -e
+## This is an optional script a customer can use to create a certificate
```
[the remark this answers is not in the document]
vkhoroz: I copied these verbatim from the API, and only made enough changes for this to work.
The word "customer" is OK here... BTW we're going to scrap the Bash implementation few months from now.

subcommands/keys/ca_create.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/278/commits/bfe925df2ad3748c975d5d86a292a8a4da8e977e

vkhoroz:
Rebased on main to pull all linter/testing enhancements and cleaned the commit history.

Will re-test this whole thing tomorrow.

x509/bash.go
```diff
@@ -12,38 +12,190 @@ import (
...
+}
+
+func signCsr(script, tmpFileMask, csrPem string) string {
+	csrFile, err := os.CreateTemp("", tmpFileMask+".csr")
 	subcommands.DieNotNil(err)
-	defer os.Remove(tmpFile.Name())
-	defer tmpFile.Close()
-	_, err = tmpFile.Write([]byte(csrPem))
+	defer os.Remove(csrFile.Name())
+	defer csrFile.Close()
```
mike-sul, line 221: Maybe a single `defer func() {}` would be slightly better. The commands are printed in the order of execution and the returned error can be verified.

mike-sul:
LGTM.

I think it's important to test happy path for all PKI usage use-cases.

x509/bash.go
vkhoroz: I'd better skip this; not sure we are interested in these errors.

vkhoroz:
@doanac @mike-sul please, give it a final cut.

I tested the following happy-path scenaria:
- Take PKI offline without HSM:
   * Verify all keys and certs exist and are valid;
   * Verify that I can register the device via API (using online CA);
   * Verify that I can register the device with registration-ref (using offline CA);
   * Verify that both devices connect well to device-gateway (the second one is auto-registered), and ostree-proxy.
      + For device gateway, the TUF metadata loading via aktualizr-lite was verified, not the full update.
      + For ostree-proxy only the connection using openssl s_client was verified.

- Take PKI offline with HSM (with SoftHSM2):
   * Verify all keys and certs exist (except factory_ca.key) and are valid;
   * Verify that the factory CA key is stored in the HSM;
   * Verify that I can register the device via API (using online CA);
   * Verify that I can register the device with registration-ref (using offline CA);
   * Verify that both devices connect well to device-gateway and ostree-proxy.

Please, let me know if you have any other test scenario in mind.

doanac (approved):

PR #279, opened 2023-09-28
(chore): linter incorporates common rules that yield no violations

camilamacedo86:


.golangci.yml
```diff
@@ -1,3 +1,18 @@
...
+    - exportloopref
+    - gocyclo
+    - gofmt
+    - gosimple
+    - govet
+    - ineffassign
+    - staticcheck
+    - typecheck
+    - unparam
+    - unused
```
vkhoroz, line 18: It looks like your IDE or something is not adding the trailing line break to the last line.
This is a usual issue for everybody, and it makes GitHub mad about that.

vkhoroz (approved):
LGTM

Please, fix the line break and merge.

PR #280, opened 2023-09-28
(chore): Add misspelling linter check and fix issues

camilamacedo86:


subcommands/devices/list.go
```diff
@@ -68,25 +68,25 @@ func ownerFormatter(d *client.Device) string {
...
+	"apps":           {func(d *client.Device) string { return strings.Join(d.DockerApps, ",") }},
+	"up-to-date":     {func(d *client.Device) string { return fmt.Sprintf("%v", d.UpToDate) }},
+	"tag":            {func(d *client.Device) string { return d.Tag }},
+	"created-at":     {func(d *client.Device) string { return d.ChangeMeta.CreatedAt }},
+	"created-by":     {func(d *client.Device) string { return d.ChangeMeta.CreatedBy }},
+	"updated-at":     {func(d *client.Device) string { return d.ChangeMeta.UpdatedAt }},
+	"updated-by":     {func(d *client.Device) string { return d.ChangeMeta.UpdatedBy }},
+	"last-seen":      {func(d *client.Device) string { return d.LastSeen }},
+	"ostree-hash":    {func(d *client.Device) string { return d.OstreeHash }},
+	"current-update": {func(d *client.Device) string { return d.CurrentUpdate }},
```
doanac, line 87: while it is a typo - this does become a breaking change in functionality being slipped in. Something like this needs to go into the release notes so that when we break a user's shell script, we can show them the release notes mentioned it.

vkhoroz: @camilamacedo86 what this means is literally this:
- extract this bug fix into a separate commit/PR, and add smth like a `[release-note]` to the first commit line.

Later on when somebody will do an annotated tag (for a new release), this commit message will shout out loud at us.

camilamacedo86: I will do that 👍

camilamacedo86: done

doanac (approved):

vkhoroz (approved):
LGTM

PR #282, opened 2023-09-28
(chore): add unconvert linter check and fix issues

camilamacedo86:


vkhoroz (commented):
I like this one, but lets pause on this.
I mean, keep open but don't merge yet.

A current way how we run linter does not cover the Golang based PKI implementation (because of the build tag).
I've enabled verification of that thing and fixed some linter warnings in https://github.com/foundriesio/fioctl/pull/278.

Once that goes in, we can enable this and fix all new warnings at once (there shouldn't be many for this specific linter check).

camilamacedo86:
HI @vkhoroz,

Regards: https://github.com/foundriesio/fioctl/pull/282#pullrequestreview-1651336058

Just to clarify, the current check is focused on unnecessary conversions. So, unless there are such conversions in the changes you mentioned, this check shouldn't raise any issues. 

But if you want to wait for #278 that is all fine.

PR #284, opened 2023-09-28
chore: Improve Tool's Output for Better User Visibility and unify CLI output library

camilamacedo86:
## Description

Ensuring that users can swiftly identify critical warnings and errors is important for a smooth user experience. This PR focuses on making these critical messages stand out amidst the standard output of our tool. Specifically, the changes introduced are:
- Unify CLI Output Library: To provide a consistent output experience across all messages. (see that has places we are using fmt.Println others logrus which are not in debug mode and by last other, we are using the colour.faith )
- Color-coded Messages:
    - Errors are now highlighted in red, making them immediately noticeable.
    - Warnings are displayed in yellow, ensuring they are distinguishable from regular logs and errors.

Example:

Screenshot 2023-10-02 at 13 28 55

### DIFF
--- a/subcommands/common.go
+++ b/subcommands/common.go
@@ -8,6 +8,8 @@ import (
 	"path/filepath"
 	"text/tabwriter"
 
+	"github.com/fatih/color"
+
 	"github.com/cheynewallace/tabby"
 	canonical "github.com/docker/go/canonical/json"
 	homedir "github.com/mitchellh/go-homedir"
@@ -138,7 +140,7 @@ func DieNotNil(err error, message ...string) {
 			parts = append(parts, p)
 		}
 		parts = append(parts, err)
-		fmt.Println(parts...)
+		color.Red(fmt.Sprint(parts...))
 		for _, w := range onLastWill {
 			w()
 		}
@@ -177,7 +179,7 @@ func AssertWritable(path string) {
 	DieNotNil(err)
 	f, err := os.OpenFile(path, os.O_APPEND|os.O_WRONLY, st.Mode())
 	if err != nil {
-		fmt.Println("ERROR: File is not writeable:", path)
+		color.Red(fmt.Sprintf("ERROR: File is not writeable:%s", path))
 		os.Exit(1)
 	}
 	f.Close()
--- a/subcommands/targets/edit.go
+++ b/subcommands/targets/edit.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"os/exec"
 
+	"github.com/fatih/color"
+
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -41,16 +43,16 @@ func doEdit(cmd *cobra.Command, args []string) {
 	// Create temp file to edit with
 	tmpfile, err := os.CreateTemp("", "targets.*.json")
 	if err != nil {
-		fmt.Println("Unable to create tempfile: ", err)
+		color.Red("Unable to create tempfile: ", err)
 		os.Exit(1)
 	}
 	defer os.Remove(tmpfile.Name())
 	if _, err := tmpfile.Write(orig); err != nil {
-		fmt.Println("Unable to write tempfile: ", err)
+		color.Red("Unable to write tempfile: ", err)
 		os.Exit(1)
 	}
 	if err := tmpfile.Close(); err != nil {
-		fmt.Println("Unable to close tempfile: ", err)
+		color.Red("Unable to close tempfile: ", err)
 		os.Exit(1)
 	}
 
@@ -72,7 +74,7 @@ func doEdit(cmd *cobra.Command, args []string) {
 	// Read it and see if its changed
 	content, err := os.ReadFile(tmpfile.Name())
 	if err != nil {
-		fmt.Println("ERROR: Unable to re-read tempfile:", err)
+		color.Red("ERROR: Unable to re-read tempfile:", err)
 	}
 	if bytes.Equal(content, orig) {
 		fmt.Println("No changes found, exiting.")
@@ -83,7 +85,7 @@ func doEdit(cmd *cobra.Command, args []string) {
 	var newTargets tuf.Files
 	err = json.Unmarshal(content, &newTargets)
 	if err != nil {
-		fmt.Println("Unable to parse new targets: ", err)
+		color.Red("Unable to parse new targets: ", err)
 		os.Exit(1)
 	}
 	type TargetsUp struct {
@@ -92,7 +94,7 @@ func doEdit(cmd *cobra.Command, args []string) {
 	upload := TargetsUp{newTargets}
 	content, err = json.Marshal(upload)
 	if err != nil {
-		fmt.Println("Unable to marshall targets data: ", err)
+		color.Red("Unable to marshall targets data: ", err)
 		os.Exit(1)
 	}
 
--- a/subcommands/targets/offline-update.go
+++ b/subcommands/targets/offline-update.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/fatih/color"
+
 	"github.com/foundriesio/fioctl/client"
 	"github.com/foundriesio/fioctl/subcommands"
 	"github.com/spf13/cobra"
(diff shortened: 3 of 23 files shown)


doanac:
I think using a logger for a CLI user interface is a little unusual. I believe the more common approach is abstracting code that prints stuff into some type of project "formatter" where the `Print` type function gets some context to help know about coloring.

camilamacedo86:
Hey @doanac! 😊

> I think using a logger for a CLI user interface is a little unusual. I believe the more common approach is abstracting code that prints stuff into some type of project "formatter" where the Print type function gets some context to help know about coloring.

Ah, you got a point! We've got logrus in the mix already, so I figured, why not stick to the beat? 

But hey, I’m totally vibing with the fmt.Print idea. It’s neat and cuts down on the extra baggage – cool beans! I’ll spin that record and make the tweaks when I catch a free moment.

Thanks a bunch! 🚀

subcommands/common_config.go
```diff
@@ -258,7 +258,7 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
 	}
 
 	if !changed {
-		fmt.Println("No changes found. Device is already configured with the specified options.")
+		logrus.Infof("No changes found. Device is already configured with the specified options.")
```
vkhoroz, line 261: These are informational messages, not info-level diagnostic logs.
Please, revert this specific commit.

subcommands/targets/edit.go
```diff
@@ -65,7 +65,7 @@ func doEdit(cmd *cobra.Command, args []string) {
 	edit.Stdin = os.Stdin
 	logrus.Debug("Running editor and waiting for it to finish...")
 	if err := edit.Run(); err != nil {
-		fmt.Println("Editing cancelled: ", err)
+		logrus.Warning("Editing cancelled: ", err)
```
vkhoroz, line 68: This one might be tempting, but based on the context it is also an informational message.

subcommands/targets/offline-update.go
```diff
@@ -89,7 +90,7 @@ func doOfflineUpdate(cmd *cobra.Command, args []string) {
 			fmt.Printf("Downloading Apps fetched by the `assemble-system-image` run; build number:  %d, tag: %s...\n", ti.version, ti.buildTag)
 			err = downloadApps(factory, targetName, ti.version, ti.buildTag, path.Join(dstDir, "apps"))
 			if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
-				fmt.Println("WARNING: The Target Apps were not fetched by the `assemble` run, make sure that App preloading is enabled if needed. The update won't include any Apps!")
+				logrus.Warning("WARNING: The Target Apps were not fetched by the `assemble` run, make sure that App preloading is enabled if needed. The update won't include any Apps!")
```
vkhoroz, line 93: This is also just text. Please, revert.

cmd/docs.go
```diff
@@ -34,7 +35,7 @@ func doGenRstDocs(cmd *cobra.Command, args []string) {
 	rootCmd.DisableAutoGenTag = true
 	err := doc.GenReSTTreeCustom(rootCmd, outDir, filePrepender, linkHandler)
 	if err != nil {
-		fmt.Println("ERROR:", err)
+		logrus.Error("ERROR:", err)
 		os.Exit(1)
 	}
```
vkhoroz, line 40: Where I see the value in this commit is that you seem to have spotted all places where we still use a legacy way to exit on error.
All these editions would need to be refactored as below:
```suggestion
subcommands.DieNotNil(doc.GenReSTTreeCustom(rootCmd, outDir, filePrepender, linkHandler))
```

Now, here is the thing: it takes time, so should ultra-low priority.

subcommands/common.go
```diff
@@ -138,7 +138,7 @@ func DieNotNil(err error, message ...string) {
 			parts = append(parts, p)
 		}
 		parts = append(parts, err)
-		fmt.Println(parts...)
+		logrus.Error(parts...)
```
vkhoroz, line 141: Now on this...

I have already posted in the other PR that we use logs in debug debug mode, but not for regular user interaction.
Although colorful dialogs look cool and fancy, the control characters used to print them break some terminals.

I suspect that logrus is smart enough to only insert colors for terminals which support them.
Because, e.g. I've never seen these colors in my terminal.

Still, it is weird to see smth like `[ERRO] ERROR: your message goes here` in a regular (non-debug) command run.
I've tried that once as an excercise, and IIRC a team consunsus was that it looks strange.

So, `fmt.Println` for the savior. A `logrus` is nice for developers debugging cloud applications, but it does not fit all sizes.

cmd/docs.go
camilamacedo86: It seems out of scope of this one as well
However, would be great if we start to track those needs down.
WDYT? Could we use issues for that?

doanac: Don't sweat the "docs" cmd stuff. that exists for us to create documentation in docs.fio.

vkhoroz: I mean, instead of using `logrus` which its fancy formatting, I'd better use a function we already introduced a while ago to handle errors.

But seriously, I would do neither of the above.
We have more important stuff to do.
That's why our new and old code styles are organically interwoven in many places, and we are all good with that.

camilamacedo86:
Hi @vkhoroz, @doanac,

Firstly, thank you for your valuable insights regarding this small suggestion.

**Regarding the use of `println` with ANSI colors**:

Using `fmt.Println("\033[31mERROR:", "Your message", "\033[0m")` might not deliver as expected. In terminals that don't support ANSI sequences (which seem prevalent in older Unix and Windows versions), the output won't display colors and will instead appear as `\033[31mERROR message\033[0m`.

On the other hand, logrus appears to be adept at navigating these scenarios, opting to simply output `ERRO` in incompatible terminals.

**The driving force behind this proposal was to enact a low-effort/good first issue that could provide tangible value to our users.** With the volume of notifications we produce, errors and warnings could easily go unnoticed. Given our ongoing use of logrus, it felt like a logical step to further embed it, ensuring greater code consistency.

**However, if logrus doesn't align with the team's vision, especially when weighing the challenges associated with `fmt.Println` and ANSI, it seems wise to reassess this one.**

For now, I'll be closing this. However, if you have alternative strategies in mind or believe there's a more effective approach worth exploring, please don't hesitate to reach out. I'm always here to assist and collaborate

subcommands/common.go
camilamacedo86: Hi @vkhoroz,

Indeed, I understand that you're right in noting:

```plaintext
I suspect that logrus is smart enough to only insert colors for terminals which support them.
```

Based on my research, Logrus has intelligence built-in to ensure it doesn't break terminals that can't handle ANSI sequences.

However, I noticed you mentioned:

```plaintext
Still, it is weird to see something like [ERRO] ERROR: your message goes here.
```

That is not caused by the usage of the lib itself. It is because we are adding `ERROR` and `WARNING` manually. This can be solved by:
```go
// Instead of this
logrus.Error("ERROR: ", err)
// Use this
logrus.Error(err)
```

vkhoroz: We do use https://github.com/fatih/color in some places to add some color; which also identifies terminal settings to some extent (this auto-detection is never ideal).
So, if it is about highliting `ERROR:` in some color, we could do it, if the team agrees.

The point was not to use `logrus` for something it was not created for.

camilamacedo86:
I re-opened this one because I notice that are also using `"github.com/fatih/color"` in some places
So, could we consider this one acceptable for the errors in red and warnings in yellow since we already use it? 

See: 

https://github.com/foundriesio/fioctl/blob/0ac0f5f578cf9c133b944dcf304b8196a6a809a1/subcommands/keys/tuf_updates_review.go#L93-L97

subcommands/common.go
camilamacedo86: That is great I reopen this one when I saw the bump for fatih 
Now, I am checking this message so I think I have the answer for: https://github.com/foundriesio/fioctl/pull/284#issuecomment-1742602597

Cool. It is not a high priority and I am doing along side of the other things 
So, I will address this one in this direction asap I have time to work on it.

Thank you.

camilamacedo86: I will close this one let's see

vkhoroz:
@camilamacedo86 let's table this, please.

I know that this change was done by some sort of `find magic -exec sed magic`, so it was probably easy.
But, TBH I don't like the final result.

If I were a user, I'd like only the `ERROR` part to be in red, and the message itself not so fancy.
This also means, that as a developer I'd want all these messages to be refactored to use `subcommands.DieNotNil` where applicable.

But, I feel like I'm starting a discussion by this sentence; and I don't want a new round of debates.
We have more important tasks to spend our effort.

PR #285, opened 2023-09-28
(feat): Warning users about create OSTree static deltas before rolling out waves

camilamacedo86:
Adds a warning message to inform users about the recommendation to use Static Deltas to optimize OTA downloads.
The message only appears when users create weaves that do not have static deltas.

## Local tests:

When no Static Deltas are in place: 

Screenshot 2023-10-03 at 01 37 14

### DIFF
--- a/subcommands/waves/init.go
+++ b/subcommands/waves/init.go
@@ -27,7 +27,12 @@ func init() {
 This command only initializes a wave, but does not provision its updates to devices.
 Use a "fioctl wave rollout <wave> <group>" to trigger updates of this wave to a device group.
 Use a "fioctl wave complete <wave>" to update all devices (make it globally available).
-Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).`,
+Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).
+
+We recommend that you generate static deltas for your production targets to optimize 
+OTA update download. Consider generating a static delta for targets using:
+$ fioctl targets static-deltas
+`,
 		Run:  doInitWave,
 		Args: cobra.ExactArgs(3),
 		Example: `
@@ -145,6 +150,31 @@ func doInitWave(cmd *cobra.Command, args []string) {
 	} else {
 		subcommands.DieNotNil(api.FactoryCreateWave(factory, &wave), "Failed to create a wave")
 	}
+
+	if !hasStaticDelta(new_targets) {
+
+		fmt.Print(`
+WARNING: You created a wave for a target version without static deltas.
+
+We recommend that you generate static deltas for your production targets to optimize
+OTA update download. Consider generating a static delta for targets using: 
+$ fioctl targets static-deltas
+
+You can then cancel the wave and create a new one by using a static delta.
+`)
+
+	}
+}
+
+func hasStaticDelta(new_targets tuf.Files) bool {
+	for _, file := range new_targets {
+		custom, err := api.TargetCustom(file)
+		subcommands.DieNotNil(err)
+		if custom.DeltaStats == nil {
+			return false
+		}
+	}
+	return true
 }
 
 func pruneTargets(currentTargets *client.AtsTargetsMeta, versions []string) client.AtsTargetsMeta {
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -277,6 +277,10 @@ type ProjectTrigger struct {
 	Id      int             `json:"id,omitempty"`
 	Secrets []ProjectSecret `json:"secrets"`
 }
+type DeltaStats struct {
+	Sha256 string `json:"sha256"`
+	Size   int    `json:"size"`
+}
 
 type TufCustom struct {
 	HardwareIds    []string              `json:"hardwareIds,omitempty"`
@@ -286,6 +290,7 @@ type TufCustom struct {
 	ComposeApps    map[string]ComposeApp `json:"docker_compose_apps,omitempty"`
 	Name           string                `json:"name,omitempty"`
 	ContainersSha  string                `json:"containers-sha,omitempty"`
+	DeltaStats     *DeltaStats           `json:"delta-stats,omitempty"`
 	LmpManifestSha string                `json:"lmp-manifest-sha,omitempty"`
 	OverridesSha   string                `json:"meta-subscriber-overrides-sha,omitempty"`
 	Uri            string                `json:"uri,omitempty"`


subcommands/waves/init.go
```diff
@@ -28,9 +29,15 @@ This command only initializes a wave, but does not provision its updates to devi
...
 Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).`,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			confirmWaveInitProceed()
+		},
 		Run:  doInitWave,
 		Args: cobra.ExactArgs(3),
 		Example: `
+**(Recommended)** Generate OSTree static deltas before rolling out waves to devices.
+More info: https://docs.foundries.io/latest/reference-manual/ota/static-deltas.html#ref-static-deltas
+
```
vkhoroz, line 40: I don't think this is an appropriate place to put this.
Maybe, it might fit at the end of the `Long` help...

But we decided to print it at the end of command execution; not sure we need to duplicate it in two places.
That might scare the user from using Waves instead of encouraging them to use Static Deltas.

@doanac @mike-sul your thoughts.

subcommands/waves/init.go
```diff
@@ -54,6 +61,20 @@ Example: 1,2,3`)
 	_ = initCmd.MarkFlagRequired("keys")
 }
 
+func confirmWaveInitProceed() {
+	logrus.Warning("**(Recommended)** Generate OSTree static deltas before rolling out waves to devices. " +
+		"More info: https://docs.foundries.io/latest/reference-manual/ota/static-deltas.html#ref-static-deltas \n")
```
vkhoroz, line 66: I am on the fence here.
Almost all our commands are non-interactive, requiring user input only in the form of arguments.
This is nice, because it works on all terminals and is easy to script.

It is totally fine to do waves without static deltas, or static deltas without waves.
We have customers doing all sort of magic using various machinery combinations.
Still, there is a definite value in showing something to the user if there are no static deltas.
So, I would prefer to do just that: suggest the user to generate static deltas (at the end of `doInitWave`), but don't interrupt the operation.
Our dialogs should be supportive, not annoying.

Also, I have suggestions to improve the message itself, subject to @kprosise remarks.

```suggestion
fmt.Println(`You created a wave for a target version without static deltas.
We recommend that you generate static deltas for your production targets to optimize the size of the update.
Please, consider generating a static delta for this target version using `fioctl targets static-deltas` command.
You can then cancel this wave and create a new one for a target with a static delta.`)
```

A rationale is:
- Put no links to documentation: they may be moved arbitrarily, while the user might use old Fioctl for a long time.
  Furthermore, the tool looks more professional when it can describe itself without a necessity to direct a user to external resource.
- This is a command-line tool, so please, use plain text: no markup, no colors, no log-styled clobbery.
  We use logs sometimes, but as intended - to show some diagnistics in debug mode; not for regular user interaction.

subcommands/waves/init.go
```diff
@@ -28,9 +29,15 @@ This command only initializes a wave, but does not provision its updates to devi
 Use a "fioctl wave rollout <wave> <group>" to trigger updates of this wave to a device group.
 Use a "fioctl wave complete <wave>" to update all devices (make it globally available).
 Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).`,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			confirmWaveInitProceed()
+		},
```
vkhoroz, line 34: Please, simply add the dialog into `doWaveInit`, don't make a control flow more complex to grasp.

subcommands/waves/init.go
```diff
@@ -54,6 +61,20 @@ Example: 1,2,3`)
 	_ = initCmd.MarkFlagRequired("keys")
 }
 
+func confirmWaveInitProceed() {
+	logrus.Warning("**(Recommended)** Generate OSTree static deltas before rolling out waves to devices. " +
+		"More info: https://docs.foundries.io/latest/reference-manual/ota/static-deltas.html#ref-static-deltas \n")
+
+	logrus.Infof("Do you want to proceed with the wave init? (y/n): ")
```
vkhoroz, line 68: I think we can be smarter than this.

Based on [this code](https://github.com/foundriesio/ci-scripts/blob/master/lmp/static_deltas.py#L90) I see that we can easily identify if a given target has static deltas or not.
So, it only makes sense to warn the user in case if there are no static deltas, and silently proceed if everything is fine.

mike-sul: I think this is what I and @doanac agreed on in the slack channel thread.

I think it's better to make it a non-interactive too.

I don't have a strong opinion whether to print the warning message in the beginning or at the end of the command output.

subcommands/waves/init.go
kprosise: @vkhoroz, looks good

subcommands/waves/init.go
```diff
@@ -145,6 +150,31 @@ func doInitWave(cmd *cobra.Command, args []string) {
...
+`)
+
+	}
+}
+
+func hasStaticDelta(new_targets tuf.Files) bool {
+	for _, file := range new_targets {
+		custom, err := api.TargetCustom(file)
+		subcommands.DieNotNil(err)
+		if custom.DeltaStats != nil {
```
mike-sul, line 173: I think we should return `true` only if delta is present for each target. So, maybe 
```
if custom.DeltaStats == nil {
  return false;
}
```

subcommands/waves/init.go
```diff
@@ -145,6 +150,31 @@ func doInitWave(cmd *cobra.Command, args []string) {
...
+You can then cancel this wave and create a new one for a target with a static delta.
+`)
+
+	}
+}
+
+func hasStaticDelta(new_targets tuf.Files) bool {
+	for _, file := range new_targets {
+		custom, err := api.TargetCustom(file)
+		subcommands.DieNotNil(err)
```
mike-sul, line 172: I am not sure we need to "die" in the case of error since printing the warning message about static delta is not mandatory. Maybe, just warning would suffice.

subcommands/waves/init.go
```diff
@@ -27,7 +27,12 @@ func init() {
 This command only initializes a wave, but does not provision its updates to devices.
 Use a "fioctl wave rollout <wave> <group>" to trigger updates of this wave to a device group.
 Use a "fioctl wave complete <wave>" to update all devices (make it globally available).
-Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).`,
+Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).
+
+We recommend that you generate static deltas for your production targets to optimize
```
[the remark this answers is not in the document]
vkhoroz: Thanks @mike-sul, I was waiting for your input to clarify this text.

subcommands/waves/init.go
vkhoroz: I am on the fence here.

From one perspective, we can say the user that we were not able to check if the wave has static deltas.

But, this clearly means that the JSON structure is invalid.
To make matters worse, we've already verified it to be valid earlier.
Thus, it is definitely an error.

So, I _guess_ it is better to let the user know that we have a bug in our code by shouting out loud.

subcommands/waves/init.go
```diff
@@ -145,6 +150,31 @@ func doInitWave(cmd *cobra.Command, args []string) {
...
+	if !hasStaticDelta(new_targets) {
+
+		fmt.Print(`
+WARNING: You created a wave for a target version without static deltas.
+
+We recommend that you generate static deltas for your production targets to optimize
+the size of the update. Please, consider generating a static delta for targets using:
+$ fioctl targets static-deltas.
+
+You can then cancel this wave and create a new one for a target with a static delta.
```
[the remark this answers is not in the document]
mike-sul: Thanks @vkhoroz for clarification. For some reason I've always assumed that the prod Targets are updated if their corresponding CI targets are updated, I must have mistaken it with the CI "tagged" targets update.

subcommands/waves/init.go
mike-sul: ok, and it makes the implementation simpler.

subcommands/waves/init.go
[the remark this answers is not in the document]
camilamacedo86: Done

subcommands/waves/init.go
camilamacedo86: Ops. Thx

subcommands/waves/init.go
camilamacedo86: Cool closing this one as done.
thx

subcommands/waves/init.go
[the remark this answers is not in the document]
camilamacedo86: Cool closing this one as done.
thx

subcommands/waves/init.go
```diff
@@ -145,6 +150,31 @@ func doInitWave(cmd *cobra.Command, args []string) {
 	} else {
 		subcommands.DieNotNil(api.FactoryCreateWave(factory, &wave), "Failed to create a wave")
 	}
+
+	if !hasStaticDelta(new_targets) {
+
+		fmt.Print(`
+WARNING: You created a wave for a target version without static deltas.
+
+We recommend that you generate static deltas for your production targets to optimize
```
mike-sul, line 159: I've tested this change, it works as expected.

I am not sure that the wording of this message as well as the message at #32 conveys accurately what needs to be done, though. I'll let the support team representative @vanmaegima to judge on it.

In any case, I think we can improve it at the next iteration/later, so I think it's good to go.

mike-sul (approved):

subcommands/waves/init.go
```diff
@@ -27,7 +27,12 @@ func init() {
 This command only initializes a wave, but does not provision its updates to devices.
 Use a "fioctl wave rollout <wave> <group>" to trigger updates of this wave to a device group.
 Use a "fioctl wave complete <wave>" to update all devices (make it globally available).
-Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).`,
+Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).
+
+We recommend that you generate static deltas for your production targets to optimize 
+OTA update download. Please, consider generating a static delta for targets using:
```
kprosise, line 33:
```suggestion
OTA update download. Consider generating a static delta for targets using: 
```

Optional suggestion

subcommands/waves/init.go
```diff
@@ -145,6 +150,31 @@ func doInitWave(cmd *cobra.Command, args []string) {
...
 		subcommands.DieNotNil(api.FactoryCreateWave(factory, &wave), "Failed to create a wave")
 	}
+
+	if !hasStaticDelta(new_targets) {
+
+		fmt.Print(`
+WARNING: You created a wave for a target version without static deltas.
+
+We recommend that you generate static deltas for your production targets to optimize
+OTA update download. Please, consider generating a static delta for targets using:
```
kprosise, line 160:
```suggestion
OTA update download. Consider generating a static delta for targets using:
```

As above, optional

subcommands/waves/init.go
```diff
@@ -145,6 +150,31 @@ func doInitWave(cmd *cobra.Command, args []string) {
...
+$ fioctl targets static-deltas
+
+You can then cancel this wave and create a new one for a target with a static delta.
```
kprosise, line 163:
```suggestion
You can then cancel the wave and create a new one by using a static delta.
```

Optional suggestion! Just shortening the sentence a bit.

kprosise (approved):
Left some minor, and optional suggestions.

subcommands/waves/init.go
vkhoroz: @kprosise @camilamacedo86 This suggestion makes a message somewhat misleading.

A user cannot create a wave _by using a static delta_, but a user can create a wave _for a target which has static deltas defined_.

So, I suggest to either revert this message to its original, or make a new suggestion which provides a more accurate context to the user.

subcommands/waves/init.go
```diff
@@ -27,7 +27,12 @@ func init() {
 This command only initializes a wave, but does not provision its updates to devices.
 Use a "fioctl wave rollout <wave> <group>" to trigger updates of this wave to a device group.
 Use a "fioctl wave complete <wave>" to update all devices (make it globally available).
-Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).`,
+Use a "fioctl wave cancel <wave> to cancel a wave (make it no longer available).
+
+We recommend that you generate static deltas for your production targets to optimize 
+OTA update download. Consider generating a static delta for targets using:
```
vkhoroz, line 33: We use the [semantic newlines](https://bobheadxi.dev/semantic-line-breaks/) in help messages.
Please, see the related section in the [style guide](https://foundriesio.atlassian.net/wiki/spaces/ID/pages/2392067/Foundries.io+Style+and+Communication+Guide#General-Guidelines).

```suggestion
We recommend that you generate static deltas for your production targets to optimize OTA update download.
Consider generating a static delta for targets using:
```

The line length is absolutely irrelevant here, as long as the sentence fits within [25 words](https://foundriesio.atlassian.net/wiki/spaces/ID/pages/2392067/Foundries.io+Style+and+Communication+Guide#Sentence-Length).

subcommands/waves/init.go
```diff
@@ -145,6 +150,31 @@ func doInitWave(cmd *cobra.Command, args []string) {
...
+	if !hasStaticDelta(new_targets) {
+
+		fmt.Print(`
+WARNING: You created a wave for a target version without static deltas.
+
+We recommend that you generate static deltas for your production targets to optimize
+OTA update download. Consider generating a static delta for targets using: 
+$ fioctl targets static-deltas
+
+You can then cancel the wave and create a new one by using a static delta.
```
vkhoroz, line 163: Same about semantic newlines:

```suggestion
WARNING: You created a wave for a target version without static deltas.
We recommend that you generate static deltas for your production targets to optimize OTA update download.
Consider generating a static delta for targets using:
$ fioctl targets static-deltas

You can then cancel the wave and create a new one for a target which has static deltas defined.
```

This also incorporates my above suggestion about _using static deltas_.

vkhoroz (commented):
@camilamacedo86 please, make a follow-up PR to address the 2 comments from me.

subcommands/waves/init.go
camilamacedo86: I agree. Good point.

PR #298, opened 2023-10-02
Improve: use the same linter setup in CI and locally

vkhoroz:
Locally we run format and lint checks, but only lint check in CI. In the future we also need to run linter for several build tags combinations. It is cumbersome to support two different linter setups: for local run and for CI.

So, this change makes sure that:
- GitHub actions call `make check` to run exactly the same linter as we do locally.
- JobServ action which actually did the right job can now be removed as a duplicate.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,5 @@
 name: Lint
 on:
-  push:
   pull_request:
 
 jobs:


vkhoroz:
@camilamacedo86 please, see this why this is needed:
https://github.com/foundriesio/fioctl/pull/278/commits/2cd381518d42b565388bd2820cdd28ed4c6c2803

The other option would be to have several linter actions, which would make supporting this cumbersome, as we'd have separate copies of linter runs for local vs CI environments, with a risk that they would deviate over time.

Now we can have a single source of truth.

vkhoroz:
If we have this merge, we could also merge https://github.com/foundriesio/fioctl/pull/282 faster, as I'd be able to resolve all new unconvert warnings in scope of https://github.com/foundriesio/fioctl/pull/278 (if any).

camilamacedo86 (approved):
That looks great to me! 👍

Just a small nitpick:
I believe the format check might be redundant since it's already being checked by golangci. However, let's keep it consistent for now, and we can consider cleaning it up in the future. good catcher

PR #300, opened 2023-10-06
targets: offline-update: Check if dest dir already has data

detsch:
This commit prevents the execution of "targets offlile-update" using a destination directory that already contains update data (i.e., "ostree_repo" or "apps" subdirectories). There is no check for the tuf subdirectory, since the metadata in that subdir is updated by default, and there is no conflict.
The content of the destination directory can be cleaned up automatically if the new --force option is used.

### DIFF
--- a/subcommands/targets/offline-update.go
+++ b/subcommands/targets/offline-update.go
@@ -29,11 +29,12 @@ type (
 )
 
 var (
-	ouTag       string
-	ouProd      bool
-	ouExpiresIn int
-	ouTufOnly   bool
-	ouNoApps    bool
+	ouTag                  string
+	ouProd                 bool
+	ouExpiresIn            int
+	ouTufOnly              bool
+	ouNoApps               bool
+	ouAllowMultipleTargets bool
 )
 
 func init() {
@@ -62,6 +63,8 @@ func init() {
 		"Fetch only TUF metadata")
 	offlineUpdateCmd.Flags().BoolVarP(&ouNoApps, "no-apps", "", false,
 		"Skip fetching Target Apps")
+	offlineUpdateCmd.Flags().BoolVarP(&ouAllowMultipleTargets, "allow-multiple-targets", "", false,
+		"Allow multiple targets to be stored in the same <dst> directory")
 }
 
 func doOfflineUpdate(cmd *cobra.Command, args []string) {
@@ -84,6 +87,14 @@ func doOfflineUpdate(cmd *cobra.Command, args []string) {
 	fmt.Println("Successfully refreshed and downloaded TUF metadata")
 
 	if !ouTufOnly {
+		if !isDstDirClean(dstDir) {
+			if !ouAllowMultipleTargets {
+				subcommands.DieNotNil(errors.New(`Destination directory already has update data.
+Provide a clean destination directory or re-run with --allow-multiple-targets to add a new target to a directory which already has update data.
+Notice that multiple targets in the same directory is only supported in LmP >= v92.`))
+			}
+		}
+
 		fmt.Printf("Downloading an ostree repo from the Target's OE build %d...\n", ti.ostreeVersion)
 		subcommands.DieNotNil(downloadOstree(factory, ti.ostreeVersion, ti.hardwareID, dstDir), "Failed to download Target's ostree repo:")
 		if !ouNoApps {
@@ -150,6 +161,17 @@ func getTargetInfo(factory string, targetName string) (*ouTargetInfo, error) {
 		}
 	}
 	return &info, nil
+}
+
+func isDstDirClean(dstDir string) bool {
+	for _, subDir := range []string{"ostree_repo", "apps"} {
+		fullPath := path.Join(dstDir, subDir)
+		if _, err := os.Stat(fullPath); !os.IsNotExist(err) {
+			fmt.Println(fullPath + " already exists")
+			return false
+		}
+	}
+	return true
 }
 
 func downloadTufRepo(factory string, target string, tag string, prod bool, expiresIn int, dstDir string) error {


subcommands/targets/offline-update.go
```diff
@@ -62,6 +63,8 @@ func init() {
 		"Fetch only TUF metadata")
 	offlineUpdateCmd.Flags().BoolVarP(&ouNoApps, "no-apps", "", false,
 		"Skip fetching Target Apps")
+	offlineUpdateCmd.Flags().BoolVarP(&ouForce, "force", "", false,
+		"Cleanup previous downloaded update data from <dst>, if not empty")
```
vkhoroz, line 67: I'm on the fence here.

This looks about right; but we use `--force` argument when we want to overcome the server-side validation.
Here, what you do is overwrite the local files.

Just IMHO, the user should do the directory cleanup on their own.
But, if you do want a flag, I'd call it smth like `--overwrite`.

subcommands/targets/offline-update.go
```diff
@@ -152,6 +166,31 @@ func getTargetInfo(factory string, targetName string) (*ouTargetInfo, error) {
...
 
+func isDstDirClean(dstDir string) bool {
+	for _, subDir := range []string{"ostree_repo", "apps"} {
+		fullPath := path.Join(dstDir, subDir)
+		if _, err := os.Stat(fullPath); !os.IsNotExist(err) {
+			fmt.Println(fullPath + " already exists")
+			return false
+		}
+	}
+	return true
```
vkhoroz, line 177: We have the right to require a user providing a clean folder.
So, I'd simply check here if 2 things:
- does the folder exist.
- is it empty.

Let's keep it simple; the user should be the housekeeping.

vkhoroz (commented):
@mike-sul @doanac do you want to add something?

subcommands/targets/offline-update.go
mike-sul: This is the request from customer support.

I am fine with either `--force` or `--overwrite`. Another option is `--cleanup`.

subcommands/targets/offline-update.go
mike-sul: Why not just `does the folder exist.` as currently implemented.

subcommands/targets/offline-update.go
doanac: Thing about a usability feature like this: The user is not going to know about the flag, and hits a "this exists error" when they try the command. Either way, they have to run 2 commands.

Seems like the real usability fix would be to auto-delete the directory for the user. But I think we all agree that's dangerous (say they made did `fioctl targets ... $HOME`). I'm not sure I want our code to delete anything - I'd rather make them run `rm -rf $HOME` than `fioct targets .... --force $HOME`

mike-sul: User might not know what exactly to remove inside the dst folder.

Having said that, this approach contradicts with another request - support multiple targets in one update source directory.

mike-sul: Maybe, the command should fail if these two folders exists by default.

Then add another option `--allow-multiple-targets` which would allow customers to pull another Target data to the dst folder.

subcommands/targets/offline-update.go
vkhoroz: Currently we check that `${dir}/apps` and `${dir}/ostree_repo` should not exist.

I propose to only check that `${dir}` either does not exist or is empty; which is a more generic and more strong check.

subcommands/targets/offline-update.go
vkhoroz: So, I see us mixing many things here:
- 1. A safety check that a user does not delete important data.
- 2. An usability feature to cleanup that date _on behalf_ of the user.
- 3. Another usability feature to allow adding more content to existing folder.

Just IMHO, it is better to implement 1 without 2 in any case.
Furthermore, 2 and 3 kind of contradict each other.

That's why I like the latest suggestion from @mike-sul about `--allow-multiple-targets` to implement 3.

subcommands/targets/offline-update.go
mike-sul: The dir includes `tuf` dir which is OK to be present when the command executed. This is happen when customers want to refresh tuf meta.

subcommands/targets/offline-update.go
mike-sul: OK, I think we should not remove any folder too. Let's just check and fail if one or two of those folders (ostree_repo and apps) exists, unless `--allow-multiple-targets` is explicitly specified.

subcommands/targets/offline-update.go
vkhoroz: I prefer to follow the KISS principle here.
I mean, I don't want us to pay attention to _what's OK to have in that folder and what's not OK_.
Even less so I want us to try explaining it to the user.

A requirement to provide an empty folder is justifiable and understandable.
Anything beyond that is convoluted, so I'd only do that if we have a customer request with enough motivation to complicate matters.

mike-sul: There is no need to explain anything to a user.
We just discussing here "check $dir" vs "check $dir/app and $dir/ostree_repo".
Is checking two folders compared to checking one folder makes the check so significantly complex? IMO, the code is almost the same.

vkhoroz: I simply don't get: what is the business use case to check for two folders?

There must be a justifiable answer to that question, _before_ we start talking about the pros and cons.
If there is no customer-facing use case for that, I see no rationale to make our code and support more complex.

mike-sul: >I simply don't get: what is the business use case to check for two folders?

There is no need to print any error message and ask a user to remove $dir  if only $dir/tuf  is present. A user does NOT need to know about this detail.

We just simply check if either $dir/apps or $dir/ostree is present. If so the command prints an error and exits. The error message suggest a user to remove the whole `$dst` folder.

mike-sul: Also, we should keep in mind that the $dst can be NOT empty, a user may want to to put there some custom data to be transferred to a device. So, in the end, this is up to a user to decide if to delete the whole $dst directory or just $dst/apps and $dst/ostree.

vkhoroz: > The error message suggest a user to remove the whole $dst folder.

This is where the major part of the confusion comes from.
We should never advice the user to delete `$dst` folder; we should advice to provide an empty folder.
Remember, the `$dst` might be the user's home or the root.

Now, once that confusion is settled, I don't want a user to even know all the details about _which sub-folder is safe to overwrite and which is not_, unless there is a business use case for that. I see no such use case declared in this thread so far.

vkhoroz: The difference is huge conceptually:
- When I see `foo/bar is not empty, delete this folder`, I naїvely run `rm -rf foo/bar`.
- When I see `foo/bar is not empty, provide an empty folder`, I intuitively run `mkdir -p foo/baz`.

mike-sul: >we should advice to provide an empty folder.

This kind of breaks the current API, we've never asked about that, and user expects that they can have anything in that folder.

>Remember, the $dst might be the user's home or the root.

They they may do that. But we need to allow them to have anything they want in $dst folder.

Also, similar argument can be imposed to almost any other `fioctl` command.

mike-sul: >When I see foo/bar is not empty, delete this folder, I naїvely run rm -rf foo/bar.

That is why the most safest way to do is to check those two specific folders, and advise to remove the specific folder. If we check just $dst and advise them to delete is then it's not quite save.

vkhoroz: Hmm... we require a folder to be empty, or the file to not exist _in almost any other fioctl command_ already.

Just IMHO this problem has happened _because_ we failed to require that folder to be empty initially.
So, this seems to be more a bug fix than a breaking change.

Now, if the user wants to put smth into it - they can certainly do that afterwards.
We need not be tied by such restrictive assumptions.

Our job is to perform routine user operations in a safe way.
Requiring that a folder is empty is _safer_ than what you propose @mike-sul.

vkhoroz: @mike-sul I've got your point.

I'd like to also hear from others on thie subject. @doanac @detsch

detsch:
Updated changes replacing "--force" logic with "--allow-multiple-targets"

subcommands/targets/offline-update.go
detsch: IMO requiring an empty folder would be sliglthy better, but since there are reasons to not do so, I would keep checking for the 2 specific paths.

About the messaging, I would not suggest any removal, but still present to the user with a hint on what path was detected and eventually led to the command to fail. This is the output from the updated coded:
```
$ fioctl targets offline-update intel-corei7-64-lmp-27 ./offline-update-content --tag main --expires-in-days 15 -f detsch-x86-64 
Checking whether Target exists; target: intel-corei7-64-lmp-27, tag: main, production: false
Refreshing and downloading TUF metadata for Target intel-corei7-64-lmp-27 to offline-update-content/tuf...
Successfully refreshed and downloaded TUF metadata
offline-update-content/ostree_repo already exists
Destination directory already has update data
ERROR: Please provide a clean destination directory or re-run with --allow-multiple-targets to add a new target to a directory which already has update data.
Notice that multiple targets in the same directory is only supported in LmP >= v92.
```

A user struggling with the command from some reason will notice the 
`offline-update-content/ostree_repo already exists` message.

vkhoroz: @detsch this looks way safer... taking my discussion with @mike-sul to negligance.
Might be a silver lining.

mike-sul (approved):

subcommands/targets/offline-update.go
```diff
@@ -84,6 +87,15 @@ func doOfflineUpdate(cmd *cobra.Command, args []string) {
 	fmt.Println("Successfully refreshed and downloaded TUF metadata")
 
 	if !ouTufOnly {
+		if !isDstDirClean(dstDir) {
+			fmt.Println("Destination directory already has update data")
+			if !ouAllowMultipleTargets {
+				subcommands.DieNotNil(errors.New(
+					"Please provide a clean destination directory or re-run with --allow-multiple-targets to add a new target to a directory which already has update data.\n" +
+						"Notice that multiple targets in the same directory is only supported in LmP >= v92."))
```
vkhoroz, line 95: We usually use multi-line messages for this:
```suggestion
				subcommands.DieNotNil(errors.New(`
Provide a clean destination directory, or re-run with --allow-multiple-targets to add a new target to a directory which already has update data.
Notice that multiple targets in the same directory is only supported in LmP >= v92.`))
```

If you don't want an extra newline (usually it is OK) - move the opening backtick above to the line start.

Also, it is kind of strange to see a message which starts with `ERROR: Please`: it is not clear what was an error.
IIUC @kprosise would recommend to have no `Please` above, so I also included that suggestion.
Still, some error needs to be shown _after_ `ERROR:` and _before_ giving an instruction how to tackle it.

vkhoroz, line 95: Just IMHO this message is only needed for the `if !ouAllowMultipleTargets` code path.
I'm not strong on this, however, the error message seems cleaner in that case.
In contrast, when the command is run with `--allow-multiple-targets` , it might look confusing.

```suggestion
			if !ouAllowMultipleTargets {
				subcommands.DieNotNil(errors.New(`Destination directory already has update data.
Provide a clean destination directory or re-run with --allow-multiple-targets to add a new target to a directory which already has update data.
Notice that multiple targets in the same directory is only supported in LmP >= v92.`))
```

vkhoroz (approved):
LGTM

Only a couple of nits for error message.

subcommands/targets/offline-update.go
detsch: @vkhoroz , I've added your suggestions. Error now looks like this:
```
ERROR: Destination directory already has update data.
Provide a clean destination directory or re-run with --allow-multiple-targets to add a new target to a directory which already has update data.
Notice that multiple targets in the same directory is only supported in LmP >= v92.
```

PR #301, opened 2023-10-06
Cleanup the Makefile of unused and obsolete build targets

vkhoroz:


vkhoroz:
@doanac this is rather opinionated.

But we've made so many changes to Makefile recently, that it is worth finishing this story.
If you find any of these 3 commits irrelevant - I'll better remove those.

Makefile
```diff
@@ -8,19 +8,13 @@ linter:=$(shell which golangci-lint 2>/dev/null || echo $(HOME)/go/bin/golangci-
...
 
-fioctl-linux-amd64:
-fioctl-linux-arm64:
-fioctl-linux-armv7:
-fioctl-windows-amd64:
-fioctl-darwin-amd64:
-fioctl-darwin-arm64:
+# Any target of "fioctl-{GOOS}-{GOARCH}" can be built for your desired platform.
+# See "build" for a list of platforms supported and verified by Foundries.io.
+# Set GOARM={version} when compiling for specific Arm version; default is to auto-detect or 6 if cannot detect.
```
vkhoroz, line 17: Taken from here: https://go.dev/doc/install/source#environment
Maybe it is worth to reference it instead of copy-pasting a part of that content.

doanac: what's nice about having them is that you get auto-completion when running `make <tab>`

doanac (approved):

Makefile
vkhoroz: hmm... I never thought about that. Always doing `^R` instead.

vkhoroz: I can revert this one and only remove the armv7 (which does not work anyway).

vkhoroz: reverted in force-push

PR #303, opened 2023-10-06
Fix crash with targets show

doanac:
`fioctl targets show ""` causes the tool to crash. This helps it fail more cleanly.

### DIFF
--- a/subcommands/targets/show.go
+++ b/subcommands/targets/show.go
@@ -3,6 +3,7 @@ package targets
 import (
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -210,6 +211,10 @@ func doShowComposeApp(cmd *cobra.Command, args []string) {
 }
 
 func getTargets(factory string, prodTag string, version string) ([]string, map[string]string, map[string]client.TufCustom) {
+	if len(version) == 0 {
+		subcommands.DieNotNil(errors.New("Missing required argument <version>"))
+	}
+
 	var targets tuf.Files
 	var prodMeta *client.AtsTufTargets
 


vkhoroz (approved):
I am still puzzled about why cobra did not catch this.
Does it mean that a dozen of other commands can fail on this too?

doanac:
I think we probably have to set minlen for string args that require it

MrCry0 (approved):

vkhoroz:
@doanac I'm kind of fine with this as a quick fix for the _exact problem_.

Still, I'm a bit of worried about the generic use case: that we might end up fixing a few dozen of places like this here and there.
So, I was curious if we can apply some generic logic somewhere.

**TL;DR**
The first idea that came to my mind is: that we literally never expect a user to pass us arguments like `""` or `''`, or ever `" "`.
So, based on that assumption, I'd like us to have some generic check which iterates over all arguments and prevents that junk from entering.
Unfortunately, Cobra has no such attribute as `DenyEmptyArgs` or similar.
Fortunately, it has `PersistentPreRun & PersistentPreRunE`, which we can use to our savior.

The next idea is to put that check somewhere in the root command: https://github.com/foundriesio/fioctl/blob/main/cmd/root.go#L42 like below:
```
var rootCmd = &cobra.Command{
    ...
    PersistentPreRunE: func (cmd *cobra.Command, args []string) error {
        for pos, val := range args {
            if len(strings.Trim(val)) == 0 {
                // You can also use PersistentPreRun and subcommand.DieNotNil(...) here
                return fmt.Errorf("Empty values or values containing only white space are not allowed for positional argument at position %d", pos)
            }
        }
        return nil
    },
```

This way, every single command is covered by this check uniformly, and we don't need to worry about empty values anymore.
What do you think?

doanac:
@vkhoroz works for me. I'm not going to have a lot of time to hack this week. you want to close this out and push your own fix?

vkhoroz:
@doanac ok, let me try to craft smth by EOW. Then I will close this PR.

vkhoroz:
Closing this in favor of a more generic fix to all commands: https://github.com/foundriesio/fioctl/pull/312

PR #312, opened 2023-10-11
Bugfix: deny empty positional arguments in all commands

vkhoroz:
We expected this function to be provided by Cobra, but it does not make that check. So, before the some commands accepting positional arguments allowed values like "" or ''. These are affectively empty strings, that is equivalent to no value in all or commands.

As a matter of fact, some commands were panicking on this, some failing on API level with 400, 401, 404, 405, or 500 errors. The exact error which happened seems quite random. In the worst case this could lead to bad behavior, although I found no such use case.

This disallows erroneous (empty) positional arguments once and for all commands. All (random) misbehaving commands I tried are now fixed.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -40,8 +40,9 @@ var (
 )
 
 var rootCmd = &cobra.Command{
-	Use:   "fioctl",
-	Short: "Manage Foundries Factories",
+	Use:               "fioctl",
+	Short:             "Manage Foundries Factories",
+	PersistentPreRunE: rootArgValidation,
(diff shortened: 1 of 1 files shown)


vkhoroz:
A fix is a bit more complex because of https://github.com/spf13/cobra/issues/216

vkhoroz:
If Cobra mainteiners accept this fix: https://github.com/spf13/cobra/pull/2044 - the workaround may be removed.

doanac (approved):

PR #314, opened 2023-10-13
Fix El2Go and EST cert ops after switching to Golang based PKI

vkhoroz:
Signed-off-by: Volodymyr Khoroz <volodymyr.khoroz@foundries.io>

PR #315, opened 2023-10-13
Feature: TLS certificate rotation for the factory PKI

vkhoroz:
Supported operations:

- Rotate EST certificate (only cleanup, tested happy-path, TBD end-2-end).
- Rotate TLS certificate (new feature, tested end-2-end).

Device CA rotations will be a separate PR, as they are more complex.

### DIFF
--- a/subcommands/keys/ca_create.go
+++ b/subcommands/keys/ca_create.go
@@ -83,7 +83,7 @@ func doCreateCA(cmd *cobra.Command, args []string) {
 	logrus.Debugf("Create CA for %s under %s", factory, certsDir)
 	opts := client.CaCreateOptions{
 		FirstTimeInit:  true,
-		CreateOnlineCa: true,
+		CreateOnlineCa: createOnlineCA,
 		CreateTlsCert:  true,
 	}
 	csrs, err := api.FactoryCreateCA(factory, opts)
--- a/subcommands/keys/ca_create.go
+++ b/subcommands/keys/ca_create.go
@@ -8,6 +8,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
+	"github.com/foundriesio/fioctl/client"
 	"github.com/foundriesio/fioctl/subcommands"
 	"github.com/foundriesio/fioctl/x509"
 )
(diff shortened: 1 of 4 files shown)


vkhoroz:
This is blocked by https://github.com/foundriesio/fioctl/pull/314

Please, wait for that PR to be merged, as a major part of commits were rebased from there.

subcommands/el2g/gateway.go
```diff
@@ -26,17 +28,19 @@ func init() {
...
 	factory := viper.GetString("factory")
 
-	path := filepath.Join(pkiDir, "factory_ca.key")
-	_, err := os.Stat(path)
-	subcommands.DieNotNil(err)
-	path = filepath.Join(pkiDir, "sign_ca_csr")
-	_, err = os.Stat(path)
+	subcommands.DieNotNil(os.Chdir(pkiDir))
+	hsm, err := x509.ValidateHsmArgs(
+		hsmModule, hsmPin, hsmTokenLabel, "--hsm-module", "--hsm-pin", "--hsm-token-label")
```
StealthyCoder, line 41: From just reading this code, it now looks as if using a hsm is mandatory. However we then validate the args and when supplied with empty strings, then we still continue to pass a `nil` value that will get ignored in `x509.InitHSM`.

subcommands/keys/ca_rotate_tls.go
```diff
@@ -0,0 +1,51 @@
...
+	// HSM variables defined in ca_create.go
+	cmd.Flags().StringVarP(&hsmModule, "hsm-module", "", "", "Load a root CA key on a PKCS#11 compatible HSM using this module")
+	cmd.Flags().StringVarP(&hsmPin, "hsm-pin", "", "", "The PKCS#11 PIN to log into the HSM")
+	cmd.Flags().StringVarP(&hsmTokenLabel, "hsm-token-label", "", "", "The label of the HSM token containing the root CA key")
+}
+
+func doRotateTls(cmd *cobra.Command, args []string) {
+	factory := viper.GetString("factory")
+
+	subcommands.DieNotNil(os.Chdir(args[0]))
```
StealthyCoder, line 32: Does it matter where exactly this `Chdir` happens to? Can there be invalid targets?

subcommands/keys/ca_create.go
```diff
@@ -73,46 +73,43 @@ func getDeviceCaCommonName(factory string) string {
 func doCreateCA(cmd *cobra.Command, args []string) {
 	factory := viper.GetString("factory")
 	certsDir := args[0]
-	logrus.Debugf("Create CA for %s under %s", factory, certsDir)
+
 	subcommands.DieNotNil(os.Chdir(certsDir))
+	hsm, err := x509.ValidateHsmArgs(
+		hsmModule, hsmPin, hsmTokenLabel, "--hsm-module", "--hsm-pin", "--hsm-token-label")
+	subcommands.DieNotNil(err)
+	x509.InitHsm(hsm)
```
StealthyCoder, line 81: As this is the same everywhere, why not extract this into a common collected function and call it with the `args` needed? I just do not know what to call it, maybe like `SetupHSM`? Unsure what the golang conventions are.

subcommands/keys/ca_rotate_tls.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/spf13/viper"
+
+	"github.com/foundriesio/fioctl/client"
+	"github.com/foundriesio/fioctl/subcommands"
+	"github.com/foundriesio/fioctl/x509"
+)
+
+func init() {
+	cmd := &cobra.Command{
+		Use:   "rotate-tls <PKI Directory>",
```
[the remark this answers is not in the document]
vkhoroz: Yes, you've got it.

I started doing `--rotate-tls` and was envisioning `--rotate-est` etc etc.
But, then I tried to imagine how that would look like when combined with all the HSM options, and it cried at me.

So, I had to retreat and there is no wonder that I landed on your continent of having concise _to-the-point_ commands.

subcommands/el2g/gateway.go
vkhoroz: At first I wanted to do `if hsm != nil { x509.InitHsm(hsm) }`, but that increases the amount of code.
So I decided to trade off some clarity for space, that seems forgivable here.

subcommands/keys/ca_create.go
vkhoroz: That was my initial thought.

But, I _believe_ as soon as in Q1 2024 we will allow storing other keys on HSM, like e.g. the device CA key.
And, I'm 100% sure that users would want to store it on a different HSM device than the root CA key.
That led me to understanding that this command is going to grow with other `--hsm-foo` parameters, and we're going to have more than 1 `hsm` local variable in some commands.

I don't want to rewrite a lot of code again at that time; so I kind of prepare it for minimal future changes.

vkhoroz:
Rebased on main

vkhoroz:
This is now ready for the final review.
It works with the current API in prod.

doanac (approved):
tested andy-corp and worked on first try with no confusion.

subcommands/keys/ca_create.go
StealthyCoder: I understand. :+1:

PR #317, opened 2023-10-17
Two small bug fixes for customer support tickets

doanac:


subcommands/targets/tag.go
```diff
@@ -41,11 +41,13 @@ func init() {
...
-		unique[i] = item
-		items[item] = true
+	for _, item := range a {
+		if ok := items[item]; !ok {
+			unique = append(unique, item)
+			items[item] = true
+		}
 	}
 	for _, item := range b {
 		if ok := items[item]; !ok {
```
vkhoroz, line 53: I feel like that second loop below should also have the `items[item] = true` logic to filter duplicates in it, shouldn't it.

Then, I'd aggregate the two into an outer loop like this:
```
for i, lst := 0, a; i < 2; i, lst = i+1, b {
	for _, item := range lst {
		if ok := items[item]; !ok {
			unique = append(unique, item)
			items[item] = true
		}
	}
}
```

vkhoroz: FWIW it looks like a good candidate to use https://pkg.go.dev/github.com/mpvl/unique#Strings and https://pkg.go.dev/github.com/mpvl/unique#StringsAreUnique.

vkhoroz (approved):
LGTM

subcommands/targets/tag.go
doanac: i wasn't wanting to pull in another library, look at the code license,etc. But your loop idea looks great. force pushed

PR #324, opened 2023-10-20
Feature: allow to revoke or disable device CA

vkhoroz:
Currently, an API allows to simply remove the device CA from a list of CAs.

This change implements the device CA removal via the X.509 certificate revocation.
That way, a user is required to crypto-sign the operation using the factory root CA.
Thus, we both obtain an cryptographic conscent from the user for this operation and protect against several scenarios:
- An accidental device CA removal by passing a wrong file to "keys ca update".
- An intentional device CA removal by bad admin who does not possess the factory root CA.

After this change, the API will be changed to require the "ca-crt" field to contain all device CAs which:
- Are already present in the "ca-crt" list on the backend.
- And are not explicitly listed in the "ca-revoke-crl" list.

There are two new operations introduced:
- *disable device CA*: Tell the backend to disallow registering new devices using this CA; already registered devices (using this CA) can still connect and use FoundriesFactory.
- *revoke device CA*: Tell the backend to remove this CA from the list of trusted device CAs; devices with client certs issued by this CA can no longer connect and use FoundriesFactory.

### DIFF
--- a/subcommands/keys/ca_revoke_device_ca.go
+++ b/subcommands/keys/ca_revoke_device_ca.go
@@ -137,6 +137,19 @@ func doRevokeDeviceCa(cmd *cobra.Command, args []string) {
 		ca := x509.LoadCertFromFile(filename)
 		toRevoke[ca.SerialNumber.Text(10)] = crlReason
 	}
+
+	caList, err := api.FactoryGetCA(factory)
+	subcommands.DieNotNil(err)
+	validSerials := make(map[string]bool, 0)
+	for _, c := range parseCertList(caList.CaCrt) {
+		validSerials[c.SerialNumber.Text(10)] = true
+	}
+	for serial := range toRevoke {
+		if _, ok := validSerials[serial]; !ok {
+			subcommands.DieNotNil(fmt.Errorf("There is no actual device CA with serial %s", serial))
+		}
+	}
+
 	fmt.Println("Signing CRL by factory root CA")
 	certs := client.CaCerts{CaRevokeCrl: x509.CreateCrl(toRevoke)}
 	fmt.Println("Uploading CRL to Foundries.io")
--- a/subcommands/keys/ca_revoke_device_ca.go
+++ b/subcommands/keys/ca_revoke_device_ca.go
@@ -14,8 +14,14 @@ import (
 	"github.com/foundriesio/fioctl/x509"
 )
 
+const (
+	crlCmdAnnotation      = "crl-revoke"
+	crlCmdRevokeDeviceCa  = "revoke-device-ca"
+	crlCmdDisableDeviceCa = "disable-device-ca"
+)
+
 func init() {
-	cmd := &cobra.Command{
+	revokeCmd := &cobra.Command{
 		Use:   "revoke-device-ca <PKI Directory>",
 		Short: "Revoke device CA, so that devices with client certificates it issued can no longer connect to your Factory",
 		Run:   doRevokeDeviceCa,
@@ -44,16 +50,51 @@ fioctl keys ca revoke-device-ca /path/to/pki/dir --ca-file local-ca \
 
 # Show a generated CRL that would be sent to the server to revoke a local device CA, without actually revoking it.
 fioctl keys ca revoke-device-ca /path/to/pki/dir --ca-file local-ca --dry-run --pretty`,
+		Annotations: map[string]string{crlCmdAnnotation: crlCmdRevokeDeviceCa},
 	}
-	caCmd.AddCommand(cmd)
+	caCmd.AddCommand(revokeCmd)
+	addRevokeCmdFlags(revokeCmd, "revoke")
+
+	disableCmd := &cobra.Command{
+		Use:   "disable-device-ca <PKI Directory>",
+		Short: "Disable device CA, so that new devices with client certificates it issued can no longer be registered",
+		Run:   doRevokeDeviceCa,
+		Args:  cobra.ExactArgs(1),
+		Long: `Disable device CA, so that new devices with client certificates it issued can no longer be registered.
+
+When the online or local device CA is disabled:
+- It is no longer possible to register new devices with client certificates it had issued.
+- Existing devices with client certificates it issued may continue to connect and use your Factory.
+
+Usually, when the device CA is compromised, a user should:
+1. Immediately disable a given device CA using "fioctl keys ca disable-device-ca <PKI Directory> --serial <CA Serial>".
+2. Inspect their devices with client certificates issued by that device CA, and remove compromised devices (see "fioctl devices list|delete").
+3. Create a new device CA using "fioctl keys ca add-device-ca <PKI Directory> --online-ca|--local-ca".
+4. Rotate a client certificate of legitimate devices to the certificate issued by a new device CA (see "fioctl devices config rotate-certs").
+5. Revoke a given device CA using "fioctl keys ca revoke-device-ca <PKI Directory> --serial <CA Serial>".`,
+		Example: `
+# Disable two device CAs given their serial numbers:
+fioctl keys ca disable-device-ca /path/to/pki/dir --ca-serial <base-10-serial-1> --ca-file <base-10-serial-2>
+
+# Show a generated CRL that would be sent to the server to disable a local device CA, without actually disabling it.
+fioctl keys ca disable-device-ca /path/to/pki/dir --ca-file local-ca --dry-run --pretty
+
+# See "fioctl keys ca revoke-device-ca --help" for more examples; these two commands have a very similar syntax.`,
+		Annotations: map[string]string{crlCmdAnnotation: crlCmdDisableDeviceCa},
+	}
+	caCmd.AddCommand(disableCmd)
+	addRevokeCmdFlags(disableCmd, "disable")
+}
+
+func addRevokeCmdFlags(cmd *cobra.Command, op string) {
 	cmd.Flags().BoolP("dry-run", "", false,
-		"Do not revoke the certificate, but instead show a generated CRL that will be uploaded to the server.")
+		"Do not "+op+" the certificate, but instead show a generated CRL that will be uploaded to the server.")
 	cmd.Flags().BoolP("pretty", "", false,
 		"Can be used with dry-run to show the generated CRL in a pretty format.")
 	cmd.Flags().StringArrayP("ca-file", "", nil,
-		"A file name of the device CA to revoke. Can be used multiple times to revoke several device CAs.")
+		"A file name of the device CA to "+op+". Can be used multiple times to "+op+" several device CAs")
 	cmd.Flags().StringArrayP("ca-serial", "", nil,
-		"A serial number (base 10) of the device CA to revoke. Can be used multiple times to revoke several device CAs.")
+		"A serial number (base 10) of the device CA to "+op+". Can be used multiple times to "+op+" several device CAs")
 	_ = cmd.MarkFlagFilename("ca-file")
 	// HSM variables defined in ca_create.go
 	cmd.Flags().StringVarP(&hsmModule, "hsm-module", "", "", "Load a root CA key from a PKCS#11 compatible HSM using this module")
@@ -68,6 +109,10 @@ func doRevokeDeviceCa(cmd *cobra.Command, args []string) {
 	// pretty, _ := cmd.Flags().GetBool("pretty")
 	caFiles, _ := cmd.Flags().GetStringArray("ca-file")
 	caSerials, _ := cmd.Flags().GetStringArray("ca-serial")
+	crlReason := map[string]int{
+		crlCmdRevokeDeviceCa:  x509.CrlCaRevoke,
+		crlCmdDisableDeviceCa: x509.CrlCaDisable,
+	}[cmd.Annotations[crlCmdAnnotation]]
 
 	if len(caFiles)+len(caSerials) == 0 {
 		subcommands.DieNotNil(errors.New("At least one of --ca-file or --ca-serial must be provided"))
@@ -86,11 +131,11 @@ func doRevokeDeviceCa(cmd *cobra.Command, args []string) {
 		if _, ok := num.SetString(serial, 10); !ok {
 			subcommands.DieNotNil(fmt.Errorf("Value %s is not a valid base 10 serial", serial))
 		}
-		toRevoke[serial] = x509.CrlCaRevoke
+		toRevoke[serial] = crlReason
 	}
 	for _, filename := range caFiles {
 		ca := x509.LoadCertFromFile(filename)
-		toRevoke[ca.SerialNumber.Text(10)] = x509.CrlCaRevoke
+		toRevoke[ca.SerialNumber.Text(10)] = crlReason
 	}
 	fmt.Println("Signing CRL by factory root CA")
 	certs := client.CaCerts{CaRevokeCrl: x509.CreateCrl(toRevoke)}
--- a/subcommands/keys/ca_revoke_device_ca.go
+++ b/subcommands/keys/ca_revoke_device_ca.go
@@ -1,0 +1,99 @@
+package keys
+
+import (
+	"errors"
+	"fmt"
+	"math/big"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/foundriesio/fioctl/client"
+	"github.com/foundriesio/fioctl/subcommands"
+	"github.com/foundriesio/fioctl/x509"
+)
+
+func init() {
+	cmd := &cobra.Command{
+		Use:   "revoke-device-ca <PKI Directory>",
+		Short: "Revoke device CA, so that devices with client certificates it issued can no longer connect to your Factory",
+		Run:   doRevokeDeviceCa,
+		Args:  cobra.ExactArgs(1),
+		Long: `Revoke device CA, so that devices with client certificates it issued can no longer connect to your Factory.
+
+When the online or local device CA is revoked:
+- It is no longer possible to register new devices with client certificates it had issued.
+- Existing devices with client certificates it issued can no longer connect to your Factory.
+
+You may later re-add a revoked device CA using "keys ca update", if you still keep its certificate stored somewhere.
+Once you do this, devices with client certificates issued by this device CA may connect to your Factory again.`,
+		Example: `
+# Revoke a local device CA by providing a (default) file name inside your PKI directory:
+fioctl keys ca revoke-device-ca /path/to/pki/dir --ca-file local-ca
+
+# Revoke two local device CAs given a full path to their files:
+fioctl keys ca revoke-device-ca /path/to/pki/dir --ca-file /path/to/ca1.pem --ca-file /path/to/ca2.crt
+
+# Revoke two device CAs given their serial numbers:
+fioctl keys ca revoke-device-ca /path/to/pki/dir --ca-serial <base-10-serial-1> --ca-file <base-10-serial-2>
+
+# Revoke a local device CA, when your factory root CA private key is stored on an HSM:
+fioctl keys ca revoke-device-ca /path/to/pki/dir --ca-file local-ca \
+  --hsm-module /path/to/pkcs11-module.so --hsm-pin 1234 --hsm-token-label <token-label-for-key>
+
+# Show a generated CRL that would be sent to the server to revoke a local device CA, without actually revoking it.
+fioctl keys ca revoke-device-ca /path/to/pki/dir --ca-file local-ca --dry-run --pretty`,
+	}
+	caCmd.AddCommand(cmd)
+	cmd.Flags().BoolP("dry-run", "", false,
+		"Do not revoke the certificate, but instead show a generated CRL that will be uploaded to the server.")
+	cmd.Flags().BoolP("pretty", "", false,
+		"Can be used with dry-run to show the generated CRL in a pretty format.")
+	cmd.Flags().StringArrayP("ca-file", "", nil,
+		"A file name of the device CA to revoke. Can be used multiple times to revoke several device CAs.")
+	cmd.Flags().StringArrayP("ca-serial", "", nil,
+		"A serial number (base 10) of the device CA to revoke. Can be used multiple times to revoke several device CAs.")
+	_ = cmd.MarkFlagFilename("ca-file")
+	// HSM variables defined in ca_create.go
+	cmd.Flags().StringVarP(&hsmModule, "hsm-module", "", "", "Load a root CA key from a PKCS#11 compatible HSM using this module")
+	cmd.Flags().StringVarP(&hsmPin, "hsm-pin", "", "", "The PKCS#11 PIN to log into the HSM")
+	cmd.Flags().StringVarP(&hsmTokenLabel, "hsm-token-label", "", "", "The label of the HSM token containing the root CA key")
+}
+
+func doRevokeDeviceCa(cmd *cobra.Command, args []string) {
+	factory := viper.GetString("factory")
+	// TODO: Implement the --dry-run and --pretty arguments
+	// dryRun, _ := cmd.Flags().GetBool("dry-run")
+	// pretty, _ := cmd.Flags().GetBool("pretty")
+	caFiles, _ := cmd.Flags().GetStringArray("ca-file")
+	caSerials, _ := cmd.Flags().GetStringArray("ca-serial")
+
+	if len(caFiles)+len(caSerials) == 0 {
+		subcommands.DieNotNil(errors.New("At least one of --ca-file or --ca-serial must be provided"))
+	}
+
+	subcommands.DieNotNil(os.Chdir(args[0]))
+	hsm, err := x509.ValidateHsmArgs(
+		hsmModule, hsmPin, hsmTokenLabel, "--hsm-module", "--hsm-pin", "--hsm-token-label")
+	subcommands.DieNotNil(err)
+	x509.InitHsm(hsm)
+
+	fmt.Println("Generating Certificate Revocation List")
+	toRevoke := make(map[string]int, len(caSerials)+len(caFiles))
+	for _, serial := range caSerials {
+		num := new(big.Int)
+		if _, ok := num.SetString(serial, 10); !ok {
+			subcommands.DieNotNil(fmt.Errorf("Value %s is not a valid base 10 serial", serial))
+		}
+		toRevoke[serial] = x509.CrlCaRevoke
+	}
+	for _, filename := range caFiles {
+		ca := x509.LoadCertFromFile(filename)
+		toRevoke[ca.SerialNumber.Text(10)] = x509.CrlCaRevoke
+	}
+	fmt.Println("Signing CRL by factory root CA")
+	certs := client.CaCerts{CaRevokeCrl: x509.CreateCrl(toRevoke)}
+	fmt.Println("Uploading CRL to Foundries.io")
+	subcommands.DieNotNil(api.FactoryPatchCA(factory, certs))
+}
--- a/subcommands/keys/ca_revoke_device_ca.go
+++ b/subcommands/keys/ca_revoke_device_ca.go
@@ -1,6 +1,10 @@
 package keys
 
 import (
+	x509Lib "crypto/x509"
+	"encoding/asn1"
+	"encoding/hex"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"math/big"
@@ -104,9 +108,8 @@ func addRevokeCmdFlags(cmd *cobra.Command, op string) {
 
 func doRevokeDeviceCa(cmd *cobra.Command, args []string) {
 	factory := viper.GetString("factory")
-	// TODO: Implement the --dry-run and --pretty arguments
-	// dryRun, _ := cmd.Flags().GetBool("dry-run")
-	// pretty, _ := cmd.Flags().GetBool("pretty")
+	dryRun, _ := cmd.Flags().GetBool("dry-run")
+	pretty, _ := cmd.Flags().GetBool("pretty")
 	caFiles, _ := cmd.Flags().GetStringArray("ca-file")
 	caSerials, _ := cmd.Flags().GetStringArray("ca-serial")
 	crlReason := map[string]int{
@@ -152,6 +155,81 @@ func doRevokeDeviceCa(cmd *cobra.Command, args []string) {
 
 	fmt.Println("Signing CRL by factory root CA")
 	certs := client.CaCerts{CaRevokeCrl: x509.CreateCrl(toRevoke)}
+
+	if dryRun {
+		fmt.Println(certs.CaRevokeCrl)
+		if pretty {
+			prettyPrintCrl(certs.CaRevokeCrl)
+		}
+		return
+	}
+
 	fmt.Println("Uploading CRL to Foundries.io")
 	subcommands.DieNotNil(api.FactoryPatchCA(factory, certs))
+}
+
+func prettyPrintCrl(crlPem string) {
+	block, remaining := pem.Decode([]byte(crlPem))
+	if block == nil || len(remaining) > 0 {
+		subcommands.DieNotNil(errors.New("Invalid PEM block"), "Failed to parse generated CRL:")
+		return // linter
+	}
+	c, err := x509Lib.ParseRevocationList(block.Bytes)
+	subcommands.DieNotNil(err, "Failed to parse generated CRL:")
+	fmt.Println("Certificate Revocation List:")
+	fmt.Println("\tIssuer:", c.Issuer)
+	fmt.Println("\tValidity:")
+	fmt.Println("\t\tNot Before:", c.ThisUpdate)
+	fmt.Println("\t\tNot After:", c.NextUpdate)
+	fmt.Println("\tSignature Algorithm:", c.SignatureAlgorithm)
+	fmt.Println("\tSignature:", hex.EncodeToString(c.Signature))
+	fmt.Println("\tRevoked Certificates:")
+	for _, crt := range c.RevokedCertificates {
+		fmt.Println("\t\tSerial:", crt.SerialNumber.Text(10))
+		fmt.Println("\t\t\tRevocation Date:", crt.RevocationTime)
+		if len(crt.Extensions) > 0 {
+			fmt.Println("\t\t\tExtensions:")
+			for _, ext := range crt.Extensions {
+				if ext.Id.String() == "2.5.29.21" {
+					fmt.Print("\t\t\t\tx509v3 Reason Code:")
+					if ext.Critical {
+						fmt.Print("(critical)")
+					}
+					var val asn1.Enumerated
+					if _, err := asn1.Unmarshal(ext.Value, &val); err != nil {
+						fmt.Println(err)
+					} else {
+						readable := map[int]string{
+							x509.CrlCaRevoke:  "Revoke",
+							x509.CrlCaDisable: "Disable",
+							x509.CrlCaRenew:   "Renew",
+						}[int(val)]
+						fmt.Println("\n\t\t\t\t\t", readable, "-", val)
+					}
+				} else {
+					fmt.Println("\t\t\t\tUnknown OID", ext.Id.String())
+				}
+			}
+		}
+	}
+	if len(c.Extensions) > 0 {
+		fmt.Println("\tExtensions:")
+		for _, ext := range c.Extensions {
+			if ext.Id.String() == "2.5.29.35" {
+				fmt.Print("\t\tx509v3 Authority Key Id: ")
+				if ext.Critical {
+					fmt.Print("(critical)")
+				}
+				fmt.Println("\n\t\t\t", hex.EncodeToString(c.AuthorityKeyId))
+			} else if ext.Id.String() == "2.5.29.20" {
+				fmt.Print("\t\tx509v3 CRL Number: ")
+				if ext.Critical {
+					fmt.Print("(critical)")
+				}
+				fmt.Println("\n\t\t\t", c.Number)
+			} else {
+				fmt.Println("\t\tUnknown OID", ext.Id.String())
+			}
+		}
+	}
 }
--- a/subcommands/keys/ca_show.go
+++ b/subcommands/keys/ca_show.go
@@ -58,6 +58,7 @@ func doShowCA(cmd *cobra.Command, args []string) {
 			printOneCert(resp.TlsCrt)
 		case justShowCas:
 			printOneCert(resp.CaCrt)
+			printDisabledCas(resp.CaDisabled)
 		default:
 			panic("Unknown flag: " + flag)
 		}
@@ -82,6 +83,7 @@ func doShowCA(cmd *cobra.Command, args []string) {
 	printOneCert(resp.TlsCrt)
 	fmt.Println("\n## Device Authentication Certificate(s)")
 	printOneCert(resp.CaCrt)
+	printDisabledCas(resp.CaDisabled)
 }
 
 func printOneCert(crt string) {
@@ -89,6 +91,15 @@ func printOneCert(crt string) {
 		prettyPrint(crt)
 	} else {
 		fmt.Println(crt)
+	}
+}
+
+func printDisabledCas(serials []string) {
+	if len(serials) > 0 {
+		fmt.Println("\n## Disabled Device Authentication Certificate Serial(s)")
+		for _, num := range serials {
+			fmt.Println(" - ", num)
+		}
 	}
 }
 
--- a/subcommands/keys/ca_show.go
+++ b/subcommands/keys/ca_show.go
@@ -152,23 +152,30 @@ func extKeyUsage(ext []x509.ExtKeyUsage) string {
 	return vals
 }
 
-func prettyPrint(cert string) {
-	for len(cert) > 0 {
-		block, remaining := pem.Decode([]byte(cert))
+func parseCertList(pemData string) (certs []*x509.Certificate) {
+	for len(pemData) > 0 {
+		block, remaining := pem.Decode([]byte(pemData))
 		if block == nil {
 			// could be excessive whitespace
-			if cert = strings.TrimSpace(string(remaining)); len(cert) == len(remaining) {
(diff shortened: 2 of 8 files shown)


x509/common.go
```diff
@@ -19,6 +22,23 @@ const (
...
+	// This action is permanent.
+	CrlCaRevoke = 9 // RFC 5280 - privilegeWithdrawn
+
+	// Disable the device CA, so that no new devices can be registered with client certificates issued by this CA.
+	// Devices that were already registered can still connect and operate as normal.
+	// This action can be reverted by CrlCaRenew.
+	CrlCaDisable = 6 // RFC 5280 - certificateHold
+
+	// Renew the previously disabled device CA, so that new device registrations are possible again.
+	CrlCaRenew = 8 // RFC 5280 - removeFromCRL
```
vkhoroz, line 39: I'm not sure that I want to drag this whole baggage right now.
Probably, I will simply define the `CrlCaRevoke` as internal constant inside `x509/golang.go` and change the `CreateCrl` signature to accept a list of serials.

This is here just to show you how we can further improve our functionality going forward.
Although, there is a small chance that after walking a dog on this, you would come up and tell me that this is a must have.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, serial := range caSerials {
+		num := new(big.Int)
+		if _, ok := num.SetString(serial, 10); !ok {
+			subcommands.DieNotNil(fmt.Errorf("Value %s is not a valid base 10 serial", serial))
+		}
+		toRevoke[serial] = x509.CrlCaRevoke
+	}
+	for _, filename := range caFiles {
+		ca := x509.LoadCertFromFile(filename)
+		toRevoke[ca.SerialNumber.Text(10)] = x509.CrlCaRevoke
```
StealthyCoder, line 77: Why is it specifically 10? And always 10?

x509/bash.go
```diff
@@ -159,6 +159,13 @@ func SignEl2GoCsr(csrPem string) string {
 	return signCaCsr("el2g-*", csrPem)
 }
 
+func CreateCrl(serials map[string]int) string {
+	if true {
+		panic("This function is not implemented in Bash implementation")
+	}
```
StealthyCoder, line 165: Do we even need the `if true` can we not just straight call `panic` ?

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	fmt.Println("Signing CRL by factory root CA")
+	certs := client.CaCerts{CaRevokeCrl: x509.CreateCrl(toRevoke)}
```
mike-sul, line 80: The commit message says:

```
After this change, the API will be changed to require the "ca-crt" field to contain all device CAs which:
- Are already present in the "ca-crt" list on the backend.
```
Does it mean that the line #80 will be updated so the `CaCrt: <>` is added to the `certs`?

mike-sul:
LGTM. It makes sense to me. I'll finalize this PR review after the corresponding API PR is posted.

mike-sul:
If a user wants to add a new device CA then maybe it would be more convenient for a user to add just a new device ca to the `"ca-crt"` field, instead of 
```
After this change, the API will be changed to require the "ca-crt" field to contain all device CAs which:

    Are already present in the "ca-crt" list on the backend.
```
?

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: It is base 10, and it will always be like this.

The reason: old versions of Fioctl picked base 10 as an arbitrary value to print the certificate serial number.
So, I leave it that way, and set it in stone.

x509/bash.go
vkhoroz: Without this `if` statement, Golang will inline the panic and produce compiler warnings about unreachable code at random places.
I _believe_ it still inlines this function with the `if true`, but those warnings are then gone.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: I'm on the fence here.

Initially, I planned that Fioctl will only upload the CRL list, and expect the API to do the cleanup (remove revoked certificates).
This is more convenient from the user perspective if they want to integrate with the API directly, and perfectly justifiable for the PoC.

A marginally safer way would be to require the user to provide both the CRL and the new CA list, so that the API can double-check.
I'm not sure that effort to complicate the client is worth the _doubtful_ benefit.

Still, when the user uploads a new CA list - the API will be updated to verify that no CA items are removed from it, unless the user provides a CRL revoking those items _in the same request_.

So, I think the API will support 3 modes:
1. Only CA list provided - API will validate that it must contain all old CA items, and optionally some new CA items.
2. Only CRL is provided - API will simply remove revoked CA items from the CA list.
3. Both CRL and CA list are provided - API will validate that CA list must contain all old CA items minus revoked ones, and optionally some new CA items.

Does it make sense to you?

vkhoroz:
> If a user wants to add a new device CA then maybe it would be more convenient for a user to add just a new device ca to the `"ca-crt"` field ...

The API will continue to require the client to upload the full CA list, as it is now.
But, the Fioctl allows the user to add just one CA using `fioctl keys ca add-device-ca` command.

What is changed:
- currently, the user can overwrite the list of device CAs with some garbage by mistake.
- after the change that will be no longer possible, as that list becomes append-only; with remove operation being accomplished via the crypto-signed CRL.

mike-sul:
> The API will continue to require the client to upload the full CA list

Not sure it's convenient from a user perspective since it requires to fetch the online device CA before making the update request.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: This is how an API currently works - it requires a full CA list, not just new CAs to add.
I'm not going to change that.

The change will only be made to the validation part, to verify that the user does not mistakenly remove some old CAs from the list.

FWIW from the user perspective it is irrelevant, as it is the API protocol, not the UI interface.
The `fioctl keys ca add-device-ca` allows to add just one new CA, and handles the list mangling internally.

vkhoroz: @mike-sul I agree that might be more convenient for the API integrators.

But, that would make our API backward-incompatible with existing clients. I don't think the risk is worth the gain.

mike-sul: Not sure it will make the API backward incompatible since we are reducing the requirement (current API requires the whole list, and the future potential API can support both the whole and just new device CA list).

vkhoroz: OK, let's assume that the current CA list contains `A, B, C` and the API received a new list `C, D`.
In this case, should the API assume that it is a mistaken list (missing `A, B`) from the old client or an intent to add `D` from the new client?

The correct way would be to introduce a new property, and support both old and new property.
Either way, as I said above, that increase in the API complexity is not justifiable compared to the marginally easier way to integrate with the new API.

It might also bring the confusion to the user why the `CA list` property of the `PUT` request allows either full list or partial list.
I prefer simplicity and determinism over imaginable flexibility _nobody is really going to leverage_.

mike-sul: >from the old client or an intent to add D >from the new client?

I would consider it as adding D.

That's fine, however I have different opinion on this ("that increase in the API complexity" in my world it's opposite, right now the API is more complex since it requires gathering of all device CAs in order to add a new one as well as making a request to the backend to obtain online CA).

vkhoroz: I think we use different meanings for _complex_ vs _flexibility_.

What you said, is that the API is less _flexible_, because it only supports one kind of an operation - full replacement - which means that the Fioctl is marginally more _complex_ because the command to add a new CA is 2 lines of code longer.

But, that allows to keep the internal API _complexity_ at a much lower rate, because, well, it only needs to serve one command instead of the whole range between full replacement and addition.
That saving of _complexity_ becomes even more evident when we consider the amount of unit tests needed to cover each choice.

Note: we still wrote less text here than is needed to modify an API to achieve the _flexibility_ you desire.

x509/common.go
vkhoroz: @doanac do you think it is worth pursuing this _disable but not revoke device CA_ agenda now or later.

I'm working on the API side ATM, and will scrap this code once I test this end-2-end; unless we make a decision that we need it right now.

doanac: Disable but not revoke is the feature I'm interested in. I get how CRL is useful for lots of *clients*. But for client certs in a real factory, I don't think its very useful. How often in practice is someone going to want to say, "remove this CA and never trust the potentially hundreds of devices it signed"? If I lose control over my CA, I'd want to:

 1) tell our backend to stop trusting new devices ASAP
 2) look at devices in the backend signed with this CA and try to decide if they are legit or not. If not legit, delete the device

vkhoroz: Cool, then I see 2 options how to proceed.

The background:
- The _revoke_ feature only affects one Fioctl command and one API endpoint, and is fairly easy to implement.
- The _disable_ feature adds little complexity to the Fioctl on top of the above; but it requires additional changes to registration handlers in the API and device gateway.

So, our options are:
A. Do the _revoke_ part now, and do the _disable_ part in Q1, as I have other priority items for Q4.
B. Do both _revoke_ and _disable_ features now, ETA is a few days end-2-end.

I _believe_ we still have time in reserve for option B. @doanac what do you advice?

doanac: We could probably argue your other priorities are higher than either of these options. My thinking is "revoke" can be done now - I've done it many times working on AWS and Azure IOT integrations. Its just messy/manual. But our other priorities are for things that can't be done without changes.

That said, "disable" is something we *can't* do. So if you can do disable w/o a lot of disruption - I'm fine.

x509/bash.go
StealthyCoder: I did not know that, thanks. :+1:

MrCry0: > Without this `if` statement, Golang will inline the panic and produce compiler warnings about unreachable code at random places. I _believe_ it still inlines this function with the `if true`, but those warnings are then gone.

I'm not that good in Go, but just a shot: have you tried to use a block code without a clause?
`{ panic(...) }`. IMHO it looks clearer.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,83 @@
...
+		Run:   doRevokeDeviceCa,
+		Args:  cobra.ExactArgs(1),
+		Long: `Revoke device CA from the list of CAs allowed to issue device client certificates.
+
+When the online or local device CA is revoked:
+- It is no longer possible to register new devices with client certificates issued by this device CA.
+- Existing devices with client certificates issued by this device CA can no longer connect to your FoundriesFactory.
+
+You may later re-add a revoked device CA using "keys ca update", if you still keep its certificate stored somewhere.
+Once you do it, devices with client certificates issued by this device CA may connect to your FoundriesFactory again.`,
```
camilamacedo86, line 30: So, lets see if I understood this one:

- a) **First we create the CA (PKI)**: By running `fioctl keys ca create /path/to/cert/dirs -f <factory>`
- b) **Then, we can remove the CA(PKI)**: (which is introduced here) by running `fioctl key revoke-device-ca /path/to/cert/dirs -f <factory>` will revoke all that was scaffold in the dir with the first command. Would that mean:

     - All devices using this CA will lose update capabilities. 
     - Thus, revoking a CA in a production environment has the potential to affect all associated devices. 
     
      Can you please confirm my understanding here?

- c) **We can revert, restoring from Backup:** If a backup of the certificates is available, it seems we can restore them using `fioctl keys ca update <file>`. The documentation isn't entirely clear on whether we need to update all certificate files (factory_ca.key, factory_ca.pem, local-ca.key, local-ca.pem, online-crt, and tls-crt) individually, or if a subset will suffice. Any insights on this would be appreciated. See, from the fioctl keys ca update --help command:

```
Update the list of CAs that can create client certificates for devices

Usage:
  fioctl keys ca update <ca-crts file> [flags]
```

x509/bash.go
vkhoroz: @MrCry0 let me give it a shot; it looks interesting.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: * b) This command would only revoke a user-specified device CA (by file name or serial). Other device CAs are unaffected; factory root CA cannot be revoked. Then - yes, _devices using this CA will lose connectivity to FoundriesFactory_. However, bear in mind, that this command does not remove CA file on disk; it only removes the device CA on backend.
* c) Yes, a user can revert this operation _in case if they still possess the CA file they revoked_.
   For that the user would append that CA file to the list of allowed device CAs and upload it using `fioctl keys ca update`.
* e) No, the `fioctl keys ca create` can only be run once for a factory.

vkhoroz:
@doanac @mike-sul @StealthyCoder I'm almost done with this one.

I will take a look why PKI tests failed; they were passing for me locally.
A commit list is not ideal yet; so don't rush reviewing this - I'll clean it up tomorrow.
Also, I did not yet test this end-2-end fully - will continue tomorrow as well.

vkhoroz:
@doanac @mike-sul @StealthyCoder @camilamacedo86 

This was tested with the API version currently deployed to our staging MEDS instance.
I added and tested all planned featurettes:
- Device CA can be disabled or revoked.
- Dry-run and pretty options work as expected.
- When disabled, device CA shows as disabled in `keys ca show`.
- When disabled, I can no longer register devices _using API_ with this CA.
- When revoked, devices using this CA cannot connect to device gateway, failing mTLS handshake.

What still needs to be implemented:
- The device gateway part for the auto-registration is pending.

x509/common_test.go
```diff
@@ -201,7 +201,7 @@ func runTest(t *testing.T, verifyFiles func(factoryCa, tlsCert, onlineCa, offlin
 		crl, err := x509.ParseRevocationList(pemToDer(t, crlPem))
 		require.Nil(t, err)
 		require.NotNil(t, crl)
-		assert.Equal(t, factoryCa.Subject, crl.Issuer)
+		assert.Equal(t, factoryCa.Subject.String(), crl.Issuer.String())
```
vkhoroz, line 204: This check always passes for me locally either way.
But due to some CI magic, Golang always reorders pkix.Name attributes of the CRL subject to be reversed.

So, you might see a couple of random commits trying to fix this line.

vkhoroz: Passed on 3 re-runs, so I'm going to squash and consider it fixed.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,224 @@
...
+		subcommands.DieNotNil(errors.New("At least one of --ca-file or --ca-serial must be provided"))
+	}
+
+	subcommands.DieNotNil(os.Chdir(args[0]))
+	hsm, err := x509.ValidateHsmArgs(
+		hsmModule, hsmPin, hsmTokenLabel, "--hsm-module", "--hsm-pin", "--hsm-token-label")
+	subcommands.DieNotNil(err)
+	x509.InitHsm(hsm)
+
+	fmt.Println("Generating Cartificate Revokation List")
```
detsch, line 107:
```suggestion
	fmt.Println("Generating Certificate Revocation List")
```

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,224 @@
...
+	addRevokeCmdFlags(disableCmd, "disable")
+}
+
+func addRevokeCmdFlags(cmd *cobra.Command, op string) {
+	cmd.Flags().BoolP("dry-run", "", false,
+		"Do not "+op+" the certificate, but instead show a generated CRL that will be uploaded to the server.")
+	cmd.Flags().BoolP("pretty", "", false,
+		"Can be used with dry-run to show the generated CRL in a pretty format.")
+	cmd.Flags().StringArrayP("ca-file", "", nil,
+		"A file name of the device CA to "+op+". Can be used several times to "+op+" several device CAs")
```
detsch, line 76:
```suggestion
		"A file name of the device CA to "+op+". Can be used multiple times to "+op+" multiple device CAs")
```

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,224 @@
...
+	cmd.Flags().StringArrayP("ca-serial", "", nil,
+		"A serial number (base 10) of the device CA to "+op+". Can be used several times to "+op+" several device CAs")
```
detsch, line 78:
```suggestion
		"A serial number (base 10) of the device CA to "+op+". Can be used multiple times to "+op+" multiple device CAs")
```

subcommands/keys/ca_revoke_device_ca.go
detsch: Not sure about that suggestion. "Multiple" sounds more natural to me, but after some googling, looks like both terms would be OK, just a matter of style

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,83 @@
...
+	cmd.Flags().StringArrayP("ca-file", "", nil,
+		"A file name of the device CA to revoke. Can be used several times to revoke several device CAs")
+	cmd.Flags().StringArrayP("ca-serial", "", nil,
+		"A serial number (base 10) of the device CA to revoke. Can be used several times to revoke several device CAs")
+	_ = cmd.MarkFlagFilename("ca-file")
+	// HSM variables defined in ca_create.go
+	cmd.Flags().StringVarP(&hsmModule, "hsm-module", "", "", "Load a root CA key from a PKCS#11 compatible HSM using this module")
+	cmd.Flags().StringVarP(&hsmPin, "hsm-pin", "", "", "The PKCS#11 PIN to log into the HSM")
+	cmd.Flags().StringVarP(&hsmTokenLabel, "hsm-token-label", "", "", "The label of the HSM token containing the root CA key")
+}
```
doanac, line 84: Probably should add an `Examples` section here also. I saw that ca-serial string array was thinking "is that --ca-serial serial1,serial2 or --ca-serial serial1 --ca-serial serial2"

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,224 @@
...
+		Args:  cobra.ExactArgs(1),
+		Long: `Disable device CA devices with client certificates issued by it can no longer be registered.
+
+When the online or local device CA is disabled:
+- It is no longer possible to register new devices with client certificates issued by this device CA.
+- Existing devices with client certificates issued by this device CA can continue to connect and use FoundriesFactory.
+
+Usually, when the device CA is compromised, a user would:
+1. Immediately disable a given device CA using "fioctl keys ca disable-device-ca <PKI Directory> --serial <CA Serial>".
+2. Inspect their devices with client certificates issued by that device CA, and remove any compromised devices.
```
camilamacedo86, line 60: Can we add here the command to remove the device as well?
I loved the steps that make the feature much more clear.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,224 @@
...
+3. Create a new device CA using "fioctl keys ca add-device-ca <PKI Directory> --online-ca|--local-ca".
+4. Rotate a client certificate of all legitimate devices from step 2 to the certificate issued by a new device CA.
```
camilamacedo86, line 62: Can we add the command as well?

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: I find `several device CAs` more appropriate here than `multiple device CAs`, pointing at a normally small (limited) number of device CAs per factory.
However, it makes sense to use `_multiple_ times to revoke _several_ device CAs` to avoid tautology.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: I will try to add one; maybe also extract a part of the description into examples.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: Then I'd have to also add a command to list devices and a command to inspect individual device.
Though I think it makes sense to add smth like `(see "fioctl devices")` at the end.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: IIUC this is more a process than a command. Moreover, there are many ways how the customer might implement their workflow for this case.
But I guess it makes sense to add smth like `(see e.g. "fioctl devices config rotate-certs")`.

x509/golang.go
```diff
@@ -136,6 +136,42 @@ func SignEstCsr(csrPem string) string {
...
+	crl := &x509.RevocationList{
+		Number:     big.NewInt(1),
+		ThisUpdate: now,
+		NextUpdate: now.Add(time.Minute * 15),
+	}
+	for serial, reason := range serials {
+		num := new(big.Int)
+		if _, ok := num.SetString(serial, 10); !ok {
+			// We expect a valid input here
+			panic("Value is not a valid base 10 serial:" + serial)
```
mike-sul, line 154: Maybe `DieNotNil(error.New())`?

vkhoroz: We do validation at the command level.
If we get an invalid serial here - it is a programming error; so I'd like the user to report the full traceback.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -113,6 +114,31 @@ func doRevokeDeviceCa(cmd *cobra.Command, args []string) {
...
 	}
+
+	caList, err := api.FactoryGetCA(factory)
+	subcommands.DieNotNil(err)
+	validSerials := make(map[string]bool, 0)
+	for len(caList.CaCrt) > 0 {
+		block, remaining := pem.Decode([]byte(caList.CaCrt))
+		if block == nil {
+			// could be excessive whitespace
+			caList.CaCrt = strings.TrimSpace(string(remaining))
```
mike-sul, line 128: It looks like an infinite loop may occur if the `remaining` is some garbage so `block` is `nil` while `remaining` has no spaces just some garbage.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -113,6 +114,31 @@ func doRevokeDeviceCa(cmd *cobra.Command, args []string) {
...
+		}
+		caList.CaCrt = string(remaining)
+		c, err := x509Lib.ParseCertificate(block.Bytes)
+		if err != nil {
+			fmt.Println("WARNING: Failed to parse certificate from server:" + err.Error())
+			continue
+		}
+		validSerials[c.SerialNumber.Text(10)] = true
+	}
+	for serial := range toRevoke {
```
mike-sul, line 139: Maybe it makes sense to "die" if `validSerials` is empty?

subcommands/keys/ca_show.go
```diff
@@ -74,22 +77,18 @@ func doShowCA(cmd *cobra.Command, args []string) {
...
+	printOneCert(resp.RootCrt)
 	fmt.Println("## Server TLS Certificate")
-	if prettyFormat {
-		prettyPrint(resp.TlsCrt)
-	} else {
-		fmt.Println(resp.TlsCrt)
-	}
+	printOneCert(resp.TlsCrt)
 	fmt.Println("\n## Device Authentication Certificate(s)")
+	printOneCert(resp.CaCrt)
```
mike-sul, line 85: Is this going to print all device CAs or just one?

x509/golang.go
mike-sul: I thought a user can specify `serials` via the command line. I guess the user specified serial is already verified when this code is executed.

mike-sul (approved):
LGTM

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: Hmm... I think you are right: _If no PEM data is found, p is nil and the whole of the input is returned in rest._
Which means that the original code in `show.go` where I copied it from has the same problem;.

So, I need to fix both paces now. Good catch!

subcommands/keys/ca_revoke_device_ca.go
camilamacedo86: `fioctl devices delete`

subcommands/keys/ca_revoke_device_ca.go
camilamacedo86: IMHO Yep, just give a direction I think the e.g. is very helpful so that the person can better understand the flow and look further. 

I would be happy with both ideas.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,224 @@
...
+		"Can be used with dry-run to show the generated CRL in a pretty format.")
+	cmd.Flags().StringArrayP("ca-file", "", nil,
+		"A file name of the device CA to "+op+". Can be used several times to "+op+" several device CAs")
+	cmd.Flags().StringArrayP("ca-serial", "", nil,
+		"A serial number (base 10) of the device CA to "+op+". Can be used several times to "+op+" several device CAs")
+	_ = cmd.MarkFlagFilename("ca-file")
+	// HSM variables defined in ca_create.go
+	cmd.Flags().StringVarP(&hsmModule, "hsm-module", "", "", "Load a root CA key from a PKCS#11 compatible HSM using this module")
+	cmd.Flags().StringVarP(&hsmPin, "hsm-pin", "", "", "The PKCS#11 PIN to log into the HSM")
+	cmd.Flags().StringVarP(&hsmTokenLabel, "hsm-token-label", "", "", "The label of the HSM token containing the root CA key")
```
camilamacedo86, line 83: Does fioctl also create those? Am I right? Can we add the file name we created for those cases?

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,224 @@
...
+		block, remaining := pem.Decode([]byte(caList.CaCrt))
+		if block == nil {
+			// could be excessive whitespace
+			caList.CaCrt = strings.TrimSpace(string(remaining))
+			continue
+		}
+		caList.CaCrt = string(remaining)
+		c, err := x509Lib.ParseCertificate(block.Bytes)
+		if err != nil {
+			fmt.Println("WARNING: Failed to parse certificate from server:" + err.Error())
```
camilamacedo86, line 134: Can we add the file name that we could not parse?

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,224 @@
...
+				} else {
+					fmt.Println("\t\t\t\tUnknown OID", ext.Id.String())
+				}
+			}
+		}
+	}
+	if len(c.Extensions) > 0 {
+		fmt.Println("\tExtensions:")
+		for _, ext := range c.Extensions {
+			if ext.Id.String() == "2.5.29.35" {
```
camilamacedo86, line 207: Why have we a fixed id?
Could we add it to a const and then add a comment on top to clarify?

x509/golang.go
vkhoroz: Yes, exactly.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,224 @@
...
+				fmt.Print("\t\tx509v3 Authority Key Id: ")
+				if ext.Critical {
+					fmt.Print("(critical)")
+				}
+				fmt.Println("\n\t\t\t", hex.EncodeToString(c.AuthorityKeyId))
+			} else if ext.Id.String() == "2.5.29.20" {
```
camilamacedo86, line 213: Same of https://github.com/foundriesio/fioctl/pull/324/files#r1388175053 here.

x509/common.go
```diff
@@ -65,3 +86,18 @@ func ValidateHsmArgs(hsmModule, hsmPin, hsmTokenLabel, moduleArg, pinArg, tokenA
...
+
+func parseOnePemBlock(pemBlock string) *pem.Block {
+	first, rest := pem.Decode([]byte(pemBlock))
+	if first == nil || len(rest) > 0 {
+		subcommands.DieNotNil(errors.New("Malformed PEM data"))
+	}
+	return first
+}
+
+func LoadCertFromFile(fn string) *x509.Certificate {
```
camilamacedo86, line 98: Since this method is exported, could we add a comment on top to document it?

subcommands/keys/ca_show.go
vkhoroz: It prints one cert file; which may contain several certificates.

I know the name might look misleading after that my statement.
Maybe, `printCerts` sounds better?

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: Fioctl is a user here. There are no file names on the HSM.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: I followed the same pattern as `ca_show` does.
These are extension OIDs we recognize (because certs we generate have them).
Their meaning must be evident from the code; otherwise we have a problem.
The constant at the top would disperse the (cryptic enough) code, make it less clear.

x509/common.go
vkhoroz: I don't know anybody who is using Fioctl as a library.
So we have a privilege to avoid some excessive formalities.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: The use needs to have a reference how to "inspect" and how to "delete"; `fioctl devices` covers both necessities.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,224 @@
...
+		Short: "Disable device CA devices with client certificates issued by it can no longer be registered",
+		Run:   doRevokeDeviceCa,
+		Args:  cobra.ExactArgs(1),
+		Long: `Disable device CA devices with client certificates issued by it can no longer be registered.
+
+When the online or local device CA is disabled:
+- It is no longer possible to register new devices with client certificates issued by this device CA.
+- Existing devices with client certificates issued by this device CA can continue to connect and use FoundriesFactory.
+
+Usually, when the device CA is compromised, a user would:
```
camilamacedo86, line 58: Can we add this one in the revoke-device-ca helper as well?

subcommands/keys/ca_revoke_device_ca.go
camilamacedo86: I see we just received this input and we never generate those files.
If so fair enough.

vkhoroz: Yes, exactly.

IIRC I added to the README.md how to integrate with the SoftHSM some time ago.
So, it is kind of _loosly_ self-documented; also there should be something about it in docs.
But, we're not going to copy-paste that in every single command; that'd be onerous to maintain across the product.

subcommands/keys/ca_revoke_device_ca.go
camilamacedo86: Oh. I see those values are the specific OID.
They are standard definitions. 
Thank you. Yep makes sense.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: Just another analogy: `fioctl` was inspired by `kubectl` to some extent.
When you look at the `kubectl` help text - it is usually short to nothing.
An idea is that:
- The user running some command is expected to know what they are doing _before_ they hit enter.
- The user how does not yet know what they are doing _should_ read the documentation first.

A problem with help messages is that _they cannot be long enough_ to fully understand the topic, but when they are _too long_ they can create a sense of _false confidence_ for the user. It works like _"hey, I read it and I understand it... means I can now use it"_. As such, overly descriptive help messages provoke _trying that dangerous command_ before reading the documentation. That is definitely not what we want.

x509/bash.go
vkhoroz: It does not help. Golang static analyzer needs a branch.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,83 @@
...
+		subcommands.DieNotNil(errors.New("At least one of --ca-file or --ca-serial must be provided"))
+	}
+
+	subcommands.DieNotNil(os.Chdir(args[0]))
+	hsm, err := x509.ValidateHsmArgs(
+		hsmModule, hsmPin, hsmTokenLabel, "--hsm-module", "--hsm-pin", "--hsm-token-label")
+	subcommands.DieNotNil(err)
+	x509.InitHsm(hsm)
+
+	fmt.Println("Generating Cartificate Revokation List")
```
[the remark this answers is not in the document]
vkhoroz: https://github.com/foundriesio/fioctl/pull/324/commits/5be78535bbd04285bc4d241a8401718bf4decfa7

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/324/commits/ee087eaf9f830d0f5475335f818c472e24338e47

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/324/commits/ee087eaf9f830d0f5475335f818c472e24338e47

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/324/commits/f503830f48c20eebea9541bc1e07c78afc2f91e3

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/324/commits/3cb1d444187b04a2464ea53b747b8fb651a26979 and https://github.com/foundriesio/fioctl/pull/324/commits/0467b2562c79ada90ea27f704eed5401cc5b3003

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/324/commits/8eb792f387f318827e7bb98d90b1f0a31355addc

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: https://github.com/foundriesio/fioctl/pull/324/commits/8eb792f387f318827e7bb98d90b1f0a31355addc

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: A fix https://github.com/foundriesio/fioctl/pull/324/commits/94ca3f22dfc8a98792a928bac9efe8bae23d2098, a cleanup https://github.com/foundriesio/fioctl/pull/324/commits/065dbb7f5129cc05fd2d3de389fe24ca0fd2289e, and a fixup (to be squashed) https://github.com/foundriesio/fioctl/pull/324/commits/762be5a68614b6186e001fd595dced9c13a7e285.

vkhoroz:
@kprosise @vanmaegima @angolini do you want to give a grammar check to the new commands descriptions?

I believe that by Tuesday @doanac and/or @mike-sul will approve this; so it is stable enough for the literacy polish.

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,235 @@
...
+	addRevokeCmdFlags(revokeCmd, "revoke")
+
+	disableCmd := &cobra.Command{
+		Use:   "disable-device-ca <PKI Directory>",
+		Short: "Disable device CA devices with client certificates issued by it can no longer be registered",
+		Run:   doRevokeDeviceCa,
+		Args:  cobra.ExactArgs(1),
+		Long: `Disable device CA devices with client certificates issued by it can no longer be registered.
+
+When the online or local device CA is disabled:
```
vanmaegima, line 69: disabled or revoked?

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	caCmd.AddCommand(revokeCmd)
+	addRevokeCmdFlags(revokeCmd, "revoke")
+
+	disableCmd := &cobra.Command{
+		Use:   "disable-device-ca <PKI Directory>",
+		Short: "Disable device CA devices with client certificates issued by it can no longer be registered",
+		Run:   doRevokeDeviceCa,
+		Args:  cobra.ExactArgs(1),
+		Long: `Disable device CA devices with client certificates issued by it can no longer be registered.
```
vanmaegima, line 67: I'm a bit confused by this, maybe this would be correct? Not sure if even the suggestion is correct
```suggestion
		Long: `Disable device CA devices which client certificates issued by it can no longer be registered.
```

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,235 @@
...
+	revokeCmd := &cobra.Command{
+		Use:   "revoke-device-ca <PKI Directory>",
+		Short: "Revoke device CA from the list of CAs allowed to issue device client certificates",
+		Run:   doRevokeDeviceCa,
+		Args:  cobra.ExactArgs(1),
+		Long: `Revoke device CA from the list of CAs allowed to issue device client certificates.
+
+When the online or local device CA is revoked:
+- It is no longer possible to register new devices with client certificates issued by this device CA.
+- Existing devices with client certificates issued by this device CA can no longer connect to your FoundriesFactory.
```
kprosise, line 37:
```suggestion
- Existing devices with client certificates it issued can no longer connect to your Factory.
```

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,235 @@
...
+func init() {
+	revokeCmd := &cobra.Command{
+		Use:   "revoke-device-ca <PKI Directory>",
+		Short: "Revoke device CA from the list of CAs allowed to issue device client certificates",
+		Run:   doRevokeDeviceCa,
+		Args:  cobra.ExactArgs(1),
+		Long: `Revoke device CA from the list of CAs allowed to issue device client certificates.
+
+When the online or local device CA is revoked:
+- It is no longer possible to register new devices with client certificates issued by this device CA.
```
kprosise, line 36:
```suggestion
- It is no longer possible to register new devices with client certificates it had issued.
```

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,235 @@
...
+- Existing devices with client certificates issued by this device CA can no longer connect to your FoundriesFactory.
+
+You may later re-add a revoked device CA using "keys ca update", if you still keep its certificate stored somewhere.
+Once you do it, devices with client certificates issued by this device CA may connect to your FoundriesFactory again.`,
```
kprosise, line 40:
```suggestion
Once you do this, devices with client certificates issued by this device CA may connect to your Factory again.`,
```

subcommands/keys/ca_revoke_device_ca.go
```diff
@@ -0,0 +1,235 @@
...
+		Short: "Disable device CA devices with client certificates issued by it can no longer be registered",
+		Run:   doRevokeDeviceCa,
+		Args:  cobra.ExactArgs(1),
+		Long: `Disable device CA devices with client certificates issued by it can no longer be registered.
+
+When the online or local device CA is disabled:
+- It is no longer possible to register new devices with client certificates issued by this device CA.
+- Existing devices with client certificates issued by this device CA can continue to connect and use FoundriesFactory.
+
+Usually, when the device CA is compromised, a user would:
```
kprosise, line 73:
```suggestion
Usually, when the device CA is compromised, a user should:
```
Optional suggestion

kprosise (approved):
Some minor suggestions, overall the text LGTM

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: The _disabled_ is a correct word here, as the command name suggests.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: Oopt! This is where I wrote a sentence which makes no sense to me.

How about this (an exact meaning):
```suggestion
Disable device CA, so that new devices with client certificates issued by it can no longer be registered.
```

And then, for the `revoke-device-ca` command, a better description would be this:
```diff
- Revoke device CA from the list of CAs allowed to issue device client certificates.
+ Revoke device CA, so that devices with client certificates issued by it can no longer connect to your factory.
```

vanmaegima: Looks good

vkhoroz:
Rebased, squashed fixups, and applied all wording corrections.

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: Fixed in force-push

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: Fixed in force-push

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: Fixed in force-push

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: Fixed in force-push

subcommands/keys/ca_revoke_device_ca.go
vkhoroz: Fixed in force-push

PR #334, opened 2023-11-15
feat(login): better defaults for custom URLs

StealthyCoder:
Just a simple way of shifting the way the default URL gets set.

If the command line switch is given, then use that URL. 
Otherwise if a config URL is already set, use that URL.
Otherwise fallback to the default we were using previously.

### DIFF
--- a/subcommands/login/cmd.go
+++ b/subcommands/login/cmd.go
@@ -45,7 +45,12 @@ func doLogin(cmd *cobra.Command, args []string) {
 		subcommands.SaveOauthConfig(creds.Config)
 		return
 	}
-
+	if !cmd.Flag("oauth-url").Changed {
+		authURL = subcommands.Config.ClientCredentials.URL
+	}
+	if len(authURL) == 0 {
+		authURL = client.OauthURL
+	}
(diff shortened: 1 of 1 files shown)


subcommands/login/cmd.go
```diff
@@ -28,7 +28,7 @@ func NewCommand() *cobra.Command {
 		Run:   doLogin,
 	}
 	cmd.Flags().BoolVarP(&refreshToken, "refresh-access-token", "", false, "Refresh your current oauth2 access token. This is used when a token's scopes have been updated in app.foundries.io")
-	cmd.Flags().StringVarP(&authURL, "oauth-url", "", client.OauthURL, "OAuth URL to authenticate with")
+	cmd.Flags().StringVarP(&authURL, "oauth-url", "", "", "OAuth URL to authenticate with")
```
vkhoroz, line 31: We need to show some default value to the user.

If the user knows how to override that through the config - that's fine; but I bet that only you or me would ever do that.

There is `Flags.Changed` to verify if some flag was explicitly set by the user.

vkhoroz (approved):

PR #336, opened 2023-11-22
Add max limit for update history

vanmaegima:


subcommands/devices/cmd.go
```diff
@@ -55,7 +55,7 @@ fioctl devices updates <device> $(fioctl devices updates <device> -n1 | tail -n1
 func NewCommand() *cobra.Command {
 	subcommands.RequireFactory(cmd)
 
-	updatesCmd.Flags().IntVarP(&listLimit, "limit", "n", 0, "Limit the number of updates displayed.")
+	updatesCmd.Flags().IntVarP(&listLimit, "limit", "n", 0, "Limit the number of updates displayed (max 10)")
```
vkhoroz, line 58: This is the backend limit which is set by the API.
Thus, if we ever change the API limit to a different value - this help message would immediately become misleading.

I could not found where we document this limit is docs.foundries.io; yet and the Swagger doc for it at https://api.foundries.io/ota/ is also misleading:
```
limitinteger(query) | Results per page
Available values : 10, 20, 30, 40, 50, 100, 200, 500, 1000
Default value : 1000
```

So, I'd rather fix the API doc and add this answer to the docs FAQ.

vkhoroz: I'll open a bug for our API's pagination docs in Swagger: this is not the only problem with them.

vkhoroz:
Closing as per agreement.

PR #340, opened 2023-12-08
fixup! feat(config): Show CreatedBy user name (#246)

doanac:
This was broken in v34

### DIFF
--- a/subcommands/common_config.go
+++ b/subcommands/common_config.go
@@ -78,14 +78,6 @@ type LogConfigsOptions struct {
 	ListContFunc  func(string) (*client.DeviceConfigList, error)
 }
 
-func (opts *LogConfigsOptions) GetUsersLookupById(api *client.Api, factory string) {
-	list, err := api.UsersList(factory)
-	DieNotNil(err)
-	for _, user := range list {
-		opts.UserLookup[user.PolisId] = user
-	}
-}
-
 func LogConfigs(opts *LogConfigsOptions) {
 	var dcl *client.DeviceConfigList
 	listLimit := opts.Limit
--- a/subcommands/config/log.go
+++ b/subcommands/config/log.go
@@ -25,6 +25,9 @@ func doConfigLog(cmd *cobra.Command, args []string) {
 	listLimit, _ := cmd.Flags().GetInt("limit")
 	group, _ := cmd.Flags().GetString("group")
 
+	lookups, err := api.UsersGetLookups(factory)
+	subcommands.DieNotNil(err)
+
 	if group == "" {
 		logrus.Debugf("Showing config history for %s", factory)
 		subcommands.LogConfigs(&subcommands.LogConfigsOptions{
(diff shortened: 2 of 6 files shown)


subcommands/config/log.go
```diff
@@ -25,10 +27,16 @@ func doConfigLog(cmd *cobra.Command, args []string) {
 	listLimit, _ := cmd.Flags().GetInt("limit")
 	group, _ := cmd.Flags().GetString("group")
 
+	lookups, err := subcommands.GetUsersLookupById(api, factory)
+	if err != nil {
+		fmt.Println("ERROR: Unable to look up factory user names", err)
+	}
```
vkhoroz, line 33:
```suggestion
	subcommands.DieNotNil(err, "Failed to look up factory user names")
```

Plus, I think it would work well inside the `GetUsersLookupById`, as it was before.

subcommands/common_config.go
```diff
@@ -78,12 +78,16 @@ type LogConfigsOptions struct {
 	ListContFunc  func(string) (*client.DeviceConfigList, error)
 }
 
-func (opts *LogConfigsOptions) GetUsersLookupById(api *client.Api, factory string) {
+func GetUsersLookupById(api *client.Api, factory string) (map[string]client.FactoryUser, error) {
```
vkhoroz, line 81: Just IMHO this needs to be a helper method of the client itself, it is generic enough and has nothing specific for the config.

So that one could call it like `api.GetUsersMapById(factory)`.
If you go this way, then I think your idea to return an error from it (rather than dying) makes sense.

subcommands/config/log.go
```diff
@@ -37,7 +45,8 @@ func doConfigLog(cmd *cobra.Command, args []string) {
 	} else {
 		logrus.Debugf("Showing config history for %s group %s", factory, group)
 		subcommands.LogConfigs(&subcommands.LogConfigsOptions{
-			Limit: listLimit,
+			UserLookup: lookups,
+			Limit:      listLimit,
```
vkhoroz, line 49: A nit: can you move the `UserLookup` to the end of the struct args (here and above)?

subcommands/config/log.go
doanac: The struct didn't have a member for the api client. I wasn't sure a handle to that in an "Opts" struct made sense. I'm indifferent though. We could do a "lookup table callback func" but I thought that was getting out of hand.

doanac:
@vkhoroz - i think i've address all your comments in the force-push

subcommands/config/log.go
vkhoroz: Only this one, but you can ignore it.

vkhoroz (approved):
LGTM

subcommands/config/log.go
doanac: it was a failed force-push. the new one has what you wanted

StealthyCoder (approved):
Took me one moment to go back to my own PR and see what I missed too. I never hooked up the UserLookup indeed. I wonder why it worked for me locally. Thank you for fixing this.

PR #345, opened 2023-12-14
Support docker-credential-helper using absolute path

vkhoroz:


cmd/root.go
```diff
@@ -47,7 +47,7 @@ var rootCmd = &cobra.Command{
 }
 
 func Execute() {
-	if os.Args[0] == docker.DOCKER_CREDS_HELPER {
+	if strings.Contains(os.Args[0], docker.DOCKER_CREDS_HELPER) {
```
doanac, line 50: should it be string.Contains or string.HasSuffix?

mike-sul (approved):
It works for me, not sure how though.

cmd/root.go
vkhoroz: I copied what we do for `git-credential-helper` below. I believe it has something to do with the `.exe` suffix on Windows.

Issue #344, opened 2023-12-13
`keys ca add-device-ca` fails with relative paths to new keypair


doanac:
I tried creating a CA for my EST server with:
```
fioctl keys ca add-device-ca /var/code/andy-corp/keys/gateway --local-ca --local-ca-filename ../keys/ca.pem
```
However we do an `os.Chdir` in code that broke things in a confusing way:

https://github.com/foundriesio/fioctl/blob/fc8e54f0a40844ce52c1a38b4a2bc175e06a6d74/subcommands/keys/ca_add_device_ca.go#L63

There are options:
* don't allow slashes in paths (ie. force the keypair to be created in the PKI dir)
* don't allow relative paths
* convert relative paths to absolute before calling `os.Chdir`

Not sure what the original intent of this was, so I'll leave it to @vkhoroz to decide.

vkhoroz:
Histrorically, all operations we do with the PKI are related to the PKI directory.
There were a couple of deviations from this rule here and there, but after I consolidated the PKI code they all behave the same way.

In this specific command, `--local-ca-filename` preferrably is a file name, not a file path.
However, I understand that nothing currently stops the user from specifying a path.
Just IMHO the easiest thing to do is to update the help message to tell that this file is created inside the PKI dir.

StealthyCoder:
Would be nice if Golang had something like `pushd` and `popd` in the library.

PR #349, opened 2023-12-18
keys: Don't allow slash in device-ca

doanac:
This command expects a filename and not a path. This change makes the error more clear.

fixes issue #344

### DIFF
--- a/subcommands/keys/ca_add_device_ca.go
+++ b/subcommands/keys/ca_add_device_ca.go
@@ -50,6 +50,14 @@ All such CAs will be added to the list of device CAs trusted by the device gatew
 	cmd.Flags().StringVarP(&hsmTokenLabel, "hsm-token-label", "", "", "The label of the HSM token containing the root CA key")
 }
 
+func assertFileName(flagName, value string) {
+	if len(value) > 0 {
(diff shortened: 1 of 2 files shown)


subcommands/keys/ca_add_device_ca.go
```diff
@@ -60,6 +60,12 @@ func doAddDeviceCa(cmd *cobra.Command, args []string) {
 		subcommands.DieNotNil(errors.New("At least one of --online-ca or --local-ca must be true"))
 	}
 
+	if len(localCaFilename) > 0 {
+		if strings.ContainsRune(localCaFilename, os.PathSeparator) {
+			subcommands.DieNotNil(errors.New("The --local-ca-filename argument must be filename and not a path"))
```
vkhoroz, line 65: I did not want to enforce this limitation, but I understand your justification.

Not sure if we have similar parameters in other ca subcommands.
Can you check?

vkhoroz (approved):
LGTM

subcommands/keys/ca_add_device_ca.go
doanac: found another spot. address in force push with a helper function.

vkhoroz (approved):
LGTM

PR #351, opened 2023-12-19
Hotfix: clean up key owners when rotating TUF keys

vkhoroz:
This is an important hot fix.
Currently, the API fails on subsequent key rotations due to denying old key owners.

Fortunately, customers should not see it, as the first rotation succeeds. But, when the user wants to rotate the key which was rotated using new Fioctl - that second rotation fails.

### DIFF
--- a/subcommands/keys/tuf_utils.go
+++ b/subcommands/keys/tuf_utils.go
@@ -354,6 +354,7 @@ func removeUnusedTufKeys(root *client.AtsTufRoot) {
 		if !found {
 			fmt.Println("= Removing unused key:", k)
 			delete(root.Signed.Keys, k)
+			delete(root.Signed.KeyOwners, k)
 		}
 	}
 }


subcommands/keys/tuf_utils.go
```diff
@@ -354,6 +354,7 @@ func removeUnusedTufKeys(root *client.AtsTufRoot) {
 		if !found {
 			fmt.Println("= Removing unused key:", k)
 			delete(root.Signed.Keys, k)
+			delete(root.Signed.KeyOwners, k)
```
mike-sul, line 357: iiuc, the key ID == owner ID (`root.Signed.KeyOwners[key.signer.Id]`)?

mike-sul:
Nice catch.

subcommands/keys/tuf_utils.go
vkhoroz: It is the key-id to owner-id map.
Owner ID is a user ID (of a user who created a key).

mike-sul: I see that `key.signer.Id` is used as a key in the `root.Signed.KeyOwners` map. It makes me think that a signer ID is equal to a key ID.

mike-sul: https://github.com/foundriesio/fioctl/blob/eb5d7c91d08e3c9e8c7be4eb042c85e1b906aff3/subcommands/keys/tuf_utils.go#L333

vkhoroz: Signer here is a Golang's crypto-signer interface, not a person who owns a signing key.

PR #352, opened 2023-12-20
Feature: re-sign all active waves on TUF targets key rotation

vkhoroz:
This is a set of improvements using the new APIs:

- Ability to show only active waves using `waves list --only-active`.
- Ability to auto-re-sign active waves targets when running `keys tuf updates rotate-offline-key -r targets`.
- Ability to manually re-sign selected waves using `keys tuf updates sign-prod-targets --waves=<wave-list>`.
- Detect when user tries to re-sign non-existing prod targets tag using `keys tuf updates sign-prod-targets --tags=<tag-list>`.
- Verify that all `page, limit` arguments are non-negative numbers in `devices list, devices list-denied, waves list`.

### DIFF
--- a/subcommands/keys/tuf_updates_rotate_offline_key.go
+++ b/subcommands/keys/tuf_updates_rotate_offline_key.go
@@ -110,7 +110,7 @@ func doTufUpdatesRotateOfflineRootKey(cmd *cobra.Command) {
 
 	fmt.Println("= Uploading new TUF root")
 	tmpFile := saveTempTufCreds(keysFile, newCreds)
-	err = api.TufRootUpdatesPut(factory, txid, newCiRoot, newProdRoot, nil)
+	err = api.TufRootUpdatesPut(factory, txid, newCiRoot, newProdRoot, nil, nil)
 	handleTufRootUpdatesUpload(tmpFile, keysFile, err)
 }
 
@@ -207,7 +207,7 @@ outerLoop:
 
 	fmt.Println("= Uploading new TUF root")
 	tmpFile := saveTempTufCreds(targetsKeysFile, newCreds)
-	err = api.TufRootUpdatesPut(factory, txid, newCiRoot, newProdRoot, newTargetsSigs)
+	err = api.TufRootUpdatesPut(factory, txid, newCiRoot, newProdRoot, newTargetsSigs, nil)
 	handleTufRootUpdatesUpload(tmpFile, targetsKeysFile, err)
 }
 
--- a/subcommands/keys/tuf_updates_rotate_offline_key.go
+++ b/subcommands/keys/tuf_updates_rotate_offline_key.go
@@ -186,19 +186,16 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
 		subcommands.DieNotNil(err)
 	}
 
-	targetsMap, err := api.ProdTargetsList(factory, false)
+	targetsProdMap, err := api.ProdTargetsList(factory, false)
 	subcommands.DieNotNil(err, "Failed to fetch production targets:")
-outerLoop:
-	for tag, targets := range targetsMap {
-		for _, sig := range targets.Signatures {
-			if sig.KeyID == oldestKey.Id {
-				continue outerLoop
-			}
-		}
-		// These targets does not contain a signature by a rotated key - skip them
-		delete(targetsMap, tag)
-	}
-	newTargetsSigs, err := signProdTargets(newKey, targetsMap)
+	excludeTargetsWithoutKeySigInplace(targetsProdMap, oldestKey.Id)
+	newTargetsProdSigs, err := signProdTargets(newKey, targetsProdMap)
+	subcommands.DieNotNil(err)
+
+	targetsWaveMap, err := api.WaveTargetsList(factory, false)
+	subcommands.DieNotNil(err, "Failed to fetch production wave targets:")
+	excludeTargetsWithoutKeySigInplace(targetsWaveMap, oldestKey.Id)
+	newTargetsWaveSigs, err := signProdTargets(newKey, targetsWaveMap)
 	subcommands.DieNotNil(err)
 
 	if shouldSign {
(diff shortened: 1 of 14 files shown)


subcommands/keys/tuf_updates_rotate_offline_key.go
```diff
@@ -185,16 +185,20 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
...
-			for _, sig := range targets.Signatures {
-				if sig.KeyID == oldestKey.Id {
-					return false
-				}
+
+	targetsMap, err := api.ProdTargetsList(factory, false)
+	subcommands.DieNotNil(err, "Failed to fetch production targets:")
+outerLoop:
+	for tag, targets := range targetsMap {
+		for _, sig := range targets.Signatures {
```
mike-sul: Just IMO, having the internal loop in some function (local scope) would make it a bit cleaner.

subcommands/keys/tuf_updates_rotate_offline_key.go
```diff
@@ -185,16 +185,20 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
 			subcommands.SliceRemove(curCiRoot.Signed.Roles["targets"].KeyIDs, onlineTargetsId))
 		subcommands.DieNotNil(err)
 	}
-	newTargetsSigs, err := signProdTargets(factory, newKey,
-		func(tag string, targets client.AtsTufTargets) bool {
```
mike-sul, line 189: `tag` wasn't used in the previous version at all.

subcommands/keys/tuf_updates_rotate_offline_key.go
```diff
@@ -186,19 +186,16 @@ func doTufUpdatesRotateOfflineTargetsKey(cmd *cobra.Command) {
...
-		delete(targetsMap, tag)
-	}
-	newTargetsSigs, err := signProdTargets(newKey, targetsMap)
+	excludeTargetsWithoutKeySigInplace(targetsProdMap, oldestKey.Id)
+	newTargetsProdSigs, err := signProdTargets(newKey, targetsProdMap)
+	subcommands.DieNotNil(err)
+
+	targetsWaveMap, err := api.WaveTargetsList(factory, false)
+	subcommands.DieNotNil(err, "Failed to fetch production wave targets:")
+	excludeTargetsWithoutKeySigInplace(targetsWaveMap, oldestKey.Id)
```
mike-sul, line 197: I am just wondering if it's possible that some old wave's targets wasn't re-signed after targets key rotation (because we didn't support it before). So, if we filter in by ` oldestKey.Id` it means that such `targetsWave` won't be re-signed.

mike-sul: but devices will reject such wave targets anyway.

subcommands/keys/tuf_updates_rotate_offline_key.go
vkhoroz: A `tag` was used indirectly in this code block inside `signProdTargets` (see below):
```
for tag, targets := range targetsMap {
		if excludeFunc != nil && excludeFunc(tag, targets) {
			continue
		}
```

That logic was rather convoluted and really hard to follow.
So, that code is now gone, and that loop was replicated here.

It is kind of those problems when a _complicated code just works_ until a more complex problem arrives which desperately calls for simplification.

subcommands/keys/tuf_updates_rotate_offline_key.go
vkhoroz: I extracted this thing into its own function in one of the last commits: https://github.com/foundriesio/fioctl/pull/352/commits/d1d10684cbf392b9f2bf2758ca6117e63b0218ac.

It is not exactly what you asked for; but it is rather simple now.

subcommands/keys/tuf_updates_sign_prod_targets.go
```diff
@@ -65,12 +71,25 @@ For example, add a new offline TUF targets key, before signing production target
...
+	// If only wave names or only tags specified - re-sign only what was specified (either wave names or tags).
+	if len(tags) > 0 || len(waveNames) == 0 {
+		targetsProdMap, err := api.ProdTargetsList(factory, true, tags...)
+		subcommands.DieNotNil(err, "Failed to fetch production targets:")
+		newTargetsProdSigs, err = signProdTargets(signer, targetsProdMap)
+		subcommands.DieNotNil(err)
+	}
+	if len(waveNames) > 0 || len(tags) == 0 {
+		targetsWaveMap, err := api.WaveTargetsList(factory, true, waveNames...)
+		subcommands.DieNotNil(err, "Failed to fetch production wave targets:")
```
mike-sul, line 87: just a nit, not sure `production` is needed in the error message.

mike-sul (approved):

subcommands/keys/tuf_updates_sign_prod_targets.go
mike-sul: what will happen if both `tags` and `waves` are not specified?

vkhoroz: I found it is useful.

Although `wave targets` implicitly presumes `production` attribute; I indeed found it useful to use both words in this context.

vkhoroz: > what will happen if both tags and waves are not specified?

The default action will happen, which is to sign all production targets and all active waves, using the specified targets key.

vkhoroz: I suspect the majority of users will only ever run the default command.

mike-sul: what will happen if both tags and waves are not specified?

mike-sul:
> * Ability to show only active waves using `waves list --only-active`.

Maybe it's better to call it just `--active`?

subcommands/keys/tuf_updates_rotate_offline_key.go
mike-sul: well, I don't like that idiom. Just don't like blindly following "rules"/idioms" :).
But, it's ok in this case, just imho.

vkhoroz: I wanted to break an _established rule to never use goto_... so you tell me this. OMG LOL.

StealthyCoder (approved):
Looks all very reasonable to me :+1:

PR #354, opened 2023-12-22
Support filtering waves by status and tag

vkhoroz:


subcommands/waves/list.go
```diff
@@ -18,17 +18,17 @@ func init() {
 	cmd.AddCommand(listCmd)
 	listCmd.Flags().Uint64P("limit", "n", 20, "Limit the number of results displayed.")
 	listCmd.Flags().Uint64P("page", "p", 1, "Page of waves to display when pagination is needed")
-	listCmd.Flags().BoolP("only-active", "", false, "Only show currently active waves")
+	listCmd.Flags().StringP("status", "S", "", "Only show waves with a given status; one of (active, complete, canceled)")
```
doanac, line 21: I noticed we aren't super graceful when handling a bad status flag:
```
$ fioctl waves list --status foo
ERROR: HTTP error during GET 'https://api.foundries.io/ota/factories/andy-corp/waves/?limit=20&page=1&status=foo': 400 BAD REQUEST
= Input payload validation failed
 * status: Filter waves by status The value 'foo' is not a valid choice for 'status'.
```

I seem to remember Cobra doesn't really do enums so this might be about the best we can do w/o adding a bunch of code.

PR #362, opened 2024-01-23
Chore: more verbose error messages for missing TUF private keys

vkhoroz:


doanac (approved):
thanks. this is helpful

vkhoroz:
> Why the command is trying to read the "old" TUF targets private key?

IIRC we've discussed this today:

> When the user runs rotate-offline-key they need to provide evidence that they own the key,
That evidence comes in a form of a tar.gz file containing a that private key.
Otherwise, a user could accidentally rotate someone else's key.

> If the user have lost their key - there is still an ability to add a new key, or remove a lost key, or do both...
But only in a more explicit form, using add-offline-key/delete-offline-key commands.

mike-sul:
> IIRC we've discussed this today

Yes, I know why the command "reading" the current targets role key. However, the message is still misleading for me - hence my question. The key reason is that it doesn't convey what it is trying to achieve, instead it says about mechanics to achieve it.

IMO, The message should say something like "Failed to find the current targets role key to be rotated (or for rotation) in <path>; expected to find keys with IDs listed in TUF root metadata: [] ".

Maybe I am just looking into it from a wrong perspective. Perhaps checking it with the customer support would be the right way to verify whether the updated error messages are clear.

mike-sul (approved):

vkhoroz:
@mike-sul your proposed message
```
Failed to find the current targets role key to be rotated in 'file'; expected to find keys with IDs listed in TUF root metadata: [id1 id2].
```
... is almost the same as a message added by this PR:
```
Error reading old TUF targets private key:
 Found no active private key for key IDs: [id1 id2].
```

The only 3 essential differences I found are:
1. Using `current` instead of `old`. I'm fine to make this replacement if you think it is more clear.
2. You message shows a file name. Adding this info would require much more changes, but also look uglier. IMHO a user should know what file they supplied from a command they run.
3. Your message adds `listed in TUF root metadata` to describe key IDs. IMHO this is too verbose, but also not fully exact.

I think all the user needs to know is that on of keys 'id1 id2' for role 'root or targets' is missing in a file they provided.
The messages in this PR convey that information with minimal code changes.

mike-sul:
> .. is almost the same as a message added by this PR:

As for me they are absolutely different :).

1. Yes, "current" is more clear.
2. How customer can know that it is looking for a key in some file and why? The key point is to say that the command is trying to find out which of the keys listed in the current root to rotate. At least something like in "the specified key file" would be useful.


> I think all the user needs to know is that one of the old or new private keys with IDs id1 ... idN for role root or targets is missing in a file they provided.

That's great, I would put in the error message "... targets is missing in a file they provided".

3.

vkhoroz:
@mike-sul please, take a look at updated messages:
```
$ fioctl keys tuf updates rotate-offline-key -r root -k bin/bad-tuf-root.tgz -s
ERROR: Error reading current TUF Root private key from a specified file:
 Found no private key for key IDs: [abe7f32a17acc760d06bd146e03a9aa8a8c48ba813594e133263e0e19ccbdb7f].
```

Does it work well for you?

MrCry0 (approved):

PR #367, opened 2024-02-05
offline-update: Add wave target pull support

mike-sul:
- Add ability to download wave's TUF metadata and its target content to the offline update bundle.

- Correct checking whether the production target exists for a given tag

### DIFF
--- a/subcommands/targets/offline-update.go
+++ b/subcommands/targets/offline-update.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"time"
 
+	tuf "github.com/theupdateframework/notary/tuf/data"
+
 	"github.com/foundriesio/fioctl/client"
 	"github.com/foundriesio/fioctl/subcommands"
 	"github.com/spf13/cobra"
@@ -35,15 +37,19 @@ var (
 	ouTufOnly              bool
 	ouNoApps               bool
 	ouAllowMultipleTargets bool
+	ouWave                 string
 )
 
 func init() {
 	offlineUpdateCmd := &cobra.Command{
-		Use:   "offline-update <target-name> <dst> --tag <tag> [--prod] [--expires-in-days <days>] [--tuf-only]",
+		Use:   "offline-update <target-name> <dst> --tag <tag> [--prod | --wave <wave-name>] [--expires-in-days <days>] [--tuf-only]",
 		Short: "Download Target content for an offline update",
 		Run:   doOfflineUpdate,
 		Args:  cobra.ExactArgs(2),
 		Example: `
+	# Download update content of the wave target #1451 for "intel-corei7-64" hardware type
+	fioctl targets offline-update intel-corei7-64-lmp-1451 /mnt/flash-drive/offline-update-content --wave wave-deployment-001
+
 	# Download update content of the production target #1451 tagged by "release-01" for "intel-corei7-64" hardware type
 	fioctl targets offline-update intel-corei7-64-lmp-1451 /mnt/flash-drive/offline-update-content --tag release-01 --prod
 
@@ -57,6 +63,8 @@ func init() {
 		"Target tag")
 	offlineUpdateCmd.Flags().BoolVarP(&ouProd, "prod", "", false,
 		"Instruct to fetch content of production Target")
+	offlineUpdateCmd.Flags().StringVarP(&ouWave, "wave", "", "",
+		"Instruct to fetch content of wave Target; a wave name should be specified")
 	offlineUpdateCmd.Flags().IntVarP(&ouExpiresIn, "expires-in-days", "e", 30,
 		"Desired metadata validity period in days")
 	offlineUpdateCmd.Flags().BoolVarP(&ouTufOnly, "tuf-only", "m", false,
@@ -65,6 +73,8 @@ func init() {
 		"Skip fetching Target Apps")
 	offlineUpdateCmd.Flags().BoolVarP(&ouAllowMultipleTargets, "allow-multiple-targets", "", false,
 		"Allow multiple targets to be stored in the same <dst> directory")
+	offlineUpdateCmd.MarkFlagsMutuallyExclusive("tag", "wave")
+	offlineUpdateCmd.MarkFlagsMutuallyExclusive("prod", "wave")
 }
 
 func doOfflineUpdate(cmd *cobra.Command, args []string) {
@@ -72,18 +82,34 @@ func doOfflineUpdate(cmd *cobra.Command, args []string) {
 	targetName := args[0]
 	dstDir := args[1]
 
-	if len(ouTag) == 0 {
-		subcommands.DieNotNil(errors.New("missing mandatory flag `--tag`"))
+	if len(ouTag) == 0 && len(ouWave) == 0 {
+		subcommands.DieNotNil(errors.New("Either `--tag` or `--wave` should be specified"))
 	}
 
-	fmt.Printf("Checking whether Target exists; target: %s, tag: %s, production: %v\n", targetName, ouTag, ouProd)
-	subcommands.DieNotNil(checkIfTargetExists(factory, targetName, ouTag, ouProd))
-
-	ti, err := getTargetInfo(factory, targetName)
-	subcommands.DieNotNil(err, "Failed to obtain Target's details:")
+	var targetCustomData *tuf.FileMeta
+	var targetGetErr error
+	// Get the wave/prod/CI specific target with the specified tag to check if it is really present
+	if len(ouWave) > 0 {
+		fmt.Printf("Getting Wave Target details; target: %s, wave: %s...\n", targetName, ouWave)
+		_, targetGetErr = getWaveTargetMeta(factory, targetName, ouWave)
+	} else if ouProd {
+		fmt.Printf("Getting production Target details; target: %s, tag: %s...\n", targetName, ouTag)
+		_, targetGetErr = getProdTargetMeta(factory, targetName, ouTag)
+	} else {
+		fmt.Printf("Getting CI Target details; target: %s, tag: %s...\n", targetName, ouTag)
+		_, targetGetErr = getCiTargetMeta(factory, targetName, ouTag)
+	}
+	subcommands.DieNotNil(targetGetErr)
+	// Get the specified target from the list of factory targets to obtain the "original" tag/branch that produced
+	// the target, so we can find out the correct app bundle fetch URL.
+	targetCustomData, targetGetErr = api.TargetGet(factory, targetName)
+	subcommands.DieNotNil(targetGetErr)
+	// Get the target info in order to deduce the ostree and app download URLs
+	ti, err := getTargetInfo(targetCustomData)
+	subcommands.DieNotNil(err)
 
 	fmt.Printf("Refreshing and downloading TUF metadata for Target %s to %s...\n", targetName, path.Join(dstDir, "tuf"))
-	subcommands.DieNotNil(downloadTufRepo(factory, targetName, ouTag, ouProd, ouExpiresIn, path.Join(dstDir, "tuf")), "Failed to download TUF metadata:")
+	subcommands.DieNotNil(downloadTufRepo(factory, targetName, ouTag, ouProd, ouWave, ouExpiresIn, path.Join(dstDir, "tuf")), "Failed to download TUF metadata:")
 	fmt.Println("Successfully refreshed and downloaded TUF metadata")
 
 	if !ouTufOnly {
@@ -110,37 +136,17 @@ Notice that multiple targets in the same directory is only supported in LmP >= v
 	}
 }
 
-func checkIfTargetExists(factory string, targetName string, tag string, prod bool) error {
-	data, err := api.TufMetadataGet(factory, "targets.json", tag, prod)
-	if err != nil {
-		if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
-			return fmt.Errorf("the specified Target has not been found; target: %s, tag: %s, production: %v", targetName, ouTag, ouProd)
-		}
-		return fmt.Errorf("failed to check whether Target exists: %s", err.Error())
-	}
-	targets := client.AtsTufTargets{}
-	err = json.Unmarshal(*data, &targets)
-	if err != nil {
-		return fmt.Errorf("failed to check whether Target exists: %s", err.Error())
-	}
-	for tn := range targets.Signed.Targets {
-		if tn == targetName {
-			return nil
-		}
-	}
-	return fmt.Errorf("the specified Target has not been found; target: %s, tag: %s, production: %v", targetName, ouTag, ouProd)
-}
-
-func getTargetInfo(factory string, targetName string) (*ouTargetInfo, error) {
+func getTargetInfo(targetFile *tuf.FileMeta) (*ouTargetInfo, error) {
+	// Since the wave/prod/CI specific target json usually doesn't contain the "original" branch/tag that the apps were fetched for
+	// we do the following to determine where to fetch the target app bundle from.
 	// Getting Target's custom info from the `/ota/factories/<factory>/targets/<target-name>` because:
 	// 1. a target name is unique and represents the same Target across all "tagged" targets set including prod;
 	// 2. only this target version/representation contains an original tag(s)/branch that
 	// the `image-assemble` and apps fetching was performed for (needed for determining where to download Apps from).
-	custom, err := getTargetCustomInfo(factory, targetName)
+	custom, err := api.TargetCustom(*targetFile)
 	if err != nil {
 		return nil, err
 	}
-
 	info := ouTargetInfo{}
 	info.version, err = strconv.Atoi(custom.Version)
 	if err != nil {
@@ -163,6 +169,59 @@ func getTargetInfo(factory string, targetName string) (*ouTargetInfo, error) {
 	return &info, nil
 }
 
+func getWaveTargetMeta(factory string, targetName string, wave string) (*tuf.FileMeta, error) {
+	waveTargets, err := api.WaveTargetsList(factory, true, wave)
+	if err != nil {
+		if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
+			return nil, fmt.Errorf("No active Wave with the specified name was found; wave: %s", wave)
+		}
+		return nil, fmt.Errorf("Failed to get Wave Target metadata: %s", err.Error())
+	}
+	if foundTargetMeta, ok := waveTargets[wave].Signed.Targets[targetName]; ok {
+		return &foundTargetMeta, nil
+	} else {
+		return nil, fmt.Errorf("The specified Target is not found among wave targets;"+
+			" target: %s, wave: %s", targetName, wave)
+	}
+}
+
+func getProdTargetMeta(factory string, targetName string, tag string) (*tuf.FileMeta, error) {
+	targets, err := api.ProdTargetsGet(factory, tag, true)
+	if err != nil {
+		if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
+			return nil, fmt.Errorf("No production targets were found for the specified tag `%s`", tag)
+		}
+		return nil, fmt.Errorf("Failed to get production Target metadata: %s", err.Error())
+	}
+	if foundTargetMeta, ok := targets.Signed.Targets[targetName]; ok {
+		return &foundTargetMeta, nil
+	} else {
+		return nil, fmt.Errorf("No production target with the given tag is found;"+
+			" target: %s, tag: %s", targetName, tag)
+	}
+}
+
+func getCiTargetMeta(factory string, targetName string, tag string) (*tuf.FileMeta, error) {
+	data, err := api.TufMetadataGet(factory, "targets.json", tag, false)
+	if err != nil {
+		if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
+			return nil, fmt.Errorf("No CI targets were found for the specified tag `%s`", tag)
+		}
+		return nil, fmt.Errorf("Failed to get CI Target metadata: %s", err.Error())
+	}
+	targets := client.AtsTufTargets{}
+	err = json.Unmarshal(*data, &targets)
+	if err != nil {
+		return nil, err
+	}
+	if foundTargetMeta, ok := targets.Signed.Targets[targetName]; ok {
+		return &foundTargetMeta, nil
+	} else {
+		return nil, fmt.Errorf("No CI target with the given tag is found;"+
+			" target: %s, tag: %s", targetName, tag)
+	}
+}
+
 func isDstDirClean(dstDir string) bool {
 	for _, subDir := range []string{"ostree_repo", "apps"} {
 		fullPath := path.Join(dstDir, subDir)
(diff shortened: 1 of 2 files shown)


mike-sul:
This is the changes to the `fioctl targets offline-update` command that makes use of the corresponding changes to the handler in ota-lite https://github.com/foundriesio/ota-lite/pull/38.

subcommands/targets/offline-update.go
```diff
@@ -65,25 +71,54 @@ func init() {
...
+			fmt.Printf("Getting CI Target details; target: %s, tag: %s...\n", targetName, ouTag)
+			return getCiTargetMeta(factory, targetName, ouTag)
+		}
+	}
+	// Get the wave/prod/CI specific target with the specified tag to check if it is really present, and
+	// get the specified target from the list of factory targets to obtain the "original" tag/branch that produced
+	// the target so we can find out the correct app bundle fetch URL.
+	ti, err := checkAndGetTargetInfo(getTargetMeta, func() (*tuf.FileMeta, error) {
+		return api.TargetGet(factory, targetName)
+	})
```
vkhoroz, line 117: IIUC you simply ignore the return of the `getTargetMeta` inside the `checkAndGetTargetInfo`, only processing the error.

So, just IMHO, the code flow would look more organically (easier to understand) if you do the following:
1. Rename all of `getWaveTargetMeta, getProdTargetMeta, getCiTargetMeta` into `checkXXXExists` returning only an error.
2. Check for the respective targets existance in that large `if..else if...else` clause above.
3. Keep the old vanilla `getTargetInfo` as is and use it here.

All in all, what I suggest is this:
```
	// Your big comment would better look here
	if len(ouWave) > 0 {
		fmt.Printf("Getting Wave Target details; target: %s, wave: %s...\n", targetName, ouWave)
		subcommands.DieNotNil(checkIfWaveTargetExists(factory, targetName, ouWave), "Failed to get a wave target with given parameters")
		// You might preserve your current signature if you wish, it would just make a code a bit longer... but I'm fine with that.
		// _, err := getWaveTargetMeta(factory, targetName, ouWave)
		// subcommands.DieNotNil(err, "Failed to get a wave target with given parameters")
	} else if ouProd {
		fmt.Printf("Getting production Target details; target: %s, tag: %s...\n", targetName, ouTag)
		subcommands.DieNotNil(checkIfProdTargetExists(factory, targetName, ouTag), "Failed to get a production target with given parameters")
	} else {
		fmt.Printf("Getting CI Target details; target: %s, tag: %s...\n", targetName, ouTag)
		subcommands.DieNotNil(checkIfCiTargetExists(factory, targetName, ouTag), "Failed to get a CI target with given parameters")
	}
	ti, err := getTargetInfo(factory, targetName)
```

In addition to the above it looks like in case of CI targets you're trying to fetch CI targets twice, right?
TBH I would optimize that use case. It is OK to duplicate one line of code in 2 places and have a more clear API interaction.

subcommands/targets/offline-update.go
```diff
@@ -65,25 +71,54 @@ func init() {
...
 	dstDir := args[1]
 
-	if len(ouTag) == 0 {
-		subcommands.DieNotNil(errors.New("missing mandatory flag `--tag`"))
+	if len(ouTag) == 0 && len(ouWave) == 0 {
+		subcommands.DieNotNil(errors.New("one of the mandatory parameters `--tag` or `--wave` should be specified"))
+	}
+
+	if len(ouTag) > 0 && len(ouWave) > 0 {
+		subcommands.DieNotNil(errors.New("either `--tag` or `--wave` should be specified exclusively"))
```
doanac, line 88: I think we usually word this like "`--tag` and `--wave` are mutually exclusive".

subcommands/targets/offline-update.go
```diff
@@ -163,6 +187,64 @@ func getTargetInfo(factory string, targetName string) (*ouTargetInfo, error) {
...
+		return nil, fmt.Errorf("the specified Target is not found among wave targets;"+
+			" target: %s, wave: %s", targetName, wave)
+	}
+}
+
+func getProdTargetMeta(factory string, targetName string, tag string) (*tuf.FileMeta, error) {
+	targets, err := api.ProdTargetsGet(factory, tag, true)
+	if err != nil {
+		if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
+			return nil, fmt.Errorf("the specified production Target has not been found; target: %s, tag: %s", targetName, tag)
```
vkhoroz, line 215: This is actually: `no production targets were found for tag %s`.

subcommands/targets/offline-update.go
```diff
@@ -163,6 +187,64 @@ func getTargetInfo(factory string, targetName string) (*ouTargetInfo, error) {
...
+		return nil, fmt.Errorf("no production target with the given tag is found;"+
+			" target: %s, tag: %s", targetName, tag)
+	}
+}
+
+func getCiTargetMeta(factory string, targetName string, tag string) (*tuf.FileMeta, error) {
+	data, err := api.TufMetadataGet(factory, "targets.json", tag, false)
+	if err != nil {
+		if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
+			return nil, fmt.Errorf("the specified CI Target has not been found; target: %s, tag: %s", targetName, tag)
```
vkhoroz, line 231: This is actually: no CI targets were found for tag %s.

subcommands/targets/offline-update.go
```diff
@@ -163,6 +187,64 @@ func getTargetInfo(factory string, targetName string) (*ouTargetInfo, error) {
...
+		if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
+			return nil, fmt.Errorf("no active Wave with the specified name was found; wave: %s", wave)
+		}
+		return nil, fmt.Errorf("failed to get Wave Target metadata: %s", err.Error())
+	}
+	targets, ok := waveTargets[wave]
+	if !ok {
+		// it looks rather like panic since the call at #281 is supposed to return only the given wave targets
+		subcommands.DieNotNil(fmt.Errorf("no Targets are found for the given wave %s", wave))
+	}
```
vkhoroz, line 202: Just IMHO you're paying to much attention to these errors.
I'd just return `if err != nil { return nil, err }` at line 192 and be done.

At the least, lines 198-202 indeed try to handle an impossible use case... we're kind of not trusting our own API here.  Looks weird.

vkhoroz (commented):
Functionally it seems sane to me.

I only have comments to make the implementation easier to understand and/or shorter.

subcommands/targets/offline-update.go
```diff
@@ -35,15 +37,19 @@ var (
...
 
 func init() {
 	offlineUpdateCmd := &cobra.Command{
-		Use:   "offline-update <target-name> <dst> --tag <tag> [--prod] [--expires-in-days <days>] [--tuf-only]",
+		Use:   "offline-update <target-name> <dst> --tag <tag> [--prod | --wave <wave-name>] [--expires-in-days <days>] [--tuf-only]",
 		Short: "Download Target content for an offline update",
 		Run:   doOfflineUpdate,
 		Args:  cobra.ExactArgs(2),
 		Example: `
+	# Download update content of the wave target #1451 for "intel-corei7-64" hardware type
```
doanac, line 50: I'm not sure how to word this, but I think this command is getting complicated enough that we kind of need to briefly describe the "why" of this option and maybe the others so that people can figure out what options they need.

doanac (commented):
I see you are moving towards doing error messages with a lowercase first word. eg:
```
errors.New("this is an error")
```

This makes the Golang linters happier, but its not how we've traditionally done our other stuff. 

I think this is probably fine - we need to stop the bleeding, I suppose. But lets make a conscious decision that this is what we want to do.

Everything else seems pretty good - I'll leave it to @vkhoroz to look close at the new logic

subcommands/targets/offline-update.go
```diff
@@ -65,25 +71,54 @@ func init() {
 		"Skip fetching Target Apps")
 	offlineUpdateCmd.Flags().BoolVarP(&ouAllowMultipleTargets, "allow-multiple-targets", "", false,
 		"Allow multiple targets to be stored in the same <dst> directory")
+	offlineUpdateCmd.Flags().StringVarP(&ouWave, "wave", "", "",
+		"Name of the wave Target content of which to fetch")
```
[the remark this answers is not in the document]
mike-sul: >I would also keep this argument definition next to where we define the --prod flag (near line 64-65).

What's the difference?

subcommands/targets/offline-update.go
```diff
@@ -65,25 +71,54 @@ func init() {
...
 
-	if len(ouTag) == 0 {
-		subcommands.DieNotNil(errors.New("missing mandatory flag `--tag`"))
+	if len(ouTag) == 0 && len(ouWave) == 0 {
+		subcommands.DieNotNil(errors.New("one of the mandatory parameters `--tag` or `--wave` should be specified"))
+	}
+
+	if len(ouTag) > 0 && len(ouWave) > 0 {
+		subcommands.DieNotNil(errors.New("either `--tag` or `--wave` should be specified exclusively"))
 	}
```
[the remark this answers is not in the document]
mike-sul: `tag` is not a flag, so `cmd.MarkFlagsMutuallyExclusive("tag", "wave")` generates panic.

subcommands/targets/offline-update.go
mike-sul: As we discussed offline I am going to introduce additional custom field so, there won't be a need in the additional request for CI target. Therefore, all those `get*TargetMeta` functions are indeed metadata getter, not just checker.

subcommands/targets/offline-update.go
mike-sul: >I'd just return if err != nil { return nil, err } at line 192 and be done.

I did so in the first place. The error message printed by `fioctl` is really confusing. I don't understand it. So. those `if err != nil { if herr := client.AsHttpError(err);` are intended for printing understandable error message.


>At the least, lines 198-202 indeed try to handle an impossible use case... we're kind of not trusting our own API here. Looks weird.

So, just panic?

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
vkhoroz: A difference is that similar options are grouped closer to each other.

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
vkhoroz: I don't get why do you get a panic here.

Both `tag` and `wave` are flags in `init`, so they can be declared as mutually exclusive.
It looks like you did one of the following mistakes:
- called `MarkFlagsMutuallyExclusive` before flags are defined.
- called `MarkFlagsMutuallyExclusive` outside `init`.
- called `MarkFlagsMutuallyExclusive` on a wrong command (mst be a command where they are defined).

subcommands/targets/offline-update.go
mike-sul: Not sure I understand. I think a customer is the one who should decide on "why" - whether to compose an offline bundle for wave, or prod, or CI target, we just explain mechanics to do accomplish what they decided.

subcommands/targets/offline-update.go
vkhoroz: I buy an argument about `get*TargetMeta`, but it does not rejects a need to simplify the code.

If your forward-looking statement is valid, then the `checkAndGetTargetInfo` becomes even less intuitive, as the needed target could be fetched straight inside the `if..else` clauses (in the future), requiring less code.

subcommands/targets/offline-update.go
doanac: maybe its just content that belongs in docs, but we have all this stuff floating around: tags, CI, prod, and now wave. It could be intimidating for a regular user to navigate this. i'm fine ignoring this

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
mike-sul: I fixed the panic error. However, I don't get any error if none of `tag` nor `wave` is specified. Also, when I specify both of them the error message doesn't look as clean as it with the current approach.

subcommands/targets/offline-update.go
mike-sul: I think explanation should go the user doc. Not sure I see a good way to explain all that in the command comment.

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
vkhoroz: Let's separate things:

> I don't get any error if none of tag nor wave is specified

This is not what a `MarkFlagsMutuallyExclusive` does. This needs to be handled separately.

> Also, when I specify both of them the error message doesn't look as clean as it with the current approach.

This is a trade-off. Currently, we have a half dozen of use cases like this, and use `MarkFlagsMutuallyExclusive` is (almost) all of them.
I prefer one standard error format (which someone likes and someone dislikes) than a half dozen different errors each to someone's best liking.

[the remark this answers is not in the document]
mike-sul: I find the following error really confusing.

```
fioctl-fioed targets offline-update intel-corei7-64-lmp-6 offline/6.prod1 --tag dasd --wave asdfsdf
Error: if any flags in the group [tag wave] are set none of the others can be; [tag wave] were all set
Usage:
  fioctl targets offline-update <target-name> <dst> --tag <tag> [--prod | --wave <wave-name>] [--expires-in-days <days>] [--tuf-only] [flags]

Examples:

	# Download update content of the wave target #1451 for "intel-corei7-64" hardware type
	fioctl targets offline-update intel-corei7-64-lmp-1451 /mnt/flash-drive/offline-update-content --wave wave-deployment-001

	# Download update content of the production target #1451 tagged by "release-01" for "intel-corei7-64" hardware type
	fioctl targets offline-update intel-corei7-64-lmp-1451 /mnt/flash-drive/offline-update-content --tag release-01 --prod

	# Download update content of the CI target #1451 tagged by "devel" for "raspberrypi4-64" hardware type
	fioctl targets offline-update raspberrypi4-64-lmp-1448 /mnt/flash-drive/offline-update-content --tag devel --expires-in-days 15

	

Flags:
      --allow-multiple-targets   Allow multiple targets to be stored in the same <dst> directory
  -e, --expires-in-days int      Desired metadata validity period in days (default 30)
  -h, --help                     help for offline-update
      --no-apps                  Skip fetching Target Apps
      --prod                     Instruct to fetch content of production Target
      --tag string               Target tag
  -m, --tuf-only                 Fetch only TUF metadata
      --wave string              Name of the wave Target content of which to fetch

Global Flags:
  -c, --config string    config file (default is $HOME/.config/fioctl.yaml)
  -f, --factory string   Factory to list targets for
  -t, --token string     API token from https://app.foundries.io/settings/tokens/
  -v, --verbose          Print verbose logging

if any flags in the group [tag wave] are set none of the others can be; [tag wave] were all set
```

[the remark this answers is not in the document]
mike-sul: While this looks nice and clean.

```
fioctl-fioed targets offline-update intel-corei7-64-lmp-6 offline/6.prod1 --tag dasd --wave asdfsdf
ERROR: either `--tag` or `--wave` should be specified exclusively

```

subcommands/targets/offline-update.go
vkhoroz: The problem I had with current code is because the `checkAndGetTargetInfo` accepts two functions as arguments, that is already mindblowing for a regular engineer.
Now, a `getTargetMeta` in a function call magically becomes a `checkTargetMeta` inside `checkAndGetTargetInfo`, while a single-line anonymous function becomes a `getTargetMeta`.
On top of that this code defines a total of 4 anonymous functions, which are only called once - another red flag.

Long story short: it took a decent amount of mental power to understand what was going on.
That would be easier to understand with more procedural and less functional programming.

I know we all shine in various high-level OP paradigms, and at the end I totally understand the implementation.
I only try to care about a less advanced reader of it, and from that perspective I see no benefit of uber-functional programming here.

mike-sul: but it looks cool, isn't? :)

OK, let me try non-functional approach then.

vkhoroz: > but it looks cool, isn't? :)

Yes, it looks uber-cool :)

I often try functional programming to test my level of coolness, then scramble it, and commit to less shiny yet more readable procedural way.

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
mike-sul: @kprosise Please, suggest the better wording for `"Name of the wave Target content of which to fetch"` -  this is a description of the CLI parameter `wave`, a value of this parameter should be a wave name. A user specifies a wave name, then the command gets the wave targets and find the specified target among them.

subcommands/targets/offline-update.go
mike-sul: 63bac04

subcommands/targets/offline-update.go
mike-sul: 63bac04

subcommands/targets/offline-update.go
mike-sul: 2a8847d

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
mike-sul: I just moved `wave` param to the tag/prod group. 0665b26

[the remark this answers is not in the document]
kprosise: would a simplified  "Name of wave Target content to fetch" work?

subcommands/targets/offline-update.go
kprosise: @doanac is correct.

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
vkhoroz: @mike-sul @kprosise when I read your sentences I am under impression that this argument provides a `name of a (wave) Target`, but it provides `a name of a Wave`. We need to somehow separate these two nouns, so that confusion is removed.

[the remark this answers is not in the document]
vkhoroz: The other aspect to take into consideration is that we have a similar parameter `--prod` with [this description](https://github.com/foundriesio/fioctl/commit/0665b2626a840f0cdb24a990347307e6d6fc615d#diff-54ddea03a308be581ec1bc0426e2f90dcc0bdb21073eaa68ae351e32217b53ebL65):
```
      --prod                     Instruct to fetch content of production Target
```

Ideally, when we add a new parameter with a similar meaning, it needs to have a similar look and feel.
Namely, I mean that the updated help would ideally look like this:

```
      --prod                     Instruct to fetch content of production Target
      --wave string              Something with a similar look and feel as the above
```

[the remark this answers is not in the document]
mike-sul: Maybe as simple as `Instruct to fetch content of wave Target`?

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
vkhoroz: To me both messages look equally confusing; I have a different preference.

But, the _standard_ message is already used in several places, so our users should be already familiar with it.
It is like using an _auto-formatting tool_: nobody likes it, but it is a reasonable trade-off which eliminates opinions.

subcommands/targets/offline-update.go
```diff
@@ -72,18 +80,42 @@ func doOfflineUpdate(cmd *cobra.Command, args []string) {
...
+		_, targetGetErr = getProdTargetMeta(factory, targetName, ouTag)
+	} else {
+		fmt.Printf("Getting CI Target details; target: %s, tag: %s...\n", targetName, ouTag)
+		_, targetGetErr = getCiTargetMeta(factory, targetName, ouTag)
+	}
+	subcommands.DieNotNil(targetGetErr)
+	// Get the specified target from the list of factory targets to obtain the "original" tag/branch that produced
+	// the target, so we can find out the correct app bundle fetch URL.
+	// TBD: remove it once we do something so the wave/prod/CI specific target include enough info to deduce
+	// the download URLs
```
vkhoroz, line 110: What you put in a (TBD) comment is our _future roadmap_; so let it stay in a roadmap.
Let's keep the code clean, and not create a confusion as if this feature is not fully implemented.
Later we will simply extend this feature with an improved/better way... a long-lasting TBD does not help with that.

vkhoroz (approved):
LGTM

I only have opinions on some messages.
Functionally-wise this is a sound implementation.

Good job!

subcommands/targets/offline-update.go
```diff
@@ -163,6 +175,59 @@ func getTargetInfo(factory string, targetName string) (*ouTargetInfo, error) {
 	return &info, nil
 }
 
+func getWaveTargetMeta(factory string, targetName string, wave string) (*tuf.FileMeta, error) {
+	waveTargets, err := api.WaveTargetsList(factory, true, wave)
+	if err != nil {
+		if herr := client.AsHttpError(err); herr != nil && herr.Response.StatusCode == 404 {
+			return nil, fmt.Errorf("no active Wave with the specified name was found; wave: %s", wave)
```
vkhoroz, line 182: @doanac was commenting on this, but there is no (convenient) way to reply to that comment.

> I see you are moving towards doing error messages with a lowercase first word.
> This makes the Golang linters happier, but its not how we've traditionally done our other stuff.
> I think this is probably fine - we need to stop the bleeding, I suppose. But lets make a conscious decision that this is what we want to do.

So, I have an opinion on this.

I have no strong preference for Hungarian versus lowercase notation in error messages.
One use case where the Hungarian wins is when we have multi-sentence errors, which are rather common for the CLI (fioctl in particular).

IIUC we are using Hungarian-style (sentence-like) errors in the majority of places across Fioctl (yes, with first capital letter).
So, in order to look more professional, I'd recommend to continue using that way universally, until we decide to make a global change.
A third-party Golang linter is the last thing I care about when making such decisions.

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
mike-sul: @vkhoroz @kprosise What do think about the suggested above param descriptions?

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
mike-sul: @doanac What's your opinion on this?

subcommands/targets/offline-update.go
mike-sul: Agree, I put the comment basically for the reviewers, it might be useful to understand why I called the methods `get*TargetMeta` and why it returns a value.

subcommands/targets/offline-update.go
mike-sul: So, we have an consensus on the first capital letter? @doanac

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
kprosise: @mike-sul I like the second one with the semicolon.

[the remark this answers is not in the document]
mike-sul: Updated via force push.

subcommands/targets/offline-update.go
mike-sul: removed the TBD message dfb80e8.

subcommands/targets/offline-update.go
doanac: Let's stay with Capital letters and punt this decision for now

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
doanac: i prefer Mike's error message, but I agree with Volodya. Let's keep it consistent for now - maybe later we look into fixing Cobra to make a better message - which would fix all our places we use it.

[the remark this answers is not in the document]
mike-sul: Updated with the use of `MarkFlagsMutuallyExclusive` 143871c.

subcommands/targets/offline-update.go
mike-sul: With the use of `MarkFlagsMutuallyExclusive` it becomes irrelevant.

mike-sul:
Squashed all "fixup" commits into the single commit.

vkhoroz (approved):
LGTM

Looks clean.

PR #374, opened 2024-02-13
Fix: do not add keyEncipherment or keyAgreement to our TLS cert

vkhoroz:
These are not relevant for the EC key type.


vkhoroz:
@caiotpereira FYI

doanac (approved):

PR #378, opened 2024-02-28
fix(device): always take apps input as source of truth

StealthyCoder:
Instead of overriding the input given with what our api is giving back, it is much better to just take what the user is giving us as what needs to be set.

### DIFF
--- a/subcommands/common_config.go
+++ b/subcommands/common_config.go
@@ -223,19 +223,16 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
 	configuredApps := sota.GetDefault("pacman.docker_apps", "").(string)
 	configuredTag := sota.GetDefault("pacman.tags", "").(string)
 
-	if len(configuredTag) == 0 && len(reportedTag) > 0 {
-		configuredTag = reportedTag
-	}
-	if len(configuredApps) == 0 && reportedApps != nil {
-		configuredApps = strings.Join(reportedApps, ",")
-	}
-
 	changed := false
 	if opts.UpdateApps != "" && configuredApps != opts.UpdateApps {
 		if strings.TrimSpace(opts.UpdateApps) == "," {
 			opts.UpdateApps = ""
 		}
-		fmt.Printf("Changing apps from: [%s] -> [%s]\n", configuredApps, opts.UpdateApps)
+		fmt.Printf("Currently configured apps: [%s]\n", configuredApps)
+		if reportedApps != nil {
+			fmt.Printf("Apps reported as installed on device: [%s]\n", strings.Join(reportedApps, ","))
+		}
+		fmt.Printf("Setting apps to [%s]\n", opts.UpdateApps)
 		sota.Set("pacman.docker_apps", opts.UpdateApps)
 		sota.Set("pacman.compose_apps", opts.UpdateApps)
 		changed = true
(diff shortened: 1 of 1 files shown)


subcommands/common_config.go
```diff
@@ -226,16 +226,13 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
 	if len(configuredTag) == 0 && len(reportedTag) > 0 {
 		configuredTag = reportedTag
 	}
-	if len(configuredApps) == 0 && reportedApps != nil {
-		configuredApps = strings.Join(reportedApps, ",")
-	}
```
vkhoroz, line 231: I believe that the `reportedTag` is subject to the same problem as `reportedApps`.
So, it'd be wise to update that logic accordingly as well.

subcommands/common_config.go
```diff
@@ -226,16 +226,13 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
...
 
 	changed := false
 	if opts.UpdateApps != "" && configuredApps != opts.UpdateApps {
 		if strings.TrimSpace(opts.UpdateApps) == "," {
 			opts.UpdateApps = ""
 		}
-		fmt.Printf("Changing apps from: [%s] -> [%s]\n", configuredApps, opts.UpdateApps)
+		fmt.Printf("Changing apps to [%s]\n", opts.UpdateApps)
```
vkhoroz, line 235: Our users kind of liked the `from xyz to abc` thing.

After this change is approach, I think we need to extend the output we provide, not shrink it.
One idea that came to my mind is this:
```suggestion
		fmt.Printf(`
Currently configured apps: [%s]
Apps reported by the device: [%s]
Setting apps to [%s]
`, configuredApps, reportedApps, opts.UpdateApps)
```

Obviously, it is not that simple - there needs to be an if somewhere, as the `reported by device` part is only applicable to the device config.

vkhoroz (commented):
I think this is the right direction; yet I have suggestions how to further improve it.

subcommands/common_config.go
mike-sul: We may say `Apps installed on the device` or `Apps reported as installed on the device`.

subcommands/common_config.go
StealthyCoder: Done. :+1:

subcommands/common_config.go
StealthyCoder: Added for both tags and apps the new lines now. Let me know what you guys think.

subcommands/common_config.go
doanac: note - this is rolling back a "fix" - https://github.com/foundriesio/fioctl/commit/a8a21c70d9dfda5b2218bc93af0f52a6d6fe6508

doanac: or is it. i'm confused. can you add a little more to the commit message?

vkhoroz: @doanac IIUC that fix is nice, but it introduced another bug we were discussing today.
So, Eric has now moved it to a better place (see lines below).

vkhoroz (approved):
LGTM

doanac (approved):

subcommands/common_config.go
mike-sul: > note - this is rolling back a "fix" - [a8a21c7](https://github.com/foundriesio/fioctl/commit/a8a21c70d9dfda5b2218bc93af0f52a6d6fe6508)

I think that issue can be fixed by making `fioconfig` send the config to DG. It becomes possible with this change https://github.com/foundriesio/fioconfig/pull/45.

PR #390, opened 2024-03-28
fix(offline-update): Pull prod meta if wave is specified

mike-sul:
Make sure that the production root role metadata are pulled if a user specified that one of the wave targets should be downloaded and prepared for offline update.


vkhoroz (approved):
LGTM

PR #393, opened 2024-04-08
client: Disable compression in default transport

doanac:
The default Golang client will advertise to the HTTP server that it can handle gzip compressed response encoding. This is true, but it breaks our artifacts.go code for doing download progress.

This is because the default behavior of the transport is to:

 * provide the proper reader to decompress the response
 * update the content-length header to be -1

The -1 breaks our logic. This fix gets us back to how things used to work before CloudFlare started compressing the responses.

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -506,6 +506,13 @@ func NewApiClient(serverUrl string, config Config, caCertPath string, version st
 		InsecureSkipVerify: config.InsecureSkipVerify,
 	}
 	http.DefaultTransport.(*http.Transport).TLSClientConfig = tlsCfg
+	// targets/artifacts.go needs to know the Content-Length in order to
+	// compute the download progress. If certain services like CloudFlare
+	// see the client accepts compressed responsed (content-encoding not
+	// content-type) then it will give a compressed response. Golang will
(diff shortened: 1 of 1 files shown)


client/foundries.go
```diff
@@ -506,6 +506,7 @@ func NewApiClient(serverUrl string, config Config, caCertPath string, version st
 		InsecureSkipVerify: config.InsecureSkipVerify,
 	}
 	http.DefaultTransport.(*http.Transport).TLSClientConfig = tlsCfg
+	http.DefaultTransport.(*http.Transport).DisableCompression = true
```
vkhoroz, line 509: Maybe a simple comment like `for CloudFlare` here?
Git blame is not always much helpful for one-line fixes.

vkhoroz (approved):
Wow, I'm sorry for your nerves spent on this.

LGTM

mike-sul (approved):
LGTM. I wonder if there is easy option to make it configurable, so we don't turn it off for all types of requests/clients.

client/foundries.go
doanac: good call. fixed in force push

PR #394, opened 2024-04-16
targets: Use online root meta if not found locally

mike-sul:
If a user has not taken either root or target keys offline, then there are only two root versions in a given factory.
In such a case, the bundle will have no root metadata.
Therefore, to make the 'show' command work properly, the command needs to fetch the latest (i.e., version 2) root metadata from the backend.

### DIFF
--- a/subcommands/targets/offline-update.go
+++ b/subcommands/targets/offline-update.go
@@ -512,7 +512,10 @@ func getLatestRoot(bundleTufPath string) (*client.AtsTufRoot, error) {
 	if !errors.Is(readErr, os.ErrNotExist) {
 		return nil, readErr
 	}
-
+	if latestVersionBytes == nil {
+		// None of the N.root.json where N starts from 3 was found in the bundle
+		return nil, os.ErrNotExist
+	}
 	rootMeta := client.AtsTufRoot{}
 	if err := json.Unmarshal(latestVersionBytes, &rootMeta); err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


detsch (approved):
LGTM

subcommands/targets/offline-update.go
```diff
@@ -562,6 +565,12 @@ func doShowBundle(cmd *cobra.Command, args []string) {
 	}
 
 	rootMeta, err := getLatestRoot(tufMetaPath)
+	if errors.Is(err, os.ErrNotExist) && bundleMeta.ouBundleMeta.Type == "ci" {
+		// If no any N.root.json is found in the bundle and this is the "ci" bundle,
+		// then this is the valid case - a user has not taken their TUF targets key offline.
+		// Therefore, instead of failing the command fetches the root meta from the backend.
+		rootMeta, err = api.TufRootGet(viper.GetString("factory"))
+	}
```
vkhoroz, line 573: Do you mean that TUF root versions 1 and 2 are never included into the update bundle?
This seems strange, but I might be out of context.

vkhoroz (approved):
LGTM

subcommands/targets/offline-update.go
mike-sul: Yes. From the security standpoint of view this the right things to do, otherwise anyone can generate their own chain of TUF keys and put into a bundle. Then, a non-registered device will trust it.
However, we protect our devices from it by pre-provisioning them with the first N root versions...

PR #395, opened 2024-04-17
Feature: ability to filter devices by prod or non-prod flag

vkhoroz:


subcommands/devices/list.go
```diff
@@ -142,6 +144,8 @@ func init() {
 	}
 	cmd.AddCommand(listCmd)
 	listCmd.Flags().BoolVarP(&deviceMine, "just-mine", "", false, "Only include devices owned by you")
+	listCmd.Flags().BoolVarP(&deviceOnlyProd, "only-prod", "", false, "Only include production devices")
+	listCmd.Flags().BoolVarP(&deviceOnlyNonProd, "only-non-prod", "", false, "Only include non-production devices")
```
mike-sul, line 148: Is the `prod=0` going to output only non prod devices or all devices?

subcommands/devices/list.go
```diff
@@ -221,6 +226,8 @@ func doList(cmd *cobra.Command, args []string) {
 	}
 	dl, err := api.DeviceList(
 		deviceMine,
+		deviceOnlyProd,
+		deviceOnlyNonProd,
```
doanac, line 230: Seems like this could be a single flag and simplify the function args.

vkhoroz: I've tried to play with `*bool` type of a flag, and it resulted into a mess.
Counter-intuitively, in Golang using two boolean flags is much easier...

subcommands/devices/list.go
vkhoroz: Yes, the `--only-prod=0` is going to print all devices, which is kind of expected.

Although, I only expect the user to do `--only-prod` or simply omit that flag.

subcommands/devices/list.go
doanac: ah - so this is 3 cases. all devices, non-prod only, and prod-only. I guess this is probably fine, some type of enum might simplify, but we always think we are done adding options to this function

doanac (approved):

subcommands/devices/list.go
mike-sul: Then the `only-non-prod` is not quite correct name.

PR #397, opened 2024-04-19
A set of device list cleanups

vkhoroz:


subcommands/devices/list.go
```diff
@@ -220,24 +220,26 @@ func doList(cmd *cobra.Command, args []string) {
...
+	if deviceMine {
+		filterBy["mine"] = "1"
+	}
+	if deviceOnlyProd {
+		filterBy["prod"] = "1"
+	} else if deviceOnlyNonProd {
+		filterBy["prod"] = "0"
+	}
+
+	dl, err := api.DeviceList(filterBy, strings.Join(sortBy, ","), showPage, paginationLimit)
```
doanac, line 226: I'll give it an LGTM. I wonder if using a DeviceOpts struct wouldn't be a little cleaner assuming there's a relatively easy way to convert the struct into a dict net/url.Values ?

vkhoroz: Using `DeviceOpts` was my initial idea, but I decided that a simple `map[string]string` would result into less code churn long term.

As for the easiness converting struct to map, I'm not sure... One "easy" way I know is to do `json.Marshal/json.Unmarshal`, maybe there is also something inside the `reflect` package. But those are not easy IMHO. No wonder there are [3PP packages](https://github.com/fatih/structs) to do just that. Could you elaborate what do you meant by "easy" here.

subcommands/devices/list.go
```diff
@@ -228,7 +212,7 @@ func doList(cmd *cobra.Command, args []string) {
 		"uuid":        deviceUuid,
 	}
 	if len(args) == 1 {
-		filterBy["name_ilike"] = sqlLikeIfy(args[0])
```
mike-sul: Won't it break customer clients using the API directly as well as the old `fioctl` versions. I mean the `name_ilike` removal?

vkhoroz: The `name_ilike` was not removed on the API side...
It is still supported there, but was deprecated quite a few months ago.
So, no clients were broken.

It is only the Fioctl finally moving to a newer API argument.

mike-sul: I've already checked the backend part and see that both params are supported.

mike-sul (approved):

client/foundries.go
```diff
@@ -787,10 +787,16 @@ func (a *Api) DeviceList(filterBy map[string]string, sortBy string, page, limit
 	url := a.serverUrl + "/ota/devices/?"
 	query := netUrl.Values{}
 	for key, val := range filterBy {
-		query.Set(key, val)
+		if len(val) > 0 {
+			query.Set(key, val)
```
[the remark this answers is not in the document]
vkhoroz: They are not... Each individual argument goes separately, ust as before.
The only "functional" change of this PR is that they are now properly escaped.

...Besides that `name_ilike -> name` change.

PR #411, opened 2024-05-29
targets: Handle absence of apps for offline update

mike-sul:
Apps might be explicitly turned off or intersection of target apps and the app shortlist specified for offline update is empty (for specific tag). In this case there is no a download URI to fetch an app tarball for a given bundle. This change handles this case properly instead of yielding an error.


vkhoroz (approved):
Good catch!

PR #414, opened 2024-06-05
Rework `device` subcommand and client APIs to support lookup by UUID

doanac:
This series of changes ultimately moves fioctl to support device commands where
they can be addressed by UUID instead of name. e.g.:
```
fioctl devices show --by-uuid xxx-xxx-xxxx
fioctl devices config log -n1 -u xxx-xxx-xxxx
```

To get there I cleaned up the client APIs to make this more organized and support
a "device api". This separates a `Device` object - which requires a device *lookup*
from a `DeviceApi` which can make API calls w/o looking up a device.


subcommands/devices/config_log.go
```diff
@@ -29,11 +29,13 @@ func doConfigLog(cmd *cobra.Command, args []string) {
 	lookups, err := api.UsersGetLookups(factory)
 	subcommands.DieNotNil(err)
 
+	dapi := api.DeviceApiByName(factory, device)
```
vkhoroz: Any reason this is `dapi` and not `d` like in other places?

subcommands/devices/cmd.go
```diff
@@ -61,3 +62,14 @@ func NewCommand() *cobra.Command {
...
+func getDeviceApi(_ *cobra.Command, name string) client.DeviceApi {
+	return api.DeviceApiByName(viper.GetString("factory"), name)
+}
+
+func getDevice(cmd *cobra.Command, name string) *client.Device {
+	dapi := getDeviceApi(cmd, name)
+	d, err := dapi.Get()
+	subcommands.DieNotNil(err)
+	return d
+}
```
vkhoroz, line 85: It seems to me that this change may be higher in the ~food chain~ commit list.
Then, the preceeding commits could gradually turn commands to the final desired shape (presented here), without an intermediate changes (and then one big commit) like it is now.

My gut feeling is that it would be easier to navigate this PR (and cause less code churn).
WDYT?

subcommands/devices/cmd.go
```diff
@@ -59,11 +59,21 @@ func NewCommand() *cobra.Command {
...
 
 	cmd.AddCommand(configCmd)
-	cmd.AddCommand(updatesCmd)
+	addUuidCommand(cmd, updatesCmd)
+
 	return cmd
 }
 
-func getDeviceApi(_ *cobra.Command, name string) client.DeviceApi {
+func addUuidCommand(parent, command *cobra.Command) {
```
vkhoroz, line 67: I think the `addUuidFlag` name would look more intuitive.
We have functions with `addXyzFlag` already, so it is a de-facto pattern already.

subcommands/devices/cmd.go
```diff
@@ -59,11 +59,21 @@ func NewCommand() *cobra.Command {
...
+	command.Flags().BoolP("by-uuid", "u", false, "Look up device by UUID rather than name")
+	parent.AddCommand(command)
```
vkhoroz, line 69: Just IMHO.

What I dislike here is that this function combines 2 concepts:
- Adding a flag.
- Adding a command to a parent.

Although it allows to cut 1 line of code in each subcommand, it looks like a violation of the _separation of concepts_.
Just IMHO the code looks less intuitive now, that is apparently my _subjective interpretation_.
But a stronger argument is that now applying 2 addXyzFlag functions (where needed) becomes more intricate, as one of them also does an extra task, and the other one must not do the same.

subcommands/devices/show.go
```diff
@@ -35,6 +35,7 @@ func doShow(cmd *cobra.Command, args []string) {
 	device := getDevice(cmd, args[0])
 
 	fmt.Printf("UUID:\t\t%s\n", device.Uuid)
+	fmt.Printf("Name:\t\t%s\n", device.Name)
```
vkhoroz, line 38: This looks like a separate functional change.
It clearly belongs to this PR, and I get why it is here...
...it still deserves a separate commit with a clear message.

vkhoroz (commented):
Overall, good job.

It is a bit extra-ordinary, but in the end a really clever way to implement this.

My only comments are called to make the commit chain more refined.
Why? It was a pleasure to review this up until the last 2 commits...
...Up until that point I had a feeling that everything evolves gradually, hence is under control.
...The one before last commit is a kind of an overhaul of these _nice intermediate commits_ which IMHO could be avoided by changing the commit order.

doanac:
> Why? It was a pleasure to review this up until the last 2 commits...

Ha - you caught me. I had a completely clear vision of the clean ups and then on the last two was sort of feeling my way through the dark. I'll clean that up.

subcommands/devices/show.go
doanac: I was a little on the fence there also. It wasn't necessary(possible redundant) until the addition of the `--by-uuid` flag. i'll split

subcommands/devices/cmd.go
doanac: there were several call sites where we were doing `cmd.AddCommand`, so the diff was growing in size to declare a local variable for the subcommand and then call AddCommand. I was also thinking it might help reduce the chance of someone forgetting to add the uuid-flag, but I don't think my approach lessens that.

I was about to the point of creating a zlast_alpabetical_file.go with an `init` function that is guaranteed to run last where I fix every subcommand. But that is just awful.

I'll clean this up to be more intentional.

doanac: I just tried a slightly different approach at the end. I moved the commands over one by one to the getDevice function. Then in one small change I could add the `-by-uuid` flag to all commands from a single place. Its almost too clever - but I think it makes the chances of someone messing things up smaller.

subcommands/devices/cmd.go
```diff
@@ -60,10 +62,28 @@ func NewCommand() *cobra.Command {
...
-func getDeviceApi(_ *cobra.Command, name string) client.DeviceApi {
+func addUuidFlagToChildren(c *cobra.Command) {
+	ignores := []string{"list-denied", "list", "delete-denied"}
+	for _, child := range c.Commands() {
+		if child.HasSubCommands() {
+			addUuidFlagToChildren(child)
+		} else if !slices.Contains(ignores, child.Name()) {
+			child.Flags().BoolP("by-uuid", "u", false, "Look up device by UUID rather than name")
+		}
+	}
```
vkhoroz, line 79: This is somewhat implicit, but it looks smart.

Sometimes I look at commands declaration to get the thing about its flags.
That is convenient when I have no access to terminal with Fioctl help and need to quickly answer some question using just GitHub pages.
This change kind of takes that possibility away.

But I think we may make an exception for this use case, as it it rather special.
...As long as we agree to _keep it special and not copy this pattern to other places_.

subcommands/devices/cmd.go
```diff
@@ -60,10 +62,28 @@ func NewCommand() *cobra.Command {
...
+}
+
+func getDeviceApi(cmd *cobra.Command, name string) client.DeviceApi {
+	byUuid, _ := cmd.Flags().GetBool("by-uuid")
+	if byUuid {
```
vkhoroz, line 84: This code makes an assumption that one of these conditions is met:
- The `by-uuid` flag is present in a command.
- The command without `by-uuid` flag does not call `getDeviceApi`.
- The `byUuid` is always `false` if the `by-uuid` flag is not defined in a command.

Of these, only the third condition is correct, and that only because of the `GetBool` internal implemention detail.
That nuance is IMHO to subtle to rely on it.
I think a more refined code would check for an error (an absence of flag definition) here, like below:
```suggestion
	byUuid, ok := cmd.Flags().GetBool("by-uuid")
	if ok && byUuid {
```

vkhoroz (approved):
LGTM

This looks nice.

subcommands/devices/cmd.go
doanac: i agree. I hope we don't have to do this again.

subcommands/devices/cmd.go
doanac: maybe we should `panic` if !ok - this would be a programming error, right?

vkhoroz: If you want to enforce that option - then maybe yes.
IMHO if it were us who see this panic, I'd say yes... but it will the the user who sees it.
I doubt that we want to punish the user because of our own lack of attention.

So, I'd rather assume that command without `by-uuid` only supports lookup by name.
Your code does that already, and I'm fine with that. I just wanted to make it explicit.

doanac: I force pushed a change to log an error but continue. Here's what it would look like (i forced the `show` command to fail for demo purposes):
```
$ ./bin/fioctl-darwin-arm64 devices show og
ERROR: flag accessed but not defined: by-uuid
UUID:		bfc97632-d493-4f7e-95b0-ec32ef190893
```

vkhoroz (approved):
This is good!

PR #433, opened 2024-08-19
WIP: rotate root CA

vkhoroz:
**This is incomplete and not yet fully tested amid API changes not ready too.**

@mike-sul @doanac @StealthyCoder @camilamacedo86 @detsch I'd like to start an internal team discussion on this first.

Please, pay attention to the UI flow, and not the implementation details in the first iteration.

While working on this I've completely rewritten the commands at least twice, and I am not happy with the current flow yet.
A reason is: it is hard to find a proper trade off between the user control and UI simplicity:
- Counter-intuitively, a simpler UI would mean that the user would screw up _more often_, given that the operation needs to be as extended in time as possible.
- While a more complex UI _apparently_ leads to a harder user engagement and more work for the customer success/support teams.


subcommands/keys/ca_renewal_resign_device_ca.go
```diff
@@ -0,0 +1,66 @@
...
+		Args:  cobra.RangeArgs(1, 2),
+		Long: `Re-sign all existing Device CAs with a new root CA for your Factory PKI.
+
+Both currently active and disabled Device CAs are being re-signed.
+All their properties are preserved, including a serial number.
+Only the signature and authority key ID are being changed.
+This allows old certificates (issued by a previous root CA) to continue being used to issue device client certificates.
+
+Re-signed device CA certificates are stored in the provided PKI directory.
+An old PKI directory is used to locate corresponding private keys, and copy them into the PKI directory.
```
doanac, line 29: shouldn't we use what we know from the server `fioctl keys ca show --just-device-cas`. Trusting this local directory is correct seems error-prone?

doanac (commented):
Can you walk us through the flow you see? I'm guessing its:

 * `fioctl keys ca renewal start`
 * `fioctl keys ca renewal resign-device-ca`

At that point I'm not sure if i do `active` or the config operations.

subcommands/keys/ca_renewal_resign_device_ca.go
vkhoroz: We need to copy private keys from an old PKI directory to the new PKI directory. Otherwise, a user might lose them when deleting an old directory.

We don't have those private keys on our servers. But, we have their public parts which we can use for a match.

vkhoroz:
@doanac sorry, I should have posted the command order initially.
It was planned as a help message in the final commit... but as it is not there yet, here is the flow:

```
# Start CA root renewal
fioctl keys ca renewal start

# Provision renewal bundle to all devices
fioctl config renew-root
# Just think about an alternative notation to the above command like this:
fioctl keys ca renewal provision | deploy [--device <uuid> | --group <name>]

# Some time after all devices were updated, activate the new CA root, and then re-sign all certificates
fioctl keys ca renewal activate
fioctl keys ca renewal re-sign-device-ca
fioctl keys ca renewal re-sign-tls

# Finally, two optional steps to revoke the previous root CA and provision that change to all devices
fioctl keys ca renewal revoke-root
fioctl config renew-root
```

Does it make sense to you?

StealthyCoder:
I only have a first comment. I do not think the flow is that bad, the only suggestion I would do is make `fioctl config renew-root` part of `fioctl keys ca renewal start` and `fioctl keys ca renewal revoke-root`. Then all commands are within `fioctl keys ca renewal` "namespace". Then the big question is of course, when can someone determine if all devices have received the new CA ?

mike-sul:
@vkhoroz Is it possible to try and experiment with the new functionality against the backend version that supports it? Maybe something running in the meds/dev environment? I guess to test it e2e will require fioconfig changes too?

vkhoroz:
>Is it possible to try and experiment

@mike-sul The API changes are not ready at this stage.
Fioconfig PoC is here, but it is still not tested: https://github.com/foundriesio/fioconfig/pull/51

subcommands/keys/ca_renewal.go
```diff
@@ -0,0 +1,20 @@
+package keys
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var caRenewalCmd = &cobra.Command{
+	Use:   "renewal",
+	Short: "Renew the root of trust for your factory PKI",
+	Long: `These sub-commands allow you to gradually renew a root of trust for your factory PKI.
```
doanac, line 10: factory-> Factory

subcommands/keys/ca_renewal.go
```diff
@@ -0,0 +1,20 @@
...
+
+A guided process allows you to:
+- Generate a new root of trust.
+- Create an EST standard compliant root CA renewal bundle.- Re-sign all necessary factory PKI certificates.
```
doanac, line 14: line break missing there

subcommands/keys/ca_renewal_start.go
```diff
@@ -0,0 +1,92 @@
...
+2. Second, it cross-signs a new root of trust using an old root of trust to prepare a CA renewal bundle.
+   This CA renewal bundle is compliant with the EST standard, necessary for a secure root CA update on devices.
+3. Finally, it uploads the CA renewal bundle to the backend API for validation and storage.
+
+A new root of trust needs to be stored in a separate directory from the previous root of trust.
+By the end of the renewal process, all necessary PKI pieces are migrated into this new directory.
+If you are using an HSM device - a new private key should be stored under a different label, possibly a new device.
+This extreme level of isolation is necessary to prevent an accidental rewrite of the old root of trust.
+
+Once this command completes successfully, a root of trust renewal process is started.`,
```
doanac, line 41: kind of a nitpick, but I think you are doing newlines by sentence. Cobra doesn't handle smart line breaks or anything, so it would probably render better if you just hard wrap a column 72 or 80.

subcommands/keys/ca_renewal_start.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	caRenewalCmd.AddCommand(cmd)
+}
+
+func doStartCaRenewal(cmd *cobra.Command, args []string) {
+	factory := viper.GetString("factory")
+	newCertsDir := args[0]
+	oldCertsDir := args[1]
+	if oldCertsDir == newCertsDir {
```
doanac, line 60: probably should normalize these paths before doing this check so you don't get messed up by something like `start ./ $(pwd)`

subcommands/keys/ca_renewal_start.go
```diff
@@ -0,0 +1,92 @@
...
+		Short: "Start a root of trust renewal for your factory PKI",
+		Run:   doStartCaRenewal,
+		Args:  cobra.ExactArgs(2),
+		Long: `Start a root of trust renewal for your factory PKI.
+
+This command does a few things:.
+1. First, it generates a new root of trust for your factory.
+2. Second, it cross-signs a new root of trust using an old root of trust to prepare a CA renewal bundle.
+   This CA renewal bundle is compliant with the EST standard, necessary for a secure root CA update on devices.
+3. Finally, it uploads the CA renewal bundle to the backend API for validation and storage.
```
doanac, line 34: not sure where to ask in the PR, but I'm wondering if there's going to wind up being an easy way to check "if in progress" and also make sure two different people don't try this at the same time?

subcommands/config/renew_root.go
```diff
@@ -0,0 +1,85 @@
...
+	serverName, _ := cmd.Flags().GetString("server-name")
+	dryRun, _ := cmd.Flags().GetBool("dryrun")
+
+	if estResource[0] != '/' {
+		estResource = "/" + estResource
+	}
+
+	if len(group) > 0 {
+		logrus.Debugf("Renewing device root CA for devices in group %s", group)
+	} else {
```
doanac, line 49: I think we've decided we should stop adding factory wide config operations like this - they are too dangerous.

doanac (commented):
I think this is looking sane. I've added a couple of comments.

Now that you are getting the mechanics of this out of the way: I think we need to think about how we can understand how we can determine the status of this as it gets rolled out

subcommands/keys/ca_renewal_start.go
vkhoroz: Do we still have 640x320 terminals anywhere?
AFAIK we're doing line breaks by sentence across Fioctl for ages, with some exceptions.
Man pages usually do breaks by sentence or by paragraph.

IMHO 80-char breaks look ugly on modern terminals.
And on old terminals, there is still a line wrap, so we're fine.

subcommands/keys/ca_renewal_start.go
vkhoroz: I planned to add a kind of `review|status` command similar to the one we have for TUF rotations.
Just haven't got idea about its design yet... and it is highly dependent onto what I do on the API side.

PR #437, opened 2024-10-22
fix(git): expand user home dir first

StealthyCoder:
When using it like `fioctl configure-git --creds-path ~/.config` there is no problem however if you use it like `fioctl configure-git --creds-path=~/.config` there is.

The reason lies in the fact that in the first example the shell already expands the ~ into the absolute path of the homedir, in the second it will not.

### DIFF
--- a/subcommands/git/cmd.go
+++ b/subcommands/git/cmd.go
@@ -19,9 +19,7 @@ import (
 
 const GIT_CREDS_HELPER = "git-credential-fio"
 
-var (
-	helperPath string
-)
+var helperPath string
 
 func NewCommand() *cobra.Command {
 	path, err := exec.LookPath("git")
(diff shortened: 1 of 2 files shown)


subcommands/git/cmd.go
```diff
@@ -74,6 +72,11 @@ func doGitCreds(cmd *cobra.Command, args []string) {
 	var gitHelperCommandArgs []string
 
 	helperName := "fio"
+	if strings.HasPrefix(helperPath, "~/") {
```
doanac, line 75: this is tricky:
 https://github.com/golang/go/issues/4140

I think this would also happen in the docker helper. After reading those bug comments like "what about Win32", I'm more inclined to check `string.contains("~")` and just return an error like golang chose:
 https://github.com/golang/go/commit/8b6534b78af791b80f371857a15d76bbc10fd012#diff-7c289d314f17f4e276724c62cfd8dd5e2c7dffad2d900987f8eae83192ad15bdL131

vkhoroz: I agree to this.

It also raises a ton of questions like _what about `fioctl configure-git --creds-path='~/.config'`?_
(In the above case there must be no expansion, but single quotes are not passed to the program by shell).

I'd say: let the shell do its job, and let's not pretend to "fix" it... as there are so many use cases we're burrowing a rabbit hole.

StealthyCoder: This makes sense to me. I adjusted it to also just throw an error for both `git` and `docker`. The only difference I did, was to always throw an error regardless of OS. I actually think this will be beneficial to the end user since it conveys more information.

vkhoroz (approved):
LGTM

PR #439, opened 2024-10-31
offline-updates: Allow specification of local ostree repo

detsch:
Allowing the specification of a local ostree repo during offline bundle creation is required when a custom ostree repo is in use (`fiopush` + `fioctl targets add`).

This PR also add a warning message if the ostree repo does not contain the expected hash.

### DIFF
--- a/subcommands/targets/offline-update.go
+++ b/subcommands/targets/offline-update.go
@@ -3,6 +3,7 @@ package targets
 import (
 	"archive/tar"
 	"compress/bzip2"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -32,6 +33,7 @@ type (
 		hardwareID    string
 		buildTag      string
 		fetchedApps   *client.FetchedApps
+		sha256        []byte
 	}
 
 	ouBundleMeta struct {
@@ -180,6 +182,14 @@ Notice that multiple Targets in the same directory is only supported in LmP >= v
 			fmt.Printf("Downloading an ostree repo from the Target's OE build %d...\n", ti.ostreeVersion)
 			subcommands.DieNotNil(downloadOstree(factory, ti.ostreeVersion, ti.hardwareID, dstDir), "Failed to download Target's ostree repo:")
 		}
+
+		sha256String := base64.StdEncoding.EncodeToString(ti.sha256)
+		expectedCommit := path.Join(dstDir, "ostree_repo", "objects", sha256String[0:2], sha256String[2:]+".commit")
+		if _, err := os.Stat(expectedCommit); errors.Is(err, os.ErrNotExist) {
+			fmt.Printf("ERROR: ostree repo does not contain target's hash %s. If the target references a custom ostree repo, re-run specifying --ostree-repo-source.\n", sha256String)
+			os.Exit(1)
+		}
+
 		if !ouNoApps {
 			if (len(ouWave) > 0 || ouProd) && ti.fetchedApps == nil {
 				// Get the specified target from the list of factory targets to obtain the "original" tag/branch that produced
@@ -233,6 +243,13 @@ func getTargetInfo(targetFile *tuf.FileMeta) (*ouTargetInfo, error) {
 	}
 	info.hardwareID = custom.HardwareIds[0]
 	info.fetchedApps = custom.FetchedApps
+	sha256, ok := targetFile.Hashes["sha256"]
+	if ok {
+		info.sha256 = sha256
+	} else {
+		fmt.Printf("ERROR: Target has no sha256 hash set")
+		os.Exit(1)
+	}
 	if info.fetchedApps == nil {
 		info.buildTag = custom.Tags[0] // See the assemble.py script in ci-scripts https://github.com/foundriesio/ci-scripts/blob/18b4fb154c37b6ad1bc6e7b7903a540b7a758f5d/assemble.py#L300
 	}
--- a/subcommands/targets/offline-update.go
+++ b/subcommands/targets/offline-update.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -52,6 +53,7 @@ var (
 	ouNoApps               bool
 	ouAllowMultipleTargets bool
 	ouWave                 string
+	ouOstreeRepoSrc        string
 )
 
 func init() {
@@ -87,6 +89,8 @@ func init() {
 		"Skip fetching Target Apps")
 	offlineUpdateCmd.Flags().BoolVarP(&ouAllowMultipleTargets, "allow-multiple-targets", "", false,
 		"Allow multiple Targets to be stored in the same <dst> directory")
+	offlineUpdateCmd.Flags().StringVarP(&ouOstreeRepoSrc, "ostree-repo-source", "", "",
+		"Path to the local ostree repo to be added to the offline bundle")
 	offlineUpdateCmd.MarkFlagsMutuallyExclusive("tag", "wave")
 	offlineUpdateCmd.MarkFlagsMutuallyExclusive("prod", "wave")
 	initSignCmd(offlineUpdateCmd)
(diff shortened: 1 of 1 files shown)


detsch:
@mike-sul please take a look. I can't add reviewers to the PR.

subcommands/targets/offline-update.go
```diff
@@ -366,6 +374,64 @@ func downloadTufRepo(factory string, target string, tag string, prod bool, wave
...
+		}
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func copyOstree(srcDir string, dstDir string) error {
+	return copyDirRecursive(srcDir, dstDir)
```
vkhoroz, line 447: Some time in the future (with Golang 1.23) we might use `os.CopyFS`.
For now, using `filepath.Walk` seems optimal to me e.g. smth like below:

```
import (
	"os"
	"os/filepath"
)

func copyRecursive(src, dst string) error {
	copyFile = func(src, dst string) error {
		// Basically, your copyFile function less the fs.Stat nad os.Chmod.
		srcFile, err := os.Open(src)
		if err != nil {
			return nil
		}
		defer srcFile.Close()

		dstFile, err := os.Create(dst)
		if err != nil {
			return err
		}
		defer dstFile.Close()

		return dstFile.ReadFrom(srcFile)
	}

	return filepath.Walk(src, func(srcPath string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}
		dstPath := strings.Join(dst, strings.Replace(srcPath, src, "", 1))
		if info.IsDir() {
			return os.Mkdir(dstPath, info.Mode())
		} else {
			if err = copyFile(srcPath, dstPath); err != nil {
				return err
			}
			return os.Chmod(dstPath, info.Mode())
		} 
	})
}
```

vkhoroz (commented):
I defer the review to @mike-sul indeed.

Just recommending to use stdlib to a greater extent.

subcommands/targets/offline-update.go
vkhoroz: A few notes on the above:
1. Using `filepath.Walk` saves some lines of code.
2. IIUC `os.Chmod` needs to be called after `file.Close` to avoid concurrent file writes (which are dangerous). That's why `os.Chmod` is called outside a function with deferred `file.Close` call.
3. IIRC `filepath.Walk` replaces a stack-based recursion with a tree and a loop (more efficient).
4. You call `os.Stat` on each dir entry, hence, leveraging `Walk` (and not `WalkDir`) makes sense.

detsch: Thanks, that makes sense. I've updated the code following your suggestion (with minor adjustments), and functionality was not affected.

subcommands/targets/offline-update.go
```diff
@@ -87,6 +89,8 @@ func init() {
 		"Skip fetching Target Apps")
 	offlineUpdateCmd.Flags().BoolVarP(&ouAllowMultipleTargets, "allow-multiple-targets", "", false,
 		"Allow multiple targets to be stored in the same <dst> directory")
+	offlineUpdateCmd.Flags().StringVarP(&ouOstreeRepoSrc, "ostree-repo-source", "", "",
+		"Path to the local ostree repo to be used in the offline bundle")
```
mike-sul, line 95: Maybe "to be added to the offline bundle" or "to be copied to"?

subcommands/targets/offline-update.go
```diff
@@ -366,6 +375,44 @@ func downloadTufRepo(factory string, target string, tag string, prod bool, wave
...
+		}
+		defer srcFile.Close()
+
+		dstFile, err := os.Create(dst)
+		if err != nil {
+			return err
+		}
+		defer dstFile.Close()
+
+		_, err = dstFile.ReadFrom(srcFile)
```
mike-sul, line 407: I don't have strong opinion, just `io.Copy` seems like would be more flexible.

subcommands/targets/offline-update.go
```diff
@@ -180,6 +182,15 @@ Notice that multiple targets in the same directory is only supported in LmP >= v
 			fmt.Printf("Downloading an ostree repo from the Target's OE build %d...\n", ti.ostreeVersion)
 			subcommands.DieNotNil(downloadOstree(factory, ti.ostreeVersion, ti.hardwareID, dstDir), "Failed to download Target's ostree repo:")
 		}
+
+		if ti.sha256 != nil {
+			sha256String := base64.StdEncoding.EncodeToString(ti.sha256)
+			expectedCommit := path.Join(dstDir, "ostree_repo", "objects", sha256String[0:2], sha256String[2:]+".commit")
+			if _, err := os.Stat(expectedCommit); errors.Is(err, os.ErrNotExist) {
+				fmt.Printf("WARNING: ostree repo does not contain target's hash %s, installation will fail. If the target references a custom ostree repo, re-run specifying --ostree-repo-source.\n", sha256String)
```
mike-sul, line 190: IMO, this is the critical error, it doesn't make sense to ship bundle to devices if the commit hash is missing, so I would just exit/die at this spot.

subcommands/targets/offline-update.go
```diff
@@ -180,6 +182,15 @@ Notice that multiple targets in the same directory is only supported in LmP >= v
 			fmt.Printf("Downloading an ostree repo from the Target's OE build %d...\n", ti.ostreeVersion)
 			subcommands.DieNotNil(downloadOstree(factory, ti.ostreeVersion, ti.hardwareID, dstDir), "Failed to download Target's ostree repo:")
 		}
+
+		if ti.sha256 != nil {
```
mike-sul, line 186: The hash `ti.sha256` is mandatory, it is `nil` I would just die.

subcommands/targets/offline-update.go
vkhoroz: I thought `io.Copy` could replace the entire function, but unfortunately it does not.

No idea what flexibility you mean... a doc says this:
```
If src implements [WriterTo](https://pkg.go.dev/io#WriterTo),
the copy is implemented by calling src.WriteTo(dst).
Otherwise, if dst implements [ReaderFrom](https://pkg.go.dev/io#ReaderFrom),
the copy is implemented by calling dst.ReadFrom(src).
```
So it seems to only add an extra function call.

subcommands/targets/offline-update.go
```diff
@@ -366,6 +375,44 @@ func downloadTufRepo(factory string, target string, tag string, prod bool, wave
...
 
+func copyRecursive(src, dst string) error {
+	copyFile := func(src, dst string) error {
+		srcFile, err := os.Open(src)
+		if err != nil {
+			return nil
+		}
+		defer srcFile.Close()
+
+		dstFile, err := os.Create(dst)
```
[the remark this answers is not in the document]
vkhoroz: (2) probably makes sense... although I'm not a great fan of a trade-off (added complexity).

As for (1), I think it comes with inherent risk of copying wrong files.
Think of it this way: a user might supply a wrong source, interrupt, and then supply a correct source.
If we don't forcibly replace all files, we can end up with the wrong files.

Not sure how accurate are my statements. Maybe, ostree has a directory structure preventing the above.
So, I leave it up to you to weigh on this.

subcommands/targets/offline-update.go
mike-sul: By flexibility I meant exactly what the doc extract says - it agnostic to details of src and dst implementation. Their type change won't need to change io.Copy() call.
As I said, I don't have strong opinion, just looks more uniform to use `io.Copy`.

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
mike-sul: (2) is important if a user copies files to external thumb drive to guarantee data are synced.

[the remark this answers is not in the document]
mike-sul: >a user might supply a wrong source, interrupt, and then supply a correct source.

I don't see how (1) would cause an issue here taking into account (2) and the fact that each file in ostree repo has unique name - its hash.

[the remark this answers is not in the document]
vkhoroz: > each file in ostree repo has unique name - its hash

Told you, I can be missing something.

> how (1) would cause an issue here

But, I must be backing my claims about safety still.
You see, fsync only makes sure to sync the write to the hardware level, but that data may still be stored in the hardware (firmware) level cache first, and only later written to the persistent storage.
So, if we are talking about e.g. a thumb drive, than I'd say to do one of the following:
1. Always rewrite - easy way.
2. If file exists - check the file size/hash, and rewrite if it differs.

Trust me, I remember those times when I had to trash 1 in 100 CDs because the fsync is not as reliable as assumed.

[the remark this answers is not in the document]
mike-sul: check&write vs rewrite and "write to tmp" and fsync are two different things.
If we write to `tmp` then a file does not exist until it's fsynced and renamed to the file correct name. If we assume that "fsync and rename" are 99.9999% guarantee that data are persisted then we don't need to check file hashes. We use this techniques in fioconfig and aklite and composectl and we haven't seen any issues with it so far.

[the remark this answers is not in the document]
vkhoroz: Yes, you're right.
Fsync to tmp and then rename should do the trick.

Probably in this specific use case (writing a lot of small files) it is worth an associated complexity.

subcommands/targets/offline-update.go
detsch: Now using `to be added to`

subcommands/targets/offline-update.go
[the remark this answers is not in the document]
detsch: Now using .tmp file + sync + move.

subcommands/targets/offline-update.go
detsch: Changed to ERROR, and exiting

mike-sul (approved):

vkhoroz:
@doanac I think it is worth to include this into the short-coming release too.
WDYT?

doanac:
yes. we should include this

subcommands/targets/offline-update.go
```diff
@@ -366,6 +375,57 @@ func downloadTufRepo(factory string, target string, tag string, prod bool, wave
...
+
+		tmpDstName := dst + ".tmp"
+		dstFile, err := os.Create(tmpDstName)
+		if err != nil {
+			return err
+		}
+		defer dstFile.Close()
+
+		_, err = dstFile.ReadFrom(srcFile)
+		dstFile.Sync()
```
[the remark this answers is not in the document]
detsch: Yep, added checks there now.

mike-sul:
I think it is ready for merge.

PR #440, opened 2024-11-04
feat(config): make it so you can revert override of apps

StealthyCoder:
This is to add a way to disable the override, inherit from parent concerning apps running on devices. 

[Relevant fioconfig code](https://github.com/foundriesio/fioconfig/blob/c3591ccefd4d1aab063d83d0cdc881308925efe8/contrib/aktualizr-toml-update#L36-L40)


subcommands/common_config.go
```diff
@@ -235,6 +235,16 @@ func SetUpdatesConfig(opts *SetUpdatesConfigOptions, reportedTag string, reporte
...
 		sota.Set("pacman.compose_apps", opts.UpdateApps)
+		if strings.TrimSpace(opts.UpdateApps) == "-" {
+			err = sota.Delete("pacman.docker_apps")
+			if err != nil {
+				fmt.Printf("Error trying to delete pacman.docker_apps key: %s\n", err)
+			}
+			sota.Delete("pacman.compose_apps")
+			if err != nil {
+				fmt.Printf("Error trying to delete pacman.compose_apps key: %s\n", err)
+			}
```
vkhoroz, line 246: Looking at `sota.Delete` IIUC there are two use cases when it can return an error:
1. A key is malformed.
2. A key is not found.

Here, (1) is a programming error deserving `DieNotNil`, while (2) would be better off silently ignored.
Thus, I suggest to improve this code using `sota.Has`.

```suggestion
if sota.Has("pacman.docker_apps") {
    subcommands.DieNotNil(sota.Delete("pacman.docker_apps"))
}
if sota.Has("pacman.compose_apps") {
    subcommands.DieNotNil(sota.Delete("pacman.compose_apps"))
}
```

vkhoroz: The other option, as we know a key we pass is properly formatted, would be to make an "optimistic" assumption that `sota.Delete` always succeeds from a practical perspective:
```suggestion
			_ = sota.Delete("packman.docker_apps")
			_ = sota.Delete("packman.compose_apps")
```

subcommands/config/updates.go
```diff
@@ -24,7 +24,13 @@ in a device group. When run without options, prints out the current configuratio
   fioctl config updates --group beta --apps shellhttpd
 
   # Set the docker apps and the tag for devices:
-  fioctl config updates --group beta --apps shellhttpd --tag master`,
+  fioctl config updates --group beta --apps shellhttpd --tag master
+
+  # Set the docker apps to none for devices:
```
doanac, line 29: explain that none means "run no apps"

subcommands/config/updates.go
```diff
@@ -24,7 +24,13 @@ in a device group. When run without options, prints out the current configuratio
...
+  fioctl config updates --apps ,
+
+  # Set the docker apps to inherit from parent for devices:
```
doanac, line 32: explain what "parent" means.

subcommands/devices/config_updates.go
```diff
@@ -26,7 +26,12 @@ currently configured and reporting.`,
 
   # Set the docker apps and the tag:
   fioctl devices config updates <device> --apps shellhttpd --tag master
-`,
+
+  # Set the docker apps to none:
+  fioctl devices config updates <device> --apps ,
+
+  # Set the docker apps to inherit from parent:
+  fioctl devices config updates <device> --apps -`,
```
doanac, line 34: explain what "parent" and "none" mean here also

mike-sul:
Tangential note: we don't need to support `pacman.docker_apps` not sure we've ever used it.

subcommands/common_config.go
StealthyCoder: Even though we have properly formatted ones, maybe in the future some other error check will be introduced so I went with the aforementioned one. Check of the key and delete it.

subcommands/config/updates.go
```diff
@@ -24,7 +24,15 @@ in a device group. When run without options, prints out the current configuratio
...
   # Set the docker apps and the tag for devices:
-  fioctl config updates --group beta --apps shellhttpd --tag master`,
+  fioctl config updates --group beta --apps shellhttpd --tag master
+
+  # Set the docker apps to none for devices, this will make the device run no apps:
+  fioctl config updates --apps ,
+
+  # Set the docker apps to inherit from their parent, meaning:
+  # If for a device apps are not set  - all apps are installed
+  # If for a device apps are set      - those apps are installed
```
doanac, line 34: We still aren't explain what the "parent" will be. I'm not sure parent is a good word. What we are trying to describe is how are config files are processed. We use the same logic as [systemd](https://www.freedesktop.org/software/systemd/man/latest/systemd-system.conf.html#Configuration%20Directories%20and%20Precedence) where we walk through:
 * /usr/lib/sota/conf.d/
 * /var/sota/sota.toml
 * /etc/sota/conf.d/

We need a concise way to explain this.

And that `,` and `-` are so subtle here that we should probably call out that they are special characters.

vkhoroz: IIUC the group level is the top one here (as we allow no factory-wide updates config).
So, probably, we may put here a sentence like this:
```suggestion
# Set the docker apps to a system default (all apps on most devices):
```

A _system default_ seems to be rather specific to me, without diving into internal details (which can differ from device to device).

subcommands/devices/config_updates.go
vkhoroz: I would state explicitly here that we are inheriting from the group config.

```suggestion
# Set the docker apps to inherit from a group config:
```

We might be more verbose (like below), but I think that is apparent from the `fioctl config updates` help.
```suggestion
# Set the docker apps to inherit from a group config (if set there) or a system default (otherwise):
```

vkhoroz: Once I wrote this I understood we don't yet have a way to restore the device tag to inherit from a group config or system default.
Do we need one?

@doanac @mike-scott ^^^

doanac: probably should do it also

doanac (commented):
technically it looks good. Lets see what @kprosise says about the wording

kprosise (approved):
LGTM, the latest changes seem to address feedback provided by other reviewers. Only suggestion has to do with style, but imho the descriptions communicate the usage well, neither too verbose or terse.

mike-sul:
My 2 cents on the terms we use. IMHO, we should use the term "Compose apps" not "Docker apps" to make it consistent with the user doc.

subcommands/devices/config_updates.go
doanac: note: Our UI refers to this as using "preset-apps" and "preset-tag". Probably should make this consistent

doanac (approved):

subcommands/config/updates.go
```diff
@@ -20,11 +20,31 @@ in a device group. When run without options, prints out the current configuratio
...
+       - /var/sota/sota.toml
+       - /etc/sota/conf.d/
+  fioctl config updates --apps -
+
+  # Set the device tag to a preset-tag :
+	# 	The system will look in the following locations in order to get the complete config:
+       - /usr/lib/sota/conf.d/
+       - /var/sota/sota.toml
+       - /etc/sota/conf.d/
+  fioctl config updates --tag -`,
```
vkhoroz, line 47: An indentation here seems rather weird.
Maybe, @kprosise can help with it?

At the least, these two sentences should be indented at the same level:
```
# The following explains what happens if you provide a ","
# The following explains what happens if you provide a "-"
```

The same applies to these three sentences (about indentation):
```
# Set the Compose apps to none for devices, this will make the device run no apps:
# Set the Compose apps to preset-apps (all apps on most devices):
# Set the device tag to a preset-tag :
```

Also, as sentences in these paragraphs are user-facing, they deserve a proper punctuation at their ends, like `.` or `:` (and proper spacing, like e.g. no space before a colon).

vkhoroz (approved):
LGTM

Once Katrina is happy, this can be merged.

subcommands/config/updates.go
```diff
@@ -20,11 +20,31 @@ in a device group. When run without options, prints out the current configuratio
...
   fioctl config updates --group beta --apps shellhttpd
 
-  # Set the docker apps and the tag for devices:
-  fioctl config updates --group beta --apps shellhttpd --tag master`,
+  # Set the Compose apps and the tag for devices:
+  fioctl config updates --group beta --apps shellhttpd --tag master
+
+	# There are two special characters: "," and "-"
+	# The following explains what happens if you provide a ","
+  # Set the Compose apps to none for devices, this will make the device run no apps:
```
kprosise, line 31:
```suggestion
  # There are two special characters: "," and "-".
  # Providing a "," sets the Compose apps to "none" for devices.
  # This will make the device run no apps:
```

subcommands/config/updates.go
```diff
@@ -20,11 +20,31 @@ in a device group. When run without options, prints out the current configuratio
...
+  fioctl config updates --apps ,
+
+  # The following explains what happens if you provide a "-"
+	# Set the Compose apps to preset-apps (all apps on most devices):
+	# 	The system will look in the following locations in order to get the complete config:
```
kprosise, line 36:
```suggestion
  # Providing a "-" sets the Compose apps to "preset-apps" (all apps on most devices).
  # The system looks in the following locations to get the complete config:
```

subcommands/config/updates.go
```diff
@@ -20,11 +20,31 @@ in a device group. When run without options, prints out the current configuratio
...
+       - /usr/lib/sota/conf.d/
+       - /var/sota/sota.toml
+       - /etc/sota/conf.d/
+  fioctl config updates --apps -
+
+  # Set the device tag to a preset-tag :
+	# 	The system will look in the following locations in order to get the complete config:
```
kprosise, line 43:
```suggestion
  # Set the device tag to a "preset-tag",
  # and the system will look in the following locations to get the complete config:
```

subcommands/devices/config_updates.go
```diff
@@ -21,12 +21,31 @@ currently configured and reporting.`,
...
   fioctl devices config updates <device> --apps shellhttpd
 
-  # Set the docker apps and the tag:
+  # Set the Compose apps and the tag:
   fioctl devices config updates <device> --apps shellhttpd --tag master
-`,
+
+	# There are two special characters: "," and "-"
+	# The following explains what happens if you provide a ","
+  # Set the Compose apps to none, meaning it will run no apps:
```
kprosise, line 32:
```suggestion
  # There are two special characters: "," and "-".
  # Providing a "," sets the Compose apps to "none", meaning it will run no apps:
```

subcommands/devices/config_updates.go
```diff
@@ -21,12 +21,31 @@ currently configured and reporting.`,
...
+  fioctl devices config updates <device> --apps ,
+
+  # The following explains what happens if you provide a "-"
+	# Set the Compose apps to preset-apps (all apps on most devices):
+	# 	The system will look in the following locations in order to get the complete config:
```
kprosise, line 37:
```suggestion
  # Providing a "-" sets the Compose apps to "preset-apps" (all apps on most devices).
  # The system looks in the following locations to get the complete config:
```

subcommands/devices/config_updates.go
```diff
@@ -21,12 +21,31 @@ currently configured and reporting.`,
...
+       - /usr/lib/sota/conf.d/
+       - /var/sota/sota.toml
+       - /etc/sota/conf.d/
+  fioctl devices config updates <device> --apps -
+  
+  # Set the device tag to a preset-tag :
+	# 	The system will look in the following locations in order to get the complete config:
```
kprosise, line 44:
```suggestion
  # Set the device tag to a "preset-tag",
  # and the system will look in the following locations to get the complete config:
```

kprosise (approved):
Provided suggestions based upon feedback from @vkhoroz.

mike-sul:
@StealthyCoder Is this PR ready for merging?

PR #441, opened 2024-11-11
targets: Add missing fields to TufCustom

detsch:
The `fioctl targets add` command must retain the arch and image-file custom fields from the previous target. Otherwise, the publish-compose-apps operation for new targets will fail.

This commit adds both field to the structure used in the process, allowing the required fields to be copied to the target being added.

---

@mike-sul after some additional tests with targets creation, it is clear that this change is in fact required.

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -141,6 +141,8 @@ type TufCustom struct {
 	UpdatedAt      string                `json:"updatedAt,omitempty"`
 	LmpVer         string                `json:"lmp-ver,omitempty"`
 	FetchedApps    *FetchedApps          `json:"fetched-apps,omitempty"`
+	Arch           string                `json:"arch,omitempty"`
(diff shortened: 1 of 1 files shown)


mike-sul (approved):

mike-sul:
@detsch This change LGTM, and it fixes the issue you have in mind. I would use this PR to fix another related issue - the `origUri` field is not set in the produced "app" target if the source target(s) is(are) produced by FIO CI LmP build. As result, `fioctl targets offline-update` cannot download ostree repo.

mike-sul (approved):

mike-sul:
@detsch Let's rebase and merge it, and the `orgUri` issue can be addressed in another PR then.

mike-sul:
@vkhoroz It would be great to include this commit into the upcoming release too.

PR #442, opened 2024-11-14
targets: Handle OrigUri and OrigUriApps when adding targets

detsch:
For custom apps targets, origUri must be set, while uri and origUriApps must be empty.
For custom ostree targets, origUriApps must be set, while uri and origUri must be empty.

---
@mike-sul this change seems to work, but is not enough. Apps targets generated by our CI are keeping the `origUriApps` field set. I'd say we need to fix the behavior there before merging this change to fioctl. Otherwise, an old `origUriApps` value is carried on indefinitely.

### DIFF
--- a/client/foundries.go
+++ b/client/foundries.go
@@ -137,6 +137,7 @@ type TufCustom struct {
 	OverridesSha   string                `json:"meta-subscriber-overrides-sha,omitempty"`
 	Uri            string                `json:"uri,omitempty"`
 	OrigUri        string                `json:"origUri,omitempty"`
+	OrigUriApps    string                `json:"origUriApps,omitempty"`
 	CreatedAt      string                `json:"createdAt,omitempty"`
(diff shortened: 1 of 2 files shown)


mike-sul (approved):

mike-sul:
> Apps targets generated by our CI are keeping the `origUriApps` field set.

If by "CI" do you mean the publish job/run, then setting both `uri` and `origUriApps` to the same value (URI to the given publish run) is probably fine.

detsch:
Once https://github.com/foundriesio/ci-scripts/pull/363 is merged, I'll test this patch a little bit more, to make sure there is no additional change required.

mike-sul:
> Once [foundriesio/ci-scripts#363](https://github.com/foundriesio/ci-scripts/pull/363) is merged, I'll test this patch a little bit more, to make sure there is no additional change required.

The ci-script has been merged.

detsch:
Did some additional testing, looks OK.

PR #449, opened 2025-01-13
Fixes to the git credential helper

doanac:
This set of changes fixes some rough edges with our git credential helper.

The first change fixes something that "works" but was reckless where we accidentally configured our helper for all "https:*" URLs.

The last 2 changes address the complexity we've had trying to reliably make a symlink from "fioctl -> git-credential-fio". It turns out we never needed to do that and could have gone with a simpler approach that has less ways to fail.


StealthyCoder (approved):
Looks much cleaner, easier and nicer. :muscle:

doanac:
force pushed a change tot get things working better for windows.

doanac:
testing done for windows with both cmd.exe and bash.exe. Also tested with and without spaces in file names.

PR #450, opened 2025-01-21
Bump fioctl dependencies

doanac:
The main purpose is to get rid of 2 CVEs. They don't affect us but it causes noise in our workflows. While looking at this, I also bumped a couple other direct dependencies.


doanac:
github proposed a similar set of patches that don't require a compiler change.

PR #453, opened 2025-02-10
fix(targets): make tag subcommand deny empty tags

StealthyCoder:
This will not allow the user to potentially shoot themselves in the foot by denying potential empty tags list.

### DIFF
--- a/subcommands/targets/tag.go
+++ b/subcommands/targets/tag.go
@@ -20,7 +20,7 @@ var (
 )
 
 func init() {
-	var tagCmd = &cobra.Command{
+	tagCmd := &cobra.Command{
 		Use:   "tag <target> [<target>...]",
 		Short: "Apply a comma separated list of tags to one or more Targets.",
 		Example: `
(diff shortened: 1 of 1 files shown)


doanac:
can you go ahead and add a separate commit to fix the CI issue? I know its not your problem but would be good to fix it while we are doing thins.

subcommands/targets/tag.go
```diff
@@ -78,6 +78,8 @@ func doTag(cmd *cobra.Command, args []string) {
 					targetTags := tags
 					if tagAppend {
 						targetTags = Set(custom.Tags, tags)
+					} else if len(targetTags) == 0 {
+						subcommands.DieNotNil(fmt.Errorf("Cannot change tags from %s -> %s. Empty values are not allowed.", custom.Tags, targetTags))
```
vkhoroz, line 82: I'm wondering why do we make this check here?
Wouldn't it be easier and better to check `if len(tags) == 0 { errorOut }` at line 60?
Or maybe just set something like the minlength in cobra argument definition?

vkhoroz (approved):
LGTM

I'm sorry for that "artifitial build failure" from guthub.
What is you try to fix it separately, then merge and rebase this PR?

Anyway, do whatever takes less time. I'm totally fine with your fix for the topic.

PR #463, opened 2025-04-23
fix(root): allow empty args for autocomplete

StealthyCoder:
The completion command fails now on `fish` because when you run `fioctl ` and then hit `<Tab>` then the auto complete will fail because we will always return an error, which leads to not seeing all the options possible.


vkhoroz (approved):
Seems reasonable.

PR #465, opened 2025-05-05
client: Move fiotest logic to its own file

doanac:


doanac:
low priority but I've been sitting on this cleanup change for a while. It will also make my eventual changes for "remote actions" a little cleaner

PR #466, opened 2025-05-07
Upgrade golang from 1.19 to 1.24

vkhoroz:


.golangci.yml
```diff
@@ -1,21 +1,25 @@
...
     - dupl
     - errcheck
-    - exportloopref
     - gocyclo
-    - gofmt
-    - gosimple
     - govet
     - ineffassign
     - staticcheck
-    - typecheck
```
doanac, line 16: Have typecheck, gosimple, and exportloopref been removed from the new version?

doanac (approved):
lgtm

.golangci.yml
vkhoroz: As per golangci-lint documentation:
- typecheck was never a linter... no idea what it was doing.
- gosimple is now a part of staticcheck.
- exportloopref was already deprecated when we added it and is removed in a new version (seems to be a part of gocyclo now).

I saw quite a few other interesting new linters, but assessing each of them would take time...

PR #472, opened 2025-05-21
show: Add flag to print target custom json

mike-sul:
Extend the `show` command with a flag to print target custom data in raw JSON format.

It is often useful to view all fields of a target's custom data in its original JSON form, especially when debugging or troubleshooting issues related to updating a device to a specific target.


doanac (approved):
good idea.

PR #506, opened 2025-10-03
config: Fix pattern matching for tags

doanac:
We have different rules for configuring tag names than app names. This change pulls in the regular expression used by our backend to validate tag names. This will allow users to configure a device to follow a tag like `foo.bar` which was already possible, but fioctl would not allow.


mike-scott (approved):
LGTM

